PR #84, opened 2014-06-10
Add support for building using Vagrant and Ansible

christiansimms:
This uses Vagrant (to manage VirtualBox) and Ansible (to run the initial setup commands). Unfortunately, it sometimes fails the first time it's run. You can just run "vagrant provisition" to run the build scripts a second time.


dagar:
@christiansimms this is cool. If you wanted to get this working the setup is now much simpler and you could copy required packages from the travis-ci config.
https://github.com/dropbox/pyston/blob/master/docs/INSTALLING.md
https://github.com/dropbox/pyston/blob/master/.travis.yml#L20

PR #145, opened 2014-09-02
Added improved string count function, based on CPython one

kkszysiu:
This PR is adding full support for str.count(sub[, start[, end]]) method. I mean with support for empty string as 'sub' or for support for second int as an 'end'. Also allows for negative values in both ot both values altogether - just like standard str.count do.


kmod:
cleaning up old PRs -- feel free to reopen this.

PR #196, opened 2014-10-21
Remove linux-specifics from threading

rpearl:
The `tgkill()` system call is linux-specific. The `gettid()` call is also not necessarily mapped directly to each thread of execution in pthreads.


kmod:
I did something similar in aadf86a146cf38a2c76ca24e5bb10b6e5935a226, so I'm going to close this.

PR #216, opened 2014-11-11
Implementation for set comprehension

vinzenz:
This patch brings set-comprehension support to pyston.

### DIFF
--- a/src/core/cfg.cpp
+++ b/src/core/cfg.cpp
@@ -141,6 +141,11 @@ private:
         return makeCall(makeLoadAttribute(name, "append", true), elt);
     }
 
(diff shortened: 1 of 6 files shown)


src/core/cfg.cpp
```diff
@@ -141,6 +141,11 @@ class CFGVisitor : public ASTVisitor {
         return makeCall(makeLoadAttribute(name, "append", true), elt);
     }
 
+    AST_expr* applyComprehensionCall(AST_SetComp* node, AST_Name* name) {
+        AST_expr* elt = remapExpr(node->elt);
+        return makeCall(makeLoadAttribute(name, "add", true), elt);
+    }
+
```
rntz: This looks like dead code. Where is this called?

PR #242, opened 2014-12-06
fixes #238 CMake issues

### LINKED ISSUE #238: CMake issues/updates
@dagar 

Meta-comment, but could you set up a 12.04 VM?  I'm not aware of any more 12.04-specific issues at the moment, but I'd like to reinforce that 12.04 is a first-class-citizen even though it's not quite as shiny :)

Anyway, the current issue I'm running into is that doing a debug build now takes quite a bit more memory, >1.5GB to link the debug clang it looks like.  Maybe it's because cmake doesn't use gold?  Or maybe because we didn't previously build a debug build of clang at all.

It also takes 11 seconds to link a debug build of the pyston binary, but it was 1.5 with the makefile; this seems to both be because we're not using gold (which cuts it down to 4.2s) and probably because we're doing a much larger link with all the LLVM symbols.  This is my current blocker for switching to cmake :(
## 

We might want to write a "run_cmake.sh" script; it took me quite a while to figure out that this is what I wanted:

```
CC="ccache $HOME/pyston_deps/gcc-4.8.2-install/bin/gcc" \
CXX="ccache $HOME/pyston_deps/gcc-4.8.2-install/bin/g++" \
~/pyston_deps/cmake-3.0.0/bin/cmake -GNinja \
~/pyston -DCMAKE_MAKE_PROGRAM=$HOME/pyston_deps/ninja/ninja \
-DCMAKE_EXE_LINKER_FLAGS="-Wl,-rpath,$HOME/pyston_deps/gcc-4.8.2-install/lib64" \
-DGCC_INSTALL_PREFIX=~/pyston_deps/gcc-4.8.2-install
```

I've already started to construct a psuedo-script in the installation instructions ("if your system has this, run this command, otherwise run this other one, but if you want ccache add it in these places") but it might be nicer to have the computer run it instead of the human.  I also commonly use both the release and debug builds, and for now I manually duplicate the cmake lines in the right directories and change the build type, but that's again something that could be automated.

Maybe it's better to think of it as "providing help setting up the build environment" rather than directly being part of the build system itself, but I think we should make that easy for people.  It could probably also be part of the cmake build system, though my guess is you don't want to do that.
## 

Also, I'm starting to think we should get rid of git submodules -- I definitely also thought they'd be meant for this kind of use case, but they seem to just be misleading and not really provide much benefit.  For libunwind/libpypa it's not a big deal since we didn't have version-tracking functionality anyway, but we have that for llvm and I don't want to lose that.  (Ideally we'd add it for libunwind/libpypa as well.)
## 

I would like to re-request being able to use the just-built clang to build pyston, since I prefer clang's diagnostics and features.  I guess I would actually be fine with a stable-release of clang, though we'd have to weigh adding yet-another-compiler to the dependencies:
- gcc 4.8.2 for the standard libraries
- clang trunk for building the runtime inlines, and now:
- clang 3.5 for building pyston

If we think it's easier to use the just-built clang instead of using a different copy, what I actually would like is our current set up where I can build pyston_dbg using a just-built Release-mode clang.  I think to do this we might need to mirror the current pyston_dbg (as opposed to pyston_debug) build mode, which combines a debug pyston with a release llvm -- this could also fix the increased-link-time issue.

dagar:
This improves the Debug mode linking time by effectively building LLVM Release+Asserts (unless ENABLE_LLVM_DEBUG is set) and pulls ccache and gold configuration into cmake. Gold is enabled by passing -B${CMAKE_SOURCE_DIR}/tools/build_system to gcc or clang for linking and used directly for for src/runtime/inline/*.

### DIFF
--- a/docs/INSTALLING.md
+++ b/docs/INSTALLING.md
@@ -290,24 +290,58 @@ CC="ccache $HOME/pyston_deps/gcc-4.8.2-install/bin/gcc" CXX="ccache $HOME/pyston
 ~/pyston_deps/ninja/ninja check-pyston
 ```
 
-On Ubuntu 12.04, you need a new cmake in order to use ninja, but can install ninja using `sudo apt-get install ninja-build`.  Then:
-
-```
-mkdir ~/pyston-build
-cd ~/pyston-build
-CC="ccache $HOME/pyston_deps/gcc-4.8.2-install/bin/gcc" CXX="ccache $HOME/pyston_deps/gcc-4.8.2-install/bin/g++" ~/pyston_deps/cmake-3.0.0/bin/cmake -GNinja ~/pyston -DCMAKE_EXE_LINKER_FLAGS="-Wl,-rpath,$HOME/pyston_deps/gcc-4.8.2-install/lib64" -DGCC_INSTALL_PREFIX=~/pyston_deps/gcc-4.8.2-install
-ninja check-pyston
-```
-
 If your system provides a new enough GCC and cmake, you can just do:
 
 ```
-mkdir ~/pyston-build
-cd ~/pyston-build
-CC='ccache gcc' CXX='ccache g++' cmake -GNinja ~/pyston
+mkdir ~/pyston-build && cd ~/pyston-build
+cmake -GNinja ~/pyston
 ninja check-pyston
 ```
 
+**Ubuntu 12.04**
+```
+sudo add-apt-repository --yes ppa:ubuntu-toolchain-r/test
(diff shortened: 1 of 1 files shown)


kmod:
Yeah, I'll try to encourage one of the other contributors to use the deferred-to-cmake build as well, and if that all goes well we can make that the default.  The defer-to-cmake functionality in the makefile assumes that there are separate pyston-build-dbg and pyston-build-release directories that it can defer to for either a debug or release build; we should figure out how we want to handle that and make sure that the instructions + Makefile match.

PR #246, opened 2014-12-18
Enable unicode type (WIP)

jvkersch:
This is a very preliminary PR to provide some support for unicode (#130), opened to solicit some feedback. Basically, this PR tries to do the bare minimum needed to get the unicode type working, together with its repr. On the command line, things like the following snippet work:
```
Pyston v0.2 (rev 42c3ecea3443), targeting Python 2.7.6
>> x = u'소주 한잔'
>> x
u'\uc18c\uc8fc \ud55c\uc794'
```
Since Pyston seems to be configured as a wide unicode build, the code just does some tricks to get unicode strings into UTF-32, and then wraps those into a `Unicode` instance. Calling into the existing CPython unicode implementation is then just a matter of passing in the underlying `const PyUnicode *` to the appropriate routine.

Does this sound like a viable approach? Should we instead try to support the Python 3 unicode implementation?

Things that I haven't addressed yet:
- [ ] Parsing via Pypa;
- [ ] When creating a unicode instance, `boxUnicode` iterates over the parsed string to convert it into a wide string. This can easily be avoided by reading in a wide string in the first place.
- [ ] There's no support for supplying an encoding string for a source file (and `compile` in parse_ast.py will complain if you do provide one), so that we can't run meaningful comparisons with Python.

I'm also not sure as to how much of `Objects/unicodeobject.c` I should try to include. I could definitely add that file in its entirety (surrounded by `#if 0 ... #endif`) but would rather do that as a separate PR, to avoid drowning out changes to the Pyston source tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,7 +278,7 @@ STDLIB_OBJS := stdlib.bc.o stdlib.stripped.bc.o
 STDLIB_RELEASE_OBJS := stdlib.release.bc.o
 
 STDMODULE_SRCS := errnomodule.c shamodule.c sha256module.c sha512module.c _math.c mathmodule.c md5.c md5module.c _randommodule.c _sre.c operator.c binascii.c pwdmodule.c posixmodule.c $(EXTRA_STDMODULE_SRCS)
-STDOBJECT_SRCS := structseq.c capsule.c $(EXTRA_STDOBJECT_SRCS)
+STDOBJECT_SRCS := structseq.c capsule.c unicodeobject.c $(EXTRA_STDOBJECT_SRCS)
 STDPYTHON_SRCS := pyctype.c getargs.c $(EXTRA_STDPYTHON_SRCS)
 FROM_CPYTHON_SRCS := $(addprefix lib_python/2.7_Modules/,$(STDMODULE_SRCS)) $(addprefix lib_python/2.7_Objects/,$(STDOBJECT_SRCS)) $(addprefix lib_python/2.7_Python/,$(STDPYTHON_SRCS))
 
--- a/include/ucnhash.h
+++ b/include/ucnhash.h
@@ -1,0 +1,33 @@
+/* Unicode name database interface */
+
+#ifndef Py_UCNHASH_H
+#define Py_UCNHASH_H
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* revised ucnhash CAPI interface (exported through a "wrapper") */
+
+#define PyUnicodeData_CAPSULE_NAME "unicodedata.ucnhash_CAPI"
+
+typedef struct {
+
+    /* Size of this struct */
+    int size;
+
+    /* Get name for a given character code.  Returns non-zero if
+       success, zero if not.  Does not set Python exceptions. 
+       If self is NULL, data come from the default version of the database.
+       If it is not NULL, it should be a unicodedata.ucd_X_Y_Z object */
+    int (*getname)(PyObject *self, Py_UCS4 code, char* buffer, int buflen);
+
+    /* Get character code for a given name.  Same error handling
+       as for getname. */
+    int (*getcode)(PyObject *self, const char* name, int namelen, Py_UCS4* code);
+
+} _PyUnicode_Name_CAPI;
+
+#ifdef __cplusplus
+}
+#endif
+#endif /* !Py_UCNHASH_H */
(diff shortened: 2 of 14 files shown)


kmod:
Cool!  Seems like good progress.  I'm pretty unfamiliar with the ins and outs of implementing unicode, so I'm not sure I can help much when it comes to those issues (ex I don't know how Pyston gets configured as a wide unicode build).

I think in general the strategy we're narrowing in on is to try to import as much as we can from CPython, and rewrite it if it's used enough that there are performance wins to be had.  So I think I'm glad we are reimplementing "str", but I'm leaning towards porting "unicode" if we can -- we can always reimplement it later if we want.

As for the Python 2 implementation vs Python 3, I don't know the details of the change, but my guess is there is some aspect of the C API that made them want to delay the implementation change until Python 3?  I don't know what the C API exposes about unicode, so we'd probably have to take a look through unicodeobject.h and familiarize ourselves with what unicode objects look like to extension modules.

I'd also be open to a reimplementation if that's the way you want to go :)  My suggestion though would be to take a look at unicodeobject.h and see how much leeway we have for implementation, and try importing the Python 2/3 unicodeobject.c and see how much work there might be to get that working.  Once we have a better handle on those two things I think we'll have a much better idea how to proceed.

jvkersch:
@kmod This makes a lot of sense, thanks! In the meantime I've been playing with this some more, and I have a good idea of what needs to be reimplemented vs. imported. In any case, it looks like some of the low-level stuff (e.g. `_PyUnicode_New`) will have to be reimplemented as those functions are just thin wrappers around Pyston objects.

I have one remaining practical question: how would you like to see this PR proceed? Do you prefer having small PRs that expose a little functionality at a time, or would you prefer a big PR that imports whatever can be imported from Python, and disables the rest for future work?

kmod:
Hmm I'd be interested in hearing more about what you're thinking this will look like -- in theory we could use only the CPython implementation and not have to reimplement any of it, since we're developing pretty decent support for types defined in extension modules.

I think how to proceed depends on whether it's a mostly-port or mostly-rewrite (either of which I think would be easier than something in the middle) -- I think with a mostly-rewrite we should start small and add things incrementally, but with a mostly-port we should try to copy the entire file over and Py_FatalError the parts that don't work yet.

kmod:
Hi @jvkersch, how's the unicode work going?  unicode support is starting to increase in urgency, and we'll probably want to have something within the next few weeks.  Let me know if that works for you, and if it'll be tough to do that yourself we can also help more with the process :)

kmod:
Hi @jvkersch, do you think you'll have this done soon?  It's no problem either way, but we're going to be blocked on this soon so if you don't have time we'd be happy to take this on.

kmod:
Hi @jvkersch, I took a quick look at this and it looks like the way we would want to do it (adding unicodeobject.c) gets pretty hairy.  Unless we hear back from you soon we'll probably have to start working on this soon ourselves :/

jvkersch:
> Unless we hear back from you soon we'll probably have to start working on this soon ourselves :/

Hi @kmod, sorry for dropping the ball on this. I'll look at what's currently in master, and if I find a manageable way to contribute, I'll try doing so.

kmod:
No worries, we were happy to do it but wanted to make sure you got the chance to if you had time.

I managed to get CPython's unicodeobject.c into Pyston, which was pretty hairy since it used a number of other things as well.  So our unicode support is coming along!  But there's a lot of stuff still left to do beyond simply supporting the datatype.

If you're interested in unicode-related stuff, one super useful thing would be to add the ability for us to print unicode to the console, such as `print u'\u0180'`.  This works by setting an encoding on the sys.stdout file object, which causes (or at least is supposed to cause) any writes of unicode to that file to get encoded with that particular encoding.  I'm not sure where the encoding gets set or how Python agrees with the terminal on the encoding to use, so it might take some digging in CPython.

```python
>>> import sys, os
>>> print >>sys.stdout, u'\u0180'
ƀ
>>> sys.stdout.encoding
'UTF-8'
>>> print >>os.fdopen(sys.stdout.fileno(), 'w'), u'\u0180'
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0180' in position 0: ordinal not in range(128)
```

jvkersch:
> super useful thing would be to add the ability for us to print unicode to the console

I'll take a look this weekend; the combination of digging around in Python and doing something unicode-related sounds very tempting :-)

Issue #171, opened 2014-09-19
Lists should compare by identity in addition to equality


kmod:
This is important for objects that are not equal to themselves, such as NaN.

CPython:
```
Python 2.7.1+ (r271:86832, Apr 11 2011, 18:13:53)
>>> nan = float('nan')
>>> nan == nan
False
>>> nan in [nan]
True
>>> [nan] == [nan]
True

```

Pyston:
```
Pyston v0.2 (rev 9376765775f0), targeting Python 2.7.1
>> nan = float('nan')
>> nan == nan
False
>> nan in [nan]
False
>> [nan] == [nan]
False
```

kmod:
Closed by #248

PR #248, opened 2014-12-21
Trying to fix Issue #171

dorafmon:
I tried to fix issue #171 and I also added a line ``$(LLVM_CONFIGURE_LINE) --disable-bindings`` in Makefile because if I don't have ``-disable-bindings`` flag, I won't be able to configure LLVM properly, there is an error saying since I have enabled bindings then I must install ctypes on my machine, which I have installed.

Back to bug #171, I tried to add some simple identity testing in ``__contains__`` and ``__eq__`` method. But I don't know for sure this is sufficient. I wonder should what does ``compareInternal`` do exactly? Should we put the identity test there?

Also, this is my first pull request, I have signed the agreement. Thanks!


kmod:
`nan in [nan]` is True since `list.__contains__` explicitly tests for object identity equality in addition to object equality, similar to how `[nan] == [nan]` is True.

[The thread I linked](https://mail.python.org/pipermail//python-dev/2014-July/135476.html) has some justification for this behavior.  Seems like either choice for whether `nan in [nan]` would result in weird cases.

PR #250, opened 2014-12-22
implement viewkeys viewvalues viewitems method on dict object

aisk:


src/runtime/dict.cpp
```diff
@@ -90,6 +90,21 @@ Box* dictKeys(BoxedDict* self) {
     return rtn;
 }
 
+Box* dictViewKeys(BoxedDict* self) {
+    BoxedDictView* rtn = new BoxedDictView(self, dict_keys_cls);
```
kmod: We should add a check here to make sure that self is actually a dict object -- it's well-defined to call `dict.viewkeys(1)`, which is supposed to throw a TypeError.  Here's an example: https://github.com/dropbox/pyston/blob/master/src/runtime/int.cpp#L230

We do have the ability to make sure that the function-calling machinery will only call dictViewKeys with an actual dict, but we would still need the version that can accept any type and throws a TypeError if it's not a dict.  [We do that for the int methods](https://github.com/dropbox/pyston/blob/master/src/runtime/int.cpp#L791), but I don't think that's worth it here.

test/tests/dict.py
```diff
@@ -194,3 +194,16 @@ def custom_keys():
...
+
+print list(d.viewkeys())
+print list(d.viewvalues())
+print list(d.viewitems())
+
+d['a'] = 1
+
+print list(d.viewkeys())
+print list(d.viewvalues())
+print list(d.viewitems())
```
kmod: Can we add another check here that tests the difference between viewkeys and keys (and the others)?  ie mutate the dictionary and show that the views get updated but the non-views don't.

kmod:
Cool, looks good!  Just a few minor comments.  Also, could you make sure to sign the [Dropbox CLA](https://opensource.dropbox.com/cla/)?

aisk:
Hi, the CLA is signed ~

PR #260, opened 2015-01-07
Generators

### LINKED ISSUE #152: Generator stacks are too unsafe
The stacks for generators are currently heap-allocated, which means that if there is any stack overflow (quite easy since we don't have any guards against that), we will start overwriting arbitrary heap data.

We should probably mmap separate memory for the stacks and have non-dereferenceable memory on either side.

kmod:
Some invasive generator changes.

We were getting some segfaults from LLVM taking up enough stack to use up our entire generator stack.

Fixes #152


undingen:
cool thanks for fixing my code!
I also didn't know about the ```MAP_GROWSDOWN``` flag.

Issue #167, opened 2014-09-19
Set GMP to use our GC as its allocator


kmod:
Currently GMP (the bignum library we use) uses malloc()/free() to manage bignum memory.  This means that 1) we never free bignum memory, and 2) we don't get nice behavior if we run out of memory [well that's not something we support well anyway].

GMP supports custom allocators, and it looks like we should be able to just set it to use our GC:
https://gmplib.org/manual/Custom-Allocation.html
example of another project that did it:
http://git.savannah.gnu.org/cgit/guile.git/commit/?id=b57bf2724a254cf73e255b3d2bca0a7dcad6674a

dorafmon:
Hi, I am trying to fix this issue and this is how I do it:
In long.cpp I define:
```
void * customised_allocation(size_t alloc_size){
    return gc::gc_alloc(alloc_size, gc::GCKind::CONSERVATIVE);
}

void * customised_realloc(void *ptr, size_t old_size, size_t new_size){
    return gc::gc_realloc(ptr, new_size);
}

void customised_free(void *ptr, size_t size){
    gc::gc_free(ptr);
}

```

and then in ``setupLong()`` I have this at the beginning:
```
    mp_set_memory_functions(
        customised_allocation,
        customised_realloc,
        customised_free);
```

But I kept getting error saying the free is fed with a pointer that it don't know what to do with. Is there any thing wrong I did that seems obvious for you? Thanks.

dorafmon:
Never mind I fixed the bug.

PR #266, opened 2015-01-23
fixing #167: using gc allocator for gmp

dorafmon:
This is a simple fix for #167, it passes ``make check``. I am not sure if I should test the GC but again, I don't really know how to test the GC.

### DIFF
--- a/src/runtime/long.cpp
+++ b/src/runtime/long.cpp
@@ -385,7 +385,6 @@ Box* longRepr(BoxedLong* v) {
     strcat(buf, "L");
 
     auto rtn = new BoxedString(buf);
-    free(buf);
 
     return rtn;
 }
@@ -397,7 +396,6 @@ Box* longStr(BoxedLong* v) {
 
     char* buf = mpz_get_str(NULL, 10, v->n);
     auto rtn = new BoxedString(buf);
-    free(buf);
 
     return rtn;
 }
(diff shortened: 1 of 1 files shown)


src/runtime/long.cpp
```diff
@@ -385,7 +385,6 @@ Box* longRepr(BoxedLong* v) {
     strcat(buf, "L");
 
     auto rtn = new BoxedString(buf);
-    free(buf);
```
kmod: I don't think this call to free() should be removed?  We allocate the space with malloc()

src/runtime/long.cpp
```diff
@@ -397,7 +396,6 @@ Box* longStr(BoxedLong* v) {
 
     char* buf = mpz_get_str(NULL, 10, v->n);
     auto rtn = new BoxedString(buf);
-    free(buf);
```
kmod: Hmm, could we have longStr call malloc like longRepr?  It would make this behavior not dependent on whether or not we're using the custom allocator.

dorafmon:
@kmod Hi I updated the code and now it should be fine now.

kmod:
The motivating example was the pidigits microbenchmark (uses tons of longs), so I reran it with this change and things segfaulted and I realized that neither of us thought about making sure that this new memory gets scanned :)  Fixed in https://github.com/dropbox/pyston/commit/d08aff31387e861dc51b3fd929ab02e80ec319cd

thanks for the PR!

dorafmon:
Hi, thanks for reminding, I forgot to do that. Thanks a lot for showing me. :smiley:

PR #267, opened 2015-01-25
Str just funcs

tjhance:
Implement `str` functions `ljust`, `rjust`, and `center`.

### DIFF
--- a/src/runtime/str.cpp
+++ b/src/runtime/str.cpp
@@ -976,6 +976,59 @@ extern "C" Box* strNe(BoxedString* lhs, Box* rhs) {
     return boxBool(lhs->s != srhs->s);
 }
 
+#define JUST_LEFT 0
+#define JUST_RIGHT 1
+#define JUST_CENTER 2
+static Box* pad(BoxedString* self, Box* width, Box* fillchar, int justType) {
+    assert(width->cls == int_cls);
+    assert(fillchar->cls == str_cls);
+    assert(static_cast<BoxedString*>(fillchar)->s.size() == 1);
+    int64_t curWidth = self->s.size();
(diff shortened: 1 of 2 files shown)


src/runtime/str.cpp
```diff
@@ -976,6 +976,57 @@ extern "C" Box* strNe(BoxedString* lhs, Box* rhs) {
...
+            padRight = 0;
+            break;
+        case JUST_CENTER:
+            padLeft = nNeeded / 2 + (nNeeded & targetWidth & 1);
+            padRight = nNeeded - padLeft;
+            break;
+    }
+
+    std::string res = std::string(padLeft, c) + self->s + std::string(padRight, c);
+    return new BoxedString(res);
```
undingen: nit-picking: you could add a ```std::move(res)``` here to avoid a copy.

kmod:
The str functions look good, but did you mean to include the sys_test.py change?

I'm definitely curious as to why it's failing.  What does `make check_sys_test` (new convenience target) say?

tjhance:
Sorry, I should have mentioned that...

Yeah it was failing on my machine. My sys.exec_prefix for CPython is different than Pyston's hardcoded value `/usr`:

```C++
sys_module->giveAttr("exec_prefix", boxStrConstant("/usr"));
```

Since Pyston doesn't really have a way to configure it, it seems like the functionality isn't really implemented and shouldn't be tested. I could change it to `# skip-if: sys.exec_prefix != '/usr'` or something though.

kmod:
Well, this is something that should work (it gets used when self-host building extension modules).  Could we separate this into a separate PR?  It might not be too bad to do it the right way and port CPython's `Py_GetExecPrefix` and friends.

PR #268, opened 2015-01-26
Add a simple cache for runtime ICs

undingen:
spectral_norm.py runtime improves by about 5%:
rewriter_attempts: 135109 --> 55133
rewriter_commits: 80138 --> 162
runtime_ics_num: 40003 --> 15
slowpath_binop: 82702 --> 2726

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -311,11 +311,12 @@ extern "C" Box* sum(Box* container, Box* initial) {
     if (initial->cls == str_cls)
         raiseExcHelper(TypeError, "sum() can't sum strings [use ''.join(seq) instead]");
 
-    BinopIC pp;
+    static RuntimeICCache<BinopIC, 3> runtime_ic_cache;
+    std::shared_ptr<BinopIC> pp = runtime_ic_cache.getIC(__builtin_return_address(0));
 
     Box* cur = initial;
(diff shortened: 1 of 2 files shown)


kmod:
Cool :)  I was a little bit worried about whether or not this could be reentrant (ie what if sum() called sum() which evicted an IC that was being used?) but it looks like the shared_ptr should make it work?

Do you have a sense of how much of the improvement is from the cache vs making the BinopICs polymorphic?

undingen:
Yes, to make the cache reentrant I used a shared_ptr instead of a normal pointer. I think it should be save because even when a reentrant call replaces the "parents" cache slot (when the cache is full) the runtime IC will not get deleted until the parent functions returns. I'm just not sure if I need a lock when several threads call "sum" at the same time? or if the cache should be thread local? 

You are right my numbers are not showing the improvements for the cache alone...
Here are the number for 2 two BinopIC patchpoint slots without IC cache and with:
```
rewriter_attempts: 135107 -->  55133
rewriter_commits: 80138 --> 162
rewriter_nopatch: 14972 --> 14974
rewriter_starts: 80149 -->  173
runtime_ics_num: 40003 --> 15
slowpath_binop: 82702 --> 2726
slowpath_callfunc: 292874 -> 212901
```
Why are there two more nopatch's with activated cache??

I added the cache not only for the small performance gain but in part because I found that all this temporary IC made tracking down performance issues harder.

PR #270, opened 2015-01-28
use a vector of chunks for the TraceStack, instead of a vector of individual pointers.

toshok:
also keep a free list of chunks around to make subsequent collections faster.

results in TraceStack::pop and ::push being inlined and disappearing from the perf report (::push was at 3.99% before).

Also drops aggregate GC times for raytrace by ~5%.

before: gc_collections_us: 2151827
after: gc_collections_us: 2023809

### DIFF
--- a/src/gc/collector.cpp
+++ b/src/gc/collector.cpp
@@ -39,11 +39,44 @@ namespace gc {
 
 class TraceStack {
 private:
-    std::vector<void*> v;
+    const int CHUNK_SIZE = 256;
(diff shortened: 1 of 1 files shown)


kmod:
Cool!  took me a sec to verify that this doesn't need a destructor to free `chunks` :P

kmod:
Also now that I think about it, a std::deque would probably have been a better original choice than std::vector (though this custom implementation that does more caching is probably better still).

PR #271, opened 2015-01-28
intern BoxedFloats for -2, -1, 0, 1, 2, inf, -inf, nan

toshok:
removes >1mil allocations from raytrace, reducing the number of collections from 279 to 268, reducing total gc_collection_us by ~7%.

### DIFF
--- a/src/runtime/float.cpp
+++ b/src/runtime/float.cpp
@@ -23,6 +23,16 @@
 
 namespace pyston {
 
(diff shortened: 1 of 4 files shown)


kmod:
Seems pretty reasonable to me, though I'm not sure why it's failing the testsuite.  You can also do `make format` to batch-apply our project formatting rules.

PR #275, opened 2015-02-01
give all CLFunctions a ParamNames (renamed from ArgNames) including buil...

tjhance:
Make `boxRTFunction` accept a `ParamNames` argument so the builtin functions can have param names too. I added a test case for `sorted` although it doesn't actually work because (it accepts the keyword argument, but the functionality in `sorted` isn't implemented).

This is a large diff because I modified every call to `boxRTFunction` with an empty `ParamNames` object.

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -1718,8 +1718,7 @@ extern "C" bool nonzero(Box* obj) {
     if (func == NULL) {
         ASSERT(isUserDefined(obj->cls) || obj->cls == classobj_cls || obj->cls == type_cls
                    || isSubclass(obj->cls, Exception),
-               "%s.__nonzero__",
-               getTypeName(obj)->c_str()); // TODO
+               "%s.__nonzero__", getTypeName(obj)->c_str()); // TODO
         return true;
     }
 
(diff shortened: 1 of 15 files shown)


src/runtime/objmodel.cpp
```diff
@@ -2276,19 +2276,14 @@ static std::string getFunctionName(CLFunction* f) {
...
     bool found = false;
-    for (int j = 0; j < arg_names.size(); j++) {
-        AST_expr* e = arg_names[j];
-        if (e->type != AST_TYPE::Name)
-            continue;
-
-        AST_Name* n = ast_cast<AST_Name>(e);
-        if (n->id == kw_name) {
+    for (int j = 0; j < param_names.args.size(); j++) {
+        if (param_names.args[j] == kw_name) {
```
kmod: Is the idea to specify "no keyword name for this parameter" as an empty string?  The user can still pass an empty-name parameter via kwargs `sorted(**{'':[]})` -- maybe we should add a `&& kw_name.size()` check somewhere in here?

kmod:
Cool!  Let's pick a different function than sorted() so we can enable the test -- looks like int() takes keyword arguments ('x' and 'base').

How would you feel about overloading boxRTFunction to automatically provide an empty ParamNames if none is specified?

tjhance:
> How would you feel about overloading boxRTFunction to automatically provide an empty ParamNames if none is specified?

Well, I did consider that (since it would have saved me a lot of gruntwork on this diff). I thought it would be nice to make people check what the ParamNames are supposed to be when adding functions in the future.

Although, someone has to go through all existing functions and figure out which ones need ParamNames *anyway* so maybe we don't gain anything here.

------------------

**[aside]** Ultimately I think we could really use a cleaner way of declaring these built-in functions. Like, right now we are writing
```C++
Box* sorted(Box* l, Box* cmp, Box* key, Box** args) {
    // ...
}
// SCROLL TO THE BOTTOM OF THE FILE
CLFunction* sorted_func = createRTFunction(/* confusing list of parameters here */)
```
but it would be easier to write and maintain something that looks like
```
BUILTIN_METHOD(sorted)
RETURN_TYPE(sorted, list)
ARG1_TYPE(sorted, list)
ARG2_TYPE(sorted, function, "cmp")
ARG3_TYPE(sorted, function "key")
Box* sorted(BoxedList* l, BoxedFunction* cmp, BoxedFunction* key, Box** args) {
    // ...
}
```
I don't really have a concrete proposal for how these macros would work, but I thought I would throw this out there to plant the idea.

src/runtime/objmodel.cpp
tjhance: good catch O:

tjhance:
okay I did a test with a function that we already had implemented (`open`).

I removed all the `ParamNames::empty()` arguments and fixed the other bug that kmod pointed out.

PR #276, opened 2015-02-03
add llvm and clang as git submodules

dagar:


dagar:
@kmod This refreshed PR adds llvm and clang as git submodules and pulls the patch management (with deps) into cmake. This removes the need to manually clone llvm/clang then run llvm_up and it's the final dependency required to be placed in ~/pyston_deps.

I also took a pass at removing obsolete sections of the Makefile. Let me know if there's anything you still wanted. I removed ctags because it was using source defined in the makefile, but I can quickly readd in cmake if anyone actually uses it.

kmod:
Hmm I'm hesitant to change this when we haven't rebased LLVM in a while -- I can't tell if this change will make it harder to go through that process (I assume this breaks `tools/find_problem.py`?).  Also I'm not super happy with how the patch-submodules-in-tree approach worked out, but at the same time I don't want to figure out how to create more repos for pyston (ie I don't want to keep cluttering the dropbox/ account).  I guess I have some build system fatigue at the moment :/  Maybe we can come back to this in a little bit?  I'd be happy to merge the Makefile-pruning stuff though (though I do use `make tags`).

If you're looking for things to do, it sounds like @toshok was having some pretty good results by switching to jemalloc, and there may or may not be work there to get that into our build.

toshok:
+1 on jemalloc.  it's a pretty large improvement when I use the 3.6.0
release (~6% on django_template.py, iirc).  I think we definitely want it
included in the executable (as a .a), but all my tests so far have been
with LD_PRELOAD + .so.


On Mon, Jun 22, 2015 at 6:30 PM, Kevin Modzelewski <notifications@github.com

PR #277, opened 2015-02-03
make __name__ a descriptor

tjhance:
Makes `__name__` a "getset descriptor" for functions and types instead of a direct attribute.

This enables test cases `56.py` and `69.py`. I had to disable `classname.py`, since I haven't yet added in support for *setting* in a `GetsetDescriptor`. Happy to add that into this PR, though, if it's immediately necessary.

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -234,9 +234,9 @@ extern "C" void raiseAttributeError(Box* obj, const char* attr) {
     if (obj->cls == type_cls) {
         // Slightly different error message:
         raiseExcHelper(AttributeError, "type object '%s' has no attribute '%s'",
-                       getNameOfClass(static_cast<BoxedClass*>(obj))->c_str(), attr);
+                       getNameOfClass(static_cast<BoxedClass*>(obj)), attr);
     } else {
-        raiseAttributeErrorStr(getTypeName(obj)->c_str(), attr);
+        raiseAttributeErrorStr(getTypeName(obj), attr);
     }
 }
 
@@ -286,7 +286,7 @@ extern "C" Box** unpackIntoArray(Box* obj, int64_t expected_size) {
 
 void BoxedClass::freeze() {
     assert(!is_constant);
-    assert(getattr("__name__")); // otherwise debugging will be very hard
+    assert(tp_name); // otherwise debugging will be very hard
 
     fixup_slot_dispatchers(this);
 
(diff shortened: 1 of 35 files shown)


src/runtime/objmodel.cpp
```diff
@@ -1052,6 +1064,35 @@ Box* dataDescriptorInstanceSpecialCases(GetattrRewriteArgs* rewrite_args, const
         return runtimeCallInternal1(prop->prop_get, NULL, ArgPassSpec(1), obj);
     }
 
+    // Special case: data descriptor: getset descriptor
+    else if (descr->cls == getset_cls) {
+        BoxedGetsetDescriptor* getset_descr = static_cast<BoxedGetsetDescriptor*>(descr);
+
+        // TODO some more checks should go here
```
kmod: hmm what checks / why don't we need them right now?

src/runtime/objmodel.cpp
```diff
@@ -458,38 +463,45 @@ BoxedHeapClass::BoxedHeapClass(BoxedClass* base, gcvisit_func gc_visit, int attr
     memset(&as_buffer, 0, sizeof(as_buffer));
 }
```
kmod: When would we want to have a NULL name?  (is this constructor dead?)

test/tests/classname.py
```diff
@@ -1,3 +1,5 @@
+# expected: fail
+
```
kmod: I think we should try to add a note as to why the test is marked "expected: fail".  Also you can remove the notes from the test cases that are no longer failing :)

src/runtime/objmodel.cpp
tjhance: Good question. I actually wrote this part ages ago and just dug it up last night... so I don't remember.

*Travis checks cpython*

```C
static int
descr_check(PyDescrObject *descr, PyObject *obj, PyObject **pres)
{
    if (obj == NULL) {
        Py_INCREF(descr);
        *pres = (PyObject *)descr;
        return 1;
    }
    if (!PyObject_TypeCheck(obj, descr->d_type)) {
        PyErr_Format(PyExc_TypeError,
                     "descriptor '%s' for '%s' objects "
                     "doesn't apply to '%s' object",
                     descr_name((PyDescrObject *)descr),
                     descr->d_type->tp_name,
                     obj->ob_type->tp_name);
        *pres = NULL;
        return 1;
    }
    return 0;
}
```

Looks like the getset descriptor saves what type it's supposed to apply to (`d_type`), and it checks that the type that the descriptor is on matches that.

src/runtime/objmodel.cpp
tjhance: Since the other constructors use BoxedString, I have to dance around to get it to work for any types initialized before `str_cls` and then initialize the names later. Maybe there's a better way but nothing jumped out at me.

I think I should probably do something to make this constructor harder to call on accident, since I only intend it for that purpose.

tjhance:
I added some comments. Is there anything else you want changed?

kmod:
Hmm I'm not sure if I'm following the github history correctly, but I can't tell what's changed...  and the Travis build is failing?

tjhance:
Yeah I'm dumb, but it should be fixed now.

Anyway, all I did was add comments and clarified where you asked.

kmod:
Can we add an assert to that constructor that str_cls=NULL?  And can you delete the old comments from the test files (old explanations of why things were failing), and add a comment about what checks you mean by that TODO?  Other than that lgtm.

tjhance:
merged in (via command line)

PR #278, opened 2015-02-04
Small performance improvement when interpreting

undingen:
Speeds up the interpreter by about 10-15% when the higher tiers are disabled, in the normal version it does not change the performance of our benchmarks much because they are too short / to many hot functions :-(.
+ increased the OSR threshold to 200 and the reopt threshold to 100. this values looked like a better tade-off.
+ moved the interpreter in an anonymous namespace.

### DIFF
--- a/src/analysis/scoping_analysis.cpp
+++ b/src/analysis/scoping_analysis.cpp
@@ -14,6 +14,8 @@
 
 #include "analysis/scoping_analysis.h"
 
+#include "llvm/ADT/DenseSet.h"
+
 #include "core/ast.h"
 #include "core/common.h"
 #include "core/util.h"
@@ -112,7 +114,7 @@ struct ScopingAnalysis::ScopeNameUsage {
     const std::string* private_name;
     ScopingAnalysis* scoping;
 
-    typedef std::unordered_set<InternedString> StrSet;
+    typedef llvm::DenseSet<InternedString> StrSet;
 
     // Properties determined from crawling the scope:
     StrSet read;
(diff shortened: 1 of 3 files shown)


kmod:
Cool, sorry for chatting about what you had already done :P

Let's be pretty wary of making tuning changes while we're not confident in our benchmark suite; the threshold change and the DenseMap change don't seem as clear wins as some others.  let's get some more benchmarks in there :)

undingen:
I agree, that's why I said that I think improving the current AST Interpreter / predicting if a new faster one is currently worth it is very hard. (Sorry if I told it abstruse before...)
With the current benchmarks the result is very mixed it - only showed up as an improvement when I totally disabled the llvm tier.

kmod:
Ok, glad we're on the same page :)  I think the "interpreter-only" flag and associated benchmark should make it easier to see the effects of the changes.  We can also commit these as separate PRs so they that they don't end up getting lumped together on speed.pyston.org

Can we try removing \_\_flatten__ and turning back off -fPIC?  I assume this is why the anonymous namespace helped?

undingen:
I will remove it, I put in in the namespace because I thought it's cleaner for stuff which is only visible in the cpp and it should generates slightly smaller code by not having to have an addition uninlined copy of every function I suppose.

performance wise the flatten does not make a difference

kmod:
Also, can we reenable the asserts that the interned strings are from the same pool?  We could maybe have the empty+tombstone keys have some sort of special pool value (maybe -1 again).

undingen:
I will adopt it tomorrow

PR #279, opened 2015-02-04
Replace our time module with the cpython implementation

undingen:
The reason is some benchmarks need time.clock() and the module is easy to import that's why I thought we may want to use cpythons implementation...

add a modified fasta.py (from pypy benchmarks), I modified it to not print out 10KB of output but instead only an md5 hash of the output.


kmod:
Cool :)

PR #281, opened 2015-02-05
add the array module

undingen:


src/capi/typeobject.cpp
```diff
@@ -1729,7 +1729,9 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
 
     RELEASE_ASSERT(cls->tp_getattro == NULL || cls->tp_getattro == PyObject_GenericGetAttr, "");
     RELEASE_ASSERT(cls->tp_setattro == NULL || cls->tp_setattro == PyObject_GenericSetAttr, "");
-    RELEASE_ASSERT(cls->tp_as_buffer == NULL, "");
+
+    // TODO: I think this is not implemented yet but the arraymodules uses it.
+    // RELEASE_ASSERT(cls->tp_as_buffer == NULL, "");
```
kmod: These asserts are in here for a reason :P  We should definitely prefer failing loudly over quietly doing the wrong thing.

I think in this particular case, tp_as_buffer is ok to ignore, since its uses are pretty limited and we don't have a Pyston analogue that it conflicts with.  Could you verify that and then remove this assert entirely?

src/runtime/types.cpp
```diff
@@ -727,9 +729,17 @@ extern "C" int PySlice_GetIndices(PySliceObject* r, Py_ssize_t length, Py_ssize_
...
 }
 
-extern "C" int PySlice_GetIndicesEx(PySliceObject* r, Py_ssize_t length, Py_ssize_t* start, Py_ssize_t* stop,
+extern "C" int PySlice_GetIndicesEx(PySliceObject* _r, Py_ssize_t length, Py_ssize_t* start, Py_ssize_t* stop,
                                     Py_ssize_t* step, Py_ssize_t* slicelength) noexcept {
-    Py_FatalError("unimplemented");
+    try {
+        BoxedSlice* r = (BoxedSlice*)_r;
+        assert(r->cls == slice_cls);
+        parseSlice(r, length, start, stop, step, slicelength);
```
kmod: would you mind verifying that parseSlice has the exact same behavior as this function?  I think it's supposed to, but without a comment saying that it is I feel wary of assuming that it is.

src/capi/typeobject.cpp
undingen: Is calling bf_getreadbuffer and bf_getsegcount on an instance of an array, internally inside a modified pyston runtime function and checking that it's output enough or is there another way to trigger a buffer protocol read?

e.g.:
```C++
    Py_ssize_t totalBytes = 0;
    void* p = 0;
    Py_ssize_t numSegs =  arg->cls->tp_as_buffer->bf_getsegcount(arg, &totalBytes);
    Py_ssize_t numBytes = arg->cls->tp_as_buffer->bf_getreadbuffer(arg, 0, &p);
    printf("numBytes: %zd totalBytes: %zd numSegs %zd\n", numBytes, totalBytes, numSegs);
    for (int i=0; i<numBytes; ++i)
        printf("%c", ((char*)p)[i]);
```

This displays the correct output.

kmod: Sorry, let me rephrase -- most of the tp_* methods are for a class to request non-default behavior for itself, so for something like tp_getattro, if we ignored that being set we'd start silently doing the wrong thing for that class.  For tp_as_buffer, I think it's ok to ignore it, because it doesn't really have a 'default behavior' that it overrides.  So verifying isn't so much checking that a specific class's tp_as_buffer works, but that classes get the behavior that they expect when they set it.  I think in this particular case, it's ok because we haven't implemented anything that deals with tp_as_buffer, so we can just remove the check.

src/runtime/types.cpp
undingen: Looks like they are the same with a small difference of what happens on error: parseSlice will not yet have modified the arguments while PySlice_GetIndicesEx will modify them.
Also our parseSlice can only only handle int subclass arguments (and will release assert so we will notice when we have to implement more).

src/capi/typeobject.cpp
undingen: ok I will remove the check

undingen:
- I removed the tp_as_buffer assert
- converted an under flow check to an release assert because it should never happen and the CPython version does not contain it. (but if it would happen we would catch it)
- added comments

kmod:
Awesome, thanks!

PR #283, opened 2015-02-05
add reversed() builtin function, as well as compatible __reversed__ methods on list/xrange.

toshok:


src/runtime/objmodel.cpp
```diff
@@ -3410,12 +3411,26 @@ Box* getiter(Box* o) {
...
+        return r;
+
+    if (!typeLookup(o->cls, getitem_str, NULL)) {
+        raiseExcHelper(TypeError, "'%s' object is not iterable", getTypeName(o)->c_str());
+    }
+    int64_t len = unboxedLen(o); // this will throw an exception if __len__ isn't there
+
+    return new (seqreviter_cls) BoxedSeqIter(o, len - 1);
+}
+
```
kmod: Tiny preference, but I would put this in the builtins.cpp file if nothing else will be calling this other than the reversed() function.

src/runtime/iterobject.cpp
```diff
@@ -20,6 +20,7 @@
 #include "capi/types.h"
 #include "core/common.h"
 #include "core/options.h"
+
```
kmod: Our lint script doesn't like this newline...

toshok:
Oops, should be fixed now.  I had run `make check_dbg`/`make check_release` individually but didn't run `make check`

kmod:
Cool!

PR #284, opened 2015-02-05
add richards.py and deltablue.py minibenchmarks

toshok:
a couple more benchmarks from various python (and v8, smalltalk, etc) suites, which work but could be faster


kmod:
Cool!  Surprising we do so poorly on richards (looks like it might have something to do with the inline caches?).

PR #285, opened 2015-02-06
builtin functions

tjhance:
Implement the `builtin_function_or_method` type, which all the builtin functions are of (e.g., `sorted`).

Since we use the same calling convention for Python functions as our builtin C functions, I made a common base class, `BoxedFunctionBase` which has all the calling functionality, and two subclasses `BoxedFunction` and `BoxedBuilitinFunctionOrMethod`, each of which contributes the type, either `function_cls` or `builtin_function_or_method_cls`. (What a mouthful)

Difference is the attributes they have and stuff, e.g., `function_cls` is a descriptor which wraps itself in an `instancemethod`, and `builitin_function_or_method_cls` is not.

### DIFF
--- a/src/capi/types.h
+++ b/src/capi/types.h
@@ -83,7 +83,7 @@ public:
         return rtn;
     }
 
-    static Box* callInternal(BoxedFunction* func, CallRewriteArgs* rewrite_args, ArgPassSpec argspec, Box* arg1,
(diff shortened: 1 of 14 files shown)


kmod:
Cool! I'll take a look in the morning, but can we get a test?  Even just a quick check of the descriptor behavior would be great :)

tjhance:
Added a test case just checking that the builtin function type doesn't have the descriptor behavior of functions.

kmod:
awesome :)

PR #286, opened 2015-02-06
Remove -fPIC

undingen:
now that the flatten attribute is removed we don't have to compile pyston in PIC mode.


undingen:
I did not do much performance comparison but if someone is interested I will run the benchmarks. I tested raytace.py and it looked like a small performance improvement.

kmod:
Awesome!  Yeah it'd be great if we could have automatic perf testing in addition to the travis-ci testing.  We're almost done with the speed.pyston.org backfill, though actually I think in this case it wouldn't help because the perf tester is currently configured to run using the Make system.  Maybe we should try explicitly adding -fno-PIC to the make cxxflags?

PR #287, opened 2015-02-06
moved array.py to array_test.py

tjhance:
Naming this test case `array.py` means it tries to import itself instead of the `array` module, so I'm renaming it.


kmod:
Good catch :)

PR #288, opened 2015-02-07
Getset set

tjhance:
- Implement setting for getset descriptors
- Implement setting for `__name__` for types
- Reenable test case the `classname.py` test


kmod:
lgtm!  As I mentioned offline, we have some \_\_dict__ support, though I'm not sure it helps in this case if the \_\_name__ getset descriptor is on the type class since that won't be mutable.  Maybe we could add a getset descriptor to one of the classes in the descr_test extension module if we wanted to test this?

PR #292, opened 2015-02-09
implement setting __name__ for functions

tjhance:


src/runtime/types.cpp
```diff
@@ -304,6 +304,21 @@ extern "C" BoxedFunctionBase::BoxedFunctionBase(CLFunction* f, std::initializer_
...
+
+BoxedFunction::BoxedFunction(CLFunction* f, std::initializer_list<Box*> defaults, BoxedClosure* closure,
+                             bool isGenerator)
+    : BoxedFunctionBase(f, defaults, closure, isGenerator) {
+
+    // TODO eventually we want this to assert(f->source), I think, but there are still
+    // some builtin functions that are BoxedFunctions but really ought to be a type that
+    // we don't have yet.
+    if (f->source) {
+        this->name = static_cast<BoxedString*>(boxString(f->source->getName()));
```
kmod: does `name` get a default value if this doesn't get hit?  I think we may have to explicitly initialize it to NULL.

src/runtime/types.cpp
```diff
@@ -304,6 +304,21 @@ extern "C" BoxedFunctionBase::BoxedFunctionBase(CLFunction* f, std::initializer_
...
+    }
+}
+
 // This probably belongs in dict.cpp?
 extern "C" void functionGCHandler(GCVisitor* v, Box* b) {
     boxGCHandler(v, b);
```
kmod: We have to visit `name` if it is set here :)

kmod:
Hey Travis, this PR looks good -- could you make the small changes when you get a chance and push?  Not sure if you saw but Marius's #293 is loosely blocking on this.

tjhance:
OK I addressed the comments and fixed the GC issues.

So originally I made `name` a field in `BoxedFunction`. The problem with this is that it would then get initialized *after* the `BoxedFunctionBase` constructor got run. That means the `BoxedFunctionBase` constructor would run with garbage in `name`, but the gc has a chance of getting run in that constructor.

So, I could move all that logic out of the constructor, into a method and then call it at the end of the derived classes' constructors, which is ugly but probably necessary...

 but I also figured that I'd probably want to use `name` for the other types deriving from `BoxedFunctionBase`. So I just moved `name` to `BoxedFunctionBase` and initialize it to `NULL` at the beginning. Not really sure yet about this is where the `name` field really wants to be, though. And it also seems like the constructor problem is just going to come up again later anyway, so we're probably going to have to permute this stuff around eventually, *anyway*. But I'm putting this off for now.

Unrelatedly, I also realized that the gc visitor wasn't visiting `modname`, so I added that to the `functionGCHandler`.

kmod:
A similar fun thing is that when doing `new Foo(bar())`, it decomposes to
- call `operator new`
- call `bar`
- call Foo constructor

So for the duration of the bar() method, we have uninitialized GC memory sitting around for Foo.  I actually have no idea why we don't run into this any more, since the old way we had of addressing it (which was a hack) has been removed.

I'm not sure what the best long-term solution to these things are, but I feel like we can't be the only ones who have run into it...

tjhance:
wow, I had no idea that gets executed in that order... :|

tjhance:
In a lot of cases we have a wrapper around the constructor, like `boxString`, which forces the arguments to get evaluated before `new` is called. Maybe we should just do this for all our `BoxedFoo`s.

PR #293, opened 2015-02-10
import codec module

undingen:
This is not yet ready for merging (but ```"".encode("base64")``` is working) but I would like to get some feedback on how I have to implement ```__import__(,level=0)``` correctly. The import code is confusing me...

Notes:
- ```import optparse``` starts to fail because with this change we are able to import locale.py (which will use functools) and will set ```__name__``` so this depends on #292.
-  #292 brings it further but then it can't find ```__dict__```. haven't yet looked into what's missing.
- I did not yet check if ```PyString_Repr``` has the same behaviour
- I undefined unicode support
- pyaes.py works
- have to test more stuff

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,7 +290,7 @@ SRCS := $(MAIN_SRCS) $(STDLIB_SRCS)
 STDLIB_OBJS := stdlib.bc.o stdlib.stripped.bc.o
 STDLIB_RELEASE_OBJS := stdlib.release.bc.o
 
-STDMODULE_SRCS := errnomodule.c shamodule.c sha256module.c sha512module.c _math.c mathmodule.c md5.c md5module.c _randommodule.c _sre.c operator.c binascii.c pwdmodule.c posixmodule.c _struct.c datetimemodule.c _functoolsmodule.c _collectionsmodule.c itertoolsmodule.c resource.c signalmodule.c selectmodule.c fcntlmodule.c timemodule.c arraymodule.c $(EXTRA_STDMODULE_SRCS)
+STDMODULE_SRCS := errnomodule.c shamodule.c sha256module.c sha512module.c _math.c mathmodule.c md5.c md5module.c _randommodule.c _sre.c operator.c binascii.c pwdmodule.c posixmodule.c _struct.c datetimemodule.c _functoolsmodule.c _collectionsmodule.c itertoolsmodule.c resource.c signalmodule.c selectmodule.c fcntlmodule.c timemodule.c arraymodule.c _codecsmodule.c $(EXTRA_STDMODULE_SRCS)
 STDOBJECT_SRCS := structseq.c capsule.c stringobject.c $(EXTRA_STDOBJECT_SRCS)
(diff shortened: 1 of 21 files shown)


jvkersch:
This is going to be a huge help with the unicode module that I'm trying to port.

microbenchmarks/pyaes.py
```diff
@@ -0,0 +1,508 @@
+"""Simple AES cipher implementation in pure Python following PEP-272 API
```
kmod: Maybe we could put this file in minibenchmarks?  My thinking is that microbenchmarks can be for testing single language features (ex: how fast can we throw exceptions, or do function calls, etc), and that minibenchmarks can be for testing small programs.

kmod:
Cool!  I think it's ok to treat level==-1 and level==0 as the same for now -- it looks like the difference is that -1 means "try both absolute and relative imports" and 0 means "only absolute imports".  But we don't support relative imports so they are the same for us :)  We should add relative imports soon, but if they're not needed for codecs we don't need to worry about this for this PR.

By the way, what's your plan for `from_cpython/Python/codecs.c` and `src/capi/codecs.cpp`?

undingen:
ok that's why I couldn't find the relative import code...

I updated the patch and I think it's now ready for merging except that chaos.py will stop working because of the import optparse and we may want to wait?...
- removed the stray ```from_cpython/Python/codecs.c``` file
- added dummy unicode functions so I don't have to undef the macro
- supports ```str.decode("string-escape")```
- checked that PyString_Repr should have similar behaviour to the one in cpython
- moved the pyaes.py benchmark (we may want to increase the number of text it's encrypting?)
- improved the test
- set the ```optparse_test.py``` test to expect a fail

from_cpython/Modules/_codecsmodule.c
```diff
@@ -38,6 +38,7 @@ Copyright (c) Corporation for National Research Initiatives.
 #define PY_SSIZE_T_CLEAN
 #include "Python.h"
 
+
```
kmod: (unintentional?)

kmod:
Awesome!  looks good to me, but let's wait for Travis's func_set_name change so that we can keep optparse working.

undingen:
I saw that after #292 it's still failing because we haven't yet implemented ```function_cls.__dict__``` (inside ```getattrFunc```) and ```functools.py``` needs it.
I'm not sure how to implement it, should I just add another check like in ```getattr()```? or call ```getattr()``` instead of ```getattrInternal()```?

kmod:
Oh interesting... I think the "right" thing is to move the `__dict__` check from `getattr()` to `getattrInternal()`, since the idea was that `getattr` should just handle creating the rewriter and then hand off to `getattrInternal`.  It looks like there are a couple of places that call `getattrInternal`, so we would just have to verify that they're all ok with having the `__dict__` check happen (`dir()` looks the sketchiest).

Or we could just have `getattrFunc` call `getattr` instead of `getattrInternal` :)  We'll probably have to do some refactoring of the getattr stuff in the somewhat near future anyway :P

PR #295, opened 2015-02-11
add tp_getattr support, PyCFunction_NewEx, and most of PyNumber_Int's behavior

toshok:
the select module uses tp_getattr to implement the `poll` type, so add places where we need to call it.

PyCFunction_NewEx was also needed (the tp_getattr implementation called it.)

Finally, PyNumber_Int was needed to run the changes to select_test.py.

### DIFF
--- a/from_cpython/Modules/selectmodule.c
+++ b/from_cpython/Modules/selectmodule.c
@@ -1830,6 +1830,9 @@ initselect(void)
     {
 #endif
         Py_TYPE(&poll_Type) = &PyType_Type;
+        if (PyType_Ready(&poll_Type) < 0)
+          return;
+
         PyModule_AddIntConstant(m, "POLLIN", POLLIN);
         PyModule_AddIntConstant(m, "POLLPRI", POLLPRI);
         PyModule_AddIntConstant(m, "POLLOUT", POLLOUT);
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -22,6 +22,8 @@ namespace pyston {
 
 // FIXME duplicated with objmodel.cpp
 static const std::string _new_str("__new__");
+static const std::string _getattr_str("__getattr__");
+static const std::string _getattribute_str("__getattribute__");
 
 extern "C" void conservativeGCHandler(GCVisitor* v, Box* b) noexcept {
     v->visitPotentialRange((void* const*)b, (void* const*)((char*)b + b->cls->tp_basicsize));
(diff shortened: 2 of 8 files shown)


from_cpython/Modules/selectmodule.c
```diff
@@ -1830,6 +1830,9 @@ initselect(void)
     {
 #endif
         Py_TYPE(&poll_Type) = &PyType_Type;
+	if (PyType_Ready(&poll_Type) < 0)
+	  return;
```
kmod: Please use spaces not tabs :)

src/runtime/capi.cpp
```diff
@@ -314,6 +314,7 @@ extern "C" PyObject* PyObject_GetAttr(PyObject* o, PyObject* attr_name) noexcept
 }
 
 extern "C" PyObject* PyObject_GenericGetAttr(PyObject* o, PyObject* name) noexcept {
+    return getattr(o, static_cast<BoxedString*>(name)->s.c_str());
     Py_FatalError("unimplemented");
```
kmod: (delete?)

src/runtime/objmodel.cpp
```diff
@@ -490,6 +490,17 @@ BoxedDict* Box::getDict() {
...
+    if (cls->tp_getattr) {
+        if (rewrite_args)
+            REWRITE_ABORTED("");
+
+        Box* rtn = cls->tp_getattr(this, const_cast<char*>(attr.c_str()));
+        if (rtn == NULL)
+            raiseExcHelper(AttributeError, "%s", attr.c_str());
+        return rtn;
+    }
+
```
kmod: Do we need this in Box::getattr() as well?  I would have thought that by the time we get here we would have checked for a tp_getattr already...

src/runtime/objmodel.cpp
```diff
@@ -1449,6 +1460,13 @@ extern "C" Box* getattr(Box* obj, const char* attr) {
     static StatCounter slowpath_getattr("slowpath_getattr");
     slowpath_getattr.log();
 
+    if (obj->cls->tp_getattr) {
+        Box* rtn = obj->cls->tp_getattr(obj, const_cast<char*>(attr));
+        if (rtn == NULL)
+            raiseExcHelper(AttributeError, "%s", attr);
```
kmod: Should this be a call to `throwCAPIException()` instead?

from_cpython/Modules/selectmodule.c
toshok: sigh, emacs.  fixed.

src/runtime/capi.cpp
toshok: done

src/runtime/objmodel.cpp
toshok: after we spoke I agreed, but after removing it I found a reason to keep it:

```
#0  pyston::Box::getattr (this=0x12700aa318, attr="register", rewrite_args=0x0) at src/runtime/objmodel.cpp:495
#1  0x0000000000847ba9 in pyston::getattrInternalGeneral (obj=0x12700aa318, attr="register", rewrite_args=0x0, cls_only=false, for_call=true, bind_obj_out=0x7fffffffcf38, r_bind_obj_out=0x7fffffffcf30) at src/runtime/objmodel.cpp:1290
#2  0x00000000008467f3 in callattrInternal (obj=0x12700aa318, attr=0x21994b0, scope=pyston::CLASS_OR_INST, rewrite_args=0x0, argspec=..., arg1=0x1270030508, arg2=0x0, arg3=0x0, args=0x0, keyword_names=0x7fffffffd330) at src/runtime/objmodel.cpp:2072
#3  0x000000000084bc94 in callattr (obj=0x12700aa318, attr=0x21994b0, flags=..., argspec=..., arg1=0x1270030508, arg2=0x0, arg3=0x0, args=0x0, keyword_names=0x7fffffffd330) at src/runtime/objmodel.cpp:2248
```

I could instead move the tp_getattr handling to callattr.

src/runtime/objmodel.cpp
toshok: yup.  in the case of selectmodule, Py_FindMethod will have installed AttributeError as the thread's exception.  fixing.

src/runtime/objmodel.cpp
kmod: Oh, good point.  Maybe let's have a single check in `getattrInternalGeneral`, probably near the `__getattribute__` check?

src/capi/typeobject.cpp
```diff
@@ -623,6 +625,65 @@ static PyObject* slot_tp_iternext(PyObject* self) noexcept {
...
+    descrgetfunc f = Py_TYPE(attr)->tp_descr_get;
+
+    if (f != NULL) {
+        descr = f(attr, self, (PyObject*)(Py_TYPE(self)));
+        if (descr == NULL)
+            return NULL;
+        else
+            attr = descr;
+    }
+    res = runtimeCall(attr, ArgPassSpec(1, 0, true, true), name, NULL, NULL, NULL, NULL);
```
kmod: I think this function throws both C++ and CAPI-style exceptions (it propagates exceptions from the C++ `runtimeCall` and the CAPI `f`) -- we should pick one type of exception for it to throw, and have it convert the other one.  I'd bias towards CAPI-style since this looks like a CAPI function (uses CPython's naming conventions, type names, etc).

To make it explicit which style it is, I've been trying to add `noexcept` specifiers to any CAPI functions.  Can we add those in this PR as appropriate?

src/capi/typeobject.cpp
```diff
@@ -623,6 +625,65 @@ static PyObject* slot_tp_iternext(PyObject* self) noexcept {
...
+         method. So we use _PyType_Lookup and create the method only when
+         needed, with call_attribute. */
+    getattribute = typeLookup(self->cls, _getattribute_str, NULL);
+    try {
+        if (getattribute == NULL
+            || (Py_TYPE(getattribute) == wrapperdescr_cls
+                && ((BoxedWrapperDescriptor*)getattribute)->wrapped == (void*)PyObject_GenericGetAttr)) {
+            res = PyObject_GenericGetAttr(self, name);
+        } else {
+            res = call_attribute(self, getattribute, name);
```
kmod: If call_attribute uses Pyston-style exceptions, we should probably have the try-catch just around that, and make sure to do PyErr_Clear after the PyObject_GenericGetAttr case (which throws a CAPI exception).

src/capi/typeobject.cpp
```diff
@@ -623,6 +625,65 @@ static PyObject* slot_tp_iternext(PyObject* self) noexcept {
...
+        }
+    } catch (ExcInfo e) {
+    }
+    if (res == NULL) {
+        res = runtimeCall(getattr, ArgPassSpec(2, 0, true, true), self, name, NULL, NULL, NULL);
```
kmod: Similar note as above, runtimeCall throws Pyston exceptions, but slot_tp_getattr_hook is expected to throw CAPI exceptions, so we would need to convert here.  Adding a noexcept specifier is nice because it both annotates this and causes these kinds of mismatches to fail fast, since I think the compiler will convert any exceptions to std::terminate.

src/runtime/capi.cpp
```diff
@@ -314,7 +314,7 @@ extern "C" PyObject* PyObject_GetAttr(PyObject* o, PyObject* attr_name) noexcept
 }
 
 extern "C" PyObject* PyObject_GenericGetAttr(PyObject* o, PyObject* name) noexcept {
-    Py_FatalError("unimplemented");
+    return getattr(o, static_cast<BoxedString*>(name)->s.c_str());
```
kmod: (same thing about mixing exception handling)

src/runtime/objmodel.cpp
```diff
@@ -490,6 +490,17 @@ BoxedDict* Box::getDict() {
...
+
+    if (cls->tp_getattr) {
+        if (rewrite_args)
+            REWRITE_ABORTED("");
+
+        Box* rtn = cls->tp_getattr(this, const_cast<char*>(attr.c_str()));
+        if (rtn == NULL)
+            throwCAPIException();
+        return rtn;
+    }
```
kmod: I still don't think this should be here :/

src/capi/typeobject.cpp
toshok: Agreed, having this one raise a CAPI-style exception also cleans up the code further down.  Also will add `noexcept`

src/runtime/objmodel.cpp
toshok: removed locally and running `make check`

kmod:
Cool!

PR #296, opened 2015-02-12
Generator tracebacks

undingen:
- fix tacebacks inside a generator frame: we have to also unwind the calling stack frame.
I'm not really happy with the way I implemented it because it takes advantage of an implementation detail (that ```ucontext``` and ```unw_context``` are the same).
This will have to change when we move to a different stack switching technique (probably soon) :-(. 

- added a clear exc info instruction:
Even when an exception happened inside a generator after the next ```yield``` the ```exc_info()``` will return none or the last exception in the calling frame from where the generator got called. (I hope I observed the behavior correctly.)

### DIFF
--- a/src/analysis/type_analysis.cpp
+++ b/src/analysis/type_analysis.cpp
@@ -372,6 +372,7 @@ private:
                 return UNKNOWN;
             case AST_LangPrimitive::NONE:
             case AST_LangPrimitive::SET_EXC_INFO:
+            case AST_LangPrimitive::UNCACHE_EXC_INFO:
                 return NONE;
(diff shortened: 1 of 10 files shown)


kmod:
Looks good!  Can we call the operation something other than "CLEAR_EXC_INFO" -- I got confused for a sec since "clearing" is already used to mean "saying there is no exception" (such as `sys.exc_clear()`), whereas I think this new bytecode means "clear the frame-level exception cache".  Could we call it something like "UNCACHE_EXC_INFO"?

undingen:
I fixed the misnomer to UNCACHE_EXC_INFO.

PR #297, opened 2015-02-12
Implement str.join(iterable)

undingen:
This small change will be needed by ```locale``` which gets imported by ```optparse```.

I split it out into a separate pull request because I think this is a performance regression and I would like to discuss a idea I have: 
- I would like to extend ```pyElements()``` to not use the generic iterable protocol when we know the class.
This should give us the same speed as a specialized implementation for every known iterable type while still preserving the the shortness of just a ```for (Box* e : rhs->pyElements()) ```.

### DIFF
--- a/src/runtime/str.cpp
+++ b/src/runtime/str.cpp
@@ -1311,19 +1311,15 @@ Box* strIsTitle(BoxedString* self) {
 Box* strJoin(BoxedString* self, Box* rhs) {
     assert(self->cls == str_cls);
 
-    if (rhs->cls == list_cls) {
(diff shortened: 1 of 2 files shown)


kmod:
lgtm!  I think that's a good idea about trying to optimize pyElements() further, but I wonder if we will end up being constrained by the iteration interface (ie we would still have to do some amount of extra checking for every iteration of the loop).  We could go back to that old strategy you tried of having a foreach function that takes a function parameter, and maybe there's some crazy template solution that avoids any indirect function calls :P  We might want, though, to hold off on trying any of these things until we see an actual perf issue.

PR #300, opened 2015-02-13
add a simple_destructor for file objects

toshok:
tested in gdb by setting a breakpoint in fileDestructor and running this test:

```python
import gc

def foo():
  f = open("/dev/null")

foo()
gc.collect()
```


kmod:
Can we get a test?  like the one that you put in the comment -- or something like opening 10k files (and maybe calling gc.collect every 100 of them).

Also, formatting :)

toshok:
Yeah, I clearly need a pre-commit hook to check formatting :)

test/tests/file_destructor.py
```diff
@@ -0,0 +1,13 @@
+try:
```
kmod: Couple minor things as we wait for the travis-ci build to rerun:
- It's ok to count on the `gc` module being available
- let's use 4 spaces for indenting -- we try to follow [PEP8](https://www.python.org/dev/peps/pep-0008/#id12) for our Python code
- we could try calling `gc.collect()` every ~100 loop iterations to reduce that overhead

kmod:
Cool :)

PR #301, opened 2015-02-13
pyElements: Use specialiced iterators for known types

undingen:
Small gain for benchmarks which call functions which use ```pyElements()```
Would love to get rid of the shared_ptr though....
```
pyston (calibration)                      :    0.8s stock1: 0.8 (+0.8%)
pyston interp2.py                         :    5.9s stock1: 6.1 (-3.1%)
pyston raytrace.py                        :    7.5s stock1: 7.8 (-4.5%)
pyston nbody.py                           :   10.1s stock1: 9.9 (+1.5%)
pyston fannkuch.py                        :    6.7s stock1: 6.8 (-1.5%)
pyston chaos.py                           :   21.3s stock1: 22.6 (-5.8%)
pyston spectral_norm.py                   :   26.1s stock1: 27.9 (-6.7%)
pyston fasta.py                           :   17.7s stock1: 18.8 (-5.6%)
pyston pidigits.py                        :   16.1s stock1: 16.1 (+0.0%)
pyston richards.py                        :   10.2s stock1: 10.0 (+2.3%)
pyston deltablue.py                       :    2.4s stock1: 2.3 (+4.8%)
pyston (geomean-0b9f)                     :   10.1s stock1: 10.3 (-1.9%)
```


kmod:
Hmm, I like the idea, but I'm hoping we can find some cleaner way to get these gains.  It feels like it doesn't need to be so complicated to figure out how to iterate over the object -- CPython just uses the tp_iter and tp_iternext slots; maybe we could have some of our own?  (this is pretty close to what the runtime ics are trying to do here, but I guess that a hardcoded version could do better)  Or maybe there's some way to make this approach cleaner (templates?) ?

undingen:
I updated the patch:
- I'm now using a template to reduce the code duplication
- removed the iterators.h and moved everything into a single cpp
- -10% on spectral_norm

I'm still a somewhat ugly solution but I also tried to add a runtime IC to getiter() but this did not improve the performance much.

kmod:
I still feel like there's a cleaner solution out there, but this is a definite improvement for now :+1:

PR #302, opened 2015-02-16
New context switching code for generators

undingen:
This is a huge speed improvement for generators,
fasta.py takes 8secs now instead of 18secs


kmod:
Awesome!

PR #307, opened 2015-02-18
reuse generator stacks

toshok:
after bff1661, bm_ai.py crashed with an assert because we ran out of memory allocating generator stacks.

This change adds a simple destructor for BoxedGenerators, and some code to make use of the (already present in the file) available_addrs for some simple reuse.

As it stands, performance is pretty bad.  I lowered the number of iterators of the test to 1, and:

```sh
$ time python minibenchmarks/bm_ai.py

real	0m0.631s
user	0m0.625s
sys	0m0.006s

$ time ./pyston_release minibenchmarks/bm_ai.py
Warning: converting unicode literal to str
Warning: converting unicode literal to str
Warning: converting unicode literal to str
Warning: converting unicode literal to str

real	0m33.669s
user	0m24.300s
sys	0m9.367s
```

Clearly there's a lot going on with generators.  I merged #302 locally, and while it helped it wasn't a silver bullet like it is for fasta.py:

```sh
$ time ./pyston_release minibenchmarks/bm_ai.py
Warning: converting unicode literal to str
Warning: converting unicode literal to str
Warning: converting unicode literal to str
Warning: converting unicode literal to str

real	0m30.897s
user	0m22.525s
sys	0m8.371s
```

### DIFF
--- a/minibenchmarks/bm_ai.py
+++ b/minibenchmarks/bm_ai.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python
+
+"""Wrapper script for testing the performance of simple AI systems.
+
+bm_ai.py runs the following little solvers:
+    - N-Queens
+
+This used to contain an alphametics solver, but that was found to be bound
+primarily by eval() performance.
+"""
+
+# Wanted by the alphametics solver.
+from __future__ import division
+
+__author__ = "collinwinter@google.com (Collin Winter)"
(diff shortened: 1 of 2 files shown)


toshok:
`perf report` is showing a lot of time spent in syscalls (sigprocmask, it looks like), and a lot of unwinding.

PR #308, opened 2015-02-18
len(): Add special implementations for a few known types

undingen:
-15% for fasta.py


undingen:
closing because #309 implements a better approach...

PR #310, opened 2015-02-18
Rewriter: Optimize contant value loading

undingen:
- Reuse a register if it already contains the specified value
- Generate LEA when beneficial

I checked with the Intel® Architecture Code Analyzer that the generated code using ```LEA reg, reg+offset``` has the exact same performance as a ```mov reg, imm``` (checked Sandy Bridge and Haswell)

The main motivation for this patch is that I would like to generate smaller patchpoints.

With this patch:
```ASM
<+0x00ee>        48 b8 08 a0 00 70 12 00 00 00  movabs rax,0x127000a008
<+0x00f8>        48 3b 07                       cmp    rax,QWORD PTR [rdi]
<+0x00fb>        0f 85 ec 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x0101>        48 3b 06                       cmp    rax,QWORD PTR [rsi]
<+0x0104>        0f 85 e3 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x010a>        48 8b 0f                       mov    rcx,QWORD PTR [rdi]
<+0x010d>        4c 8d 80 00 e4 ff ff           lea    r8,[rax-0x1c00]
<+0x0114>        4c 3b 01                       cmp    r8,QWORD PTR [rcx]
<+0x0117>        0f 85 d0 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x011d>        4c 8d 88 00 7b ff ff           lea    r9,[rax-0x8500]
<+0x0124>        4c 3b 89 80 01 00 00           cmp    r9,QWORD PTR [rcx+0x180]
<+0x012b>        0f 85 bc 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x0131>        48 8b 89 88 01 00 00           mov    rcx,QWORD PTR [rcx+0x188]
<+0x0138>        48 8b 49 08                    mov    rcx,QWORD PTR [rcx+0x8]
<+0x013c>        4c 8d 90 00 14 00 00           lea    r10,[rax+0x1400]
<+0x0143>        4c 3b 11                       cmp    r10,QWORD PTR [rcx]
<+0x0146>        0f 85 a1 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x014c>        4c 8d 98 60 2e 01 00           lea    r11,[rax+0x12e60]
<+0x0153>        49 39 cb                       cmp    r11,rcx
<+0x0156>        0f 85 91 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x015c>        48 b9 3c 4d 32 02 00 00 00 00  movabs rcx,0x2324d3c
<+0x0166>        ff 01                          inc    DWORD PTR [rcx]
<+0x0168>        49 bb 10 10 88 00 00 00 00 00  movabs r11,0x881010
<+0x0172>        41 ff d3                       call   r11
<+0x0175>        48 b9 3c 4d 32 02 00 00 00 00  movabs rcx,0x2324d3c
<+0x017f>        ff 09                          dec    DWORD PTR [rcx]
<+0x0181>        e9 c6 07 00 00                 jmp    0x7ffff7fd094c <eval_A_e1_4+2380>
```

Without:
```ASM
<+0x00ee>        48 b8 08 a0 00 70 12 00 00 00  movabs rax,0x127000a008
<+0x00f8>        48 3b 07                       cmp    rax,QWORD PTR [rdi]
<+0x00fb>        0f 85 ec 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x0101>        48 b8 08 a0 00 70 12 00 00 00  movabs rax,0x127000a008
<+0x010b>        48 3b 06                       cmp    rax,QWORD PTR [rsi]
<+0x010e>        0f 85 d9 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x0114>        48 8b 07                       mov    rax,QWORD PTR [rdi]
<+0x0117>        48 b9 08 84 00 70 12 00 00 00  movabs rcx,0x1270008408
<+0x0121>        48 3b 08                       cmp    rcx,QWORD PTR [rax]
<+0x0124>        0f 85 c3 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x012a>        48 b9 08 1b 00 70 12 00 00 00  movabs rcx,0x1270001b08
<+0x0134>        48 3b 88 80 01 00 00           cmp    rcx,QWORD PTR [rax+0x180]
<+0x013b>        0f 85 ac 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x0141>        48 8b 80 88 01 00 00           mov    rax,QWORD PTR [rax+0x188]
<+0x0148>        48 8b 40 08                    mov    rax,QWORD PTR [rax+0x8]
<+0x014c>        48 b9 08 b4 00 70 12 00 00 00  movabs rcx,0x127000b408
<+0x0156>        48 3b 08                       cmp    rcx,QWORD PTR [rax]
<+0x0159>        0f 85 8e 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x015f>        48 b9 68 ce 01 70 12 00 00 00  movabs rcx,0x127001ce68
<+0x0169>        48 39 c1                       cmp    rcx,rax
<+0x016c>        0f 85 7b 01 00 00              jne    0x7ffff7fd02ed <eval_A_e1_4+749>
<+0x0172>        48 b8 ac 21 3e 02 00 00 00 00  movabs rax,0x23e21ac
<+0x017c>        ff 00                          inc    DWORD PTR [rax]
<+0x017e>        49 bb f0 02 88 00 00 00 00 00  movabs r11,0x8802f0
<+0x0188>        41 ff d3                       call   r11
<+0x018b>        48 b9 ac 21 3e 02 00 00 00 00  movabs rcx,0x23e21ac
<+0x0195>        ff 09                          dec    DWORD PTR [rcx]
<+0x0197>        e9 b0 07 00 00                 jmp    0x7ffff7fd094c <eval_A_e1_4+2380>
```

Looking for suggestions from some one more familiar with the rewriter how to improve this patch.

### DIFF
--- a/src/asm_writing/rewriter.cpp
+++ b/src/asm_writing/rewriter.cpp
@@ -135,12 +135,98 @@ void RewriterVar::addGuard(uint64_t val) {
     rewriter->addAction([=]() { rewriter->_addGuard(this, val); }, { this }, ActionType::GUARD);
 }
 
+Rewriter::ConstLoader::ConstLoader(Rewriter* rewriter) : rewriter(rewriter) {
+    invalidateAll();
+}
+
+bool Rewriter::ConstLoader::tryRegRegMove(uint64_t val, assembler::Register dst_reg) {
+    // copy the value if there is a register which contains already the value
+    bool found_value = false;
+    assembler::Register src_reg = findConst(val, found_value);
+    if (found_value) {
+        if (src_reg != dst_reg)
+            rewriter->assembler->mov(src_reg, dst_reg);
+        setKnownValue(dst_reg, val);
+        return true;
+    }
+    return false;
+}
+
+bool Rewriter::ConstLoader::tryLea(uint64_t val, assembler::Register dst_reg) {
+    // for large constants it maybe beneficial to create the value with a LEA from a known const value
+    if (isLargeConstant(val)) {
+        for (int reg_num = 0; reg_num < assembler::Register::numRegs(); ++reg_num) {
+            if (!hasKnownValue(assembler::Register(reg_num)))
+                continue;
+
+            int64_t offset = val - last_known_value[reg_num];
+            if (isLargeConstant(offset))
+                continue; // LEA can only handle small offsets
+
+            rewriter->assembler->lea(assembler::Indirect(assembler::Register(reg_num), offset), dst_reg);
+            setKnownValue(dst_reg, val);
+            return true;
+        }
+        // TODO: maybe add RIP relative LEA
+    }
+    return false;
+}
+
+void Rewriter::ConstLoader::moveImmediate(uint64_t val, assembler::Register dst_reg) {
+    // fallback use a normal: mov reg, imm
+    rewriter->assembler->mov(assembler::Immediate(val), dst_reg);
+    setKnownValue(dst_reg, val);
+}
+
+void Rewriter::ConstLoader::invalidateAll() {
+    for (int reg_num = 0; reg_num < assembler::Register::numRegs(); ++reg_num)
+        last_known_value[reg_num] = unknown_value;
+}
+
+assembler::Register Rewriter::ConstLoader::findConst(uint64_t val, bool& found_value) {
+    found_value = false;
+
+    if (unknown_value == val)
+        return assembler::Register(0);
+
+    for (int reg_num = 0; reg_num < assembler::Register::numRegs(); ++reg_num) {
+        if (last_known_value[reg_num] == val) {
+            found_value = true;
+            return assembler::Register(reg_num);
+        }
+    }
+
+    return assembler::Register(0);
+}
+
+void Rewriter::ConstLoader::loadConstIntoReg(uint64_t val, assembler::Register dst_reg) {
+    if (tryRegRegMove(val, dst_reg))
+        return;
+
+    if (tryLea(val, dst_reg))
+        return;
+
+    moveImmediate(val, dst_reg);
+}
+
+assembler::Register Rewriter::ConstLoader::loadConst(uint64_t val, Location otherThan) {
+    bool found_value = false;
+    assembler::Register reg = findConst(val, found_value);
+    if (found_value)
+        return reg;
+
+    reg = rewriter->allocReg(Location::any(), otherThan);
+    if (tryLea(val, reg))
+        return reg;
+
+    moveImmediate(val, reg);
+    return reg;
+}
+
 void Rewriter::_addGuard(RewriterVar* var, uint64_t val) {
     assembler::Register var_reg = var->getInReg();
     if (isLargeConstant(val)) {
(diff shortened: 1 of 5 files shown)


src/asm_writing/rewriter.cpp
```diff
@@ -135,12 +135,98 @@ void RewriterVar::addGuard(uint64_t val) {
...
+        return;
+
+    moveImmediate(val, dst_reg);
+}
+
+assembler::Register Rewriter::ConstLoader::loadConst(uint64_t val, Location otherThan) {
+    bool found_value = false;
+    assembler::Register reg = findConst(val, found_value);
+    if (found_value)
+        return reg;
```
kmod: This should probably assert that `reg != otherThan`

kmod:
@tjhance can you take a look too?

@undingen Do you think it'd be possible to use the existing location-tracking rather than tracking additional state?  I think we already track enough state in the RewriteVar->locations array, and if we don't, that was the intention.  ie we could try looking for vars that have multiple locations, one of which is a suitable constant and one of which is a register.  The invalidation code is delicate enough already that I worry about adding more types of invalidation that we have to remember to do :/

It looks like loadConst already does this tracking (adds a `Constant` location to the var), but the other sources of constant loading (comparing to 64-bit immediate) don't do that and I'm not sure how easy or not that would be.

tjhance:
Yeah I agree with @kmod that if possible, we should use the existing `locations` tracking mechanism rather than introduce a new one, which would require us to duplicate all the clobbering logic and so on.

Somewhat weirdly, a constant is a "location" under the current framework, which means you attach the constants to `RewriterVar`s when really you want to attach constants to registers--it looks like that's what @undingen is doing here.

There's not really a need to introduce a new invalidation framework--e.g. you don't need to call `invalidate` when you emit `call` instructions, because we already call `spillRegister` on the appropriate registers. We just need to make sure that `spillRegister` understands that `Location` can map to constants as well as `RewriterVar`s. 

Really, it would make the most sense to kill the concept of a constant as a `Location`. Right now, our way of saying that a `RewriterVar` is constant is to give it a constant location, but we could just have `is_constant` and `constant_value` fields instead. Then to track that a register has a certain constant value, we just add a constant-`RewriterVar` to the `locations_to_vars` map at that register. (Potential downside is that this would involve allocating a `RewriterVar` per constant, which might be a lot...)

kmod:
I think this was superseded by #617, so I'm closing it.

PR #311, opened 2015-02-19
Eval

tjhance:
v0 of `eval`

Can read global variables, but not local variables yet. It can't write either kind.

Should either get that stuff or working or add in some asserts so it will fail properly in those cases.

It was useful to add the `AST_Expression`, which is an alternative to `AST_Module` in the Official Python Grammar:

```
mod = Module(stmt* body)
	    | Interactive(stmt* body)
	    | Expression(expr body)
```

So there are, like, three node types that are variants of the name `expr` but that's just how the grammar is, I guess.


tjhance:
rebased and fixed the build

kmod:
lgtm, but can we split up the eval test so that the whole thing doesn't have to be marked with `expected: fail`?  Also the build is failing, I think due to a missed conflict (I added a new call to `parse()`).

tjhance:
OK, I commented out most of the `eval_test.py` which isn't working.

I also got it to read local variables.

tjhance:
okay, the build should be fixed now, although it could be cleaned up a bit

tjhance:
I modified `getLocals()` to include variables passed in from closures. This unlocks a few more achievements in `eval_test.py`.

kmod:
awesome!

PR #313, opened 2015-02-19
compvar: add int <op> float handling

undingen:
Convert the integer to a float and then let the float code handle the operation
With this change the type analysis is also able to comprehend that
e.g. '1 - <float>' will return a float

This means that the math operations in the 'linear_combination' function in chaos.py
get completely inlined.

-5% on chaos.py

```python
def linear_combination(self, other, l1, l2=None):
        if l2 is None:
            l2 = 1 - l1
        v = GVector(self.x * l1 + other.x * l2,
                    self.y * l1 + other.y * l2,
                    self.z * l1 + other.z * l2)
```

### DIFF
--- a/src/codegen/compvars.cpp
+++ b/src/codegen/compvars.cpp
@@ -797,6 +797,24 @@ public:
         }
         */
 
+        static std::vector<AbstractFunctionType::Sig*> sigs;
+        if (sigs.size() == 0) {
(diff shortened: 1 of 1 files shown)


kmod:
cool :)  I think we need to blacklist Is and IsNot though:
```
$ make run ARGS=-n
>> 1 is 1.0
True
```

Issue #294, opened 2015-02-11
Support del-unpacking


kmod:
I didn't even know this was a thing, but apparently this is perfectly valid:
```python
del [a, (b,) ([c.a], d[1])]
```
and it deletes a, b, c.a, and d[1].

(See [del_multi.py](https://github.com/dropbox/pyston/blob/master/test/tests/del_multi.py))

This should be a pretty simple change in the AST-to-CFG conversion to flatten these (see `cfg.cpp::visit_delete()`).

vinzenz:
Currently still failing:

>> del [a, (b,) ([c.a], d[1])]
../../src/core/cfg.cpp:1846: virtual bool pyston::CFGVisitor::visit_delete(pyston::AST_Delete *): Assertion `0' failed: Unsupported del target: 9


=> It most likely should fail on the fact that a tuple is not callable even before this assert pops up

kmod:
Oh, I think that was a typo in my example.

I think this is fixed now though; closing.

PR #317, opened 2015-02-21
Fixes the issue of del_unpacking #294

ssarangi:
Recursively unpack all the elements of List and Tuple to add new AST delete nodes and prevent addition of empty nodes coming from List & Tuple.


ssarangi:
I will update the test to be expected pass for this.

kmod:
"Unexpected success" is probably the best way to fail the tests :)  If you wouldn't mind, it'd be nice if you could squash the commits into a single commit while you're at it.  thanks for taking this!

ssarangi:
Absolutely... :) I will update the test and send a single commit..

ssarangi:
Closing this one since I sent a new pull request

PR #318, opened 2015-02-22
Add compvars and rewriting support for GET_ITER nodes 

undingen:
If the type analysis successfully determines that the __iter__() and __hasnext__()
methods we can now replace them with direct calls to the destination.
Else we create a patchpoint and do a rewrite.

- I'm not sure if there is a cleaner way todo this?
- I inverted the check inside ```BoxdedClass::hasGenericGetattr()``` because I think this is a bug. 
- inside the compvar code I'm now passing the ```CallattrFlags``` around instead of the ```bool clsonly```
- only the most common case gets rewriten. e.g. if the object does not implement "__iter__" I fallback and call  a helper function
``` 
pyston (calibration)                      :    0.8s baseline2: 0.8 (+2.7%)
pyston interp2.py                         :    3.4s baseline2: 3.7 (-8.9%)
pyston raytrace.py                        :    6.7s baseline2: 6.7 (-0.3%)
pyston nbody.py                           :    7.0s baseline2: 7.8 (-10.4%)
pyston fannkuch.py                        :    5.9s baseline2: 5.9 (-0.4%)
pyston chaos.py                           :   15.7s baseline2: 18.0 (-12.7%)
pyston spectral_norm.py                   :   16.9s baseline2: 16.9 (-0.0%)
pyston fasta.py                           :    5.8s baseline2: 5.4 (+5.8%)
pyston pidigits.py                        :    4.2s baseline2: 4.2 (-0.4%)
pyston richards.py                        :    1.4s baseline2: 1.5 (-5.0%)
pyston deltablue.py                       :    2.1s baseline2: 2.1 (+0.2%)
pyston (geomean-0b9f)                     :    5.3s baseline2: 5.5 (-3.4%)
```

### DIFF
--- a/src/analysis/type_analysis.cpp
+++ b/src/analysis/type_analysis.cpp
@@ -365,8 +365,9 @@ private:
                 return BOOL;
             case AST_LangPrimitive::LOCALS:
                 return DICT;
-            case AST_LangPrimitive::LANDINGPAD:
             case AST_LangPrimitive::GET_ITER:
+                return getType(node->args[0])->getPystonIterType();
+            case AST_LangPrimitive::LANDINGPAD:
             case AST_LangPrimitive::IMPORT_FROM:
             case AST_LangPrimitive::IMPORT_STAR:
             case AST_LangPrimitive::IMPORT_NAME:
--- a/src/codegen/compvars.cpp
+++ b/src/codegen/compvars.cpp
@@ -210,7 +210,7 @@ public:
                            const std::vector<CompilerVariable*>& args,
                            const std::vector<const std::string*>* keyword_names) override;
     CompilerVariable* callattr(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var,
-                               const std::string* attr, bool clsonly, ArgPassSpec argspec,
+                               const std::string* attr, CallattrFlags flags, ArgPassSpec argspec,
                                const std::vector<CompilerVariable*>& args,
                                const std::vector<const std::string*>* keyword_names) override;
     ConcreteCompilerVariable* nonzero(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var) override;
@@ -579,7 +579,7 @@ CompilerVariable* UnknownType::call(IREmitter& emitter, const OpInfo& info, Conc
 }
 
 CompilerVariable* UnknownType::callattr(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var,
-                                        const std::string* attr, bool clsonly, ArgPassSpec argspec,
+                                        const std::string* attr, CallattrFlags flags, ArgPassSpec argspec,
                                         const std::vector<CompilerVariable*>& args,
                                         const std::vector<const std::string*>* keyword_names) {
     bool pass_keywords = (argspec.num_keywords != 0);
@@ -599,10 +599,17 @@ CompilerVariable* UnknownType::callattr(IREmitter& emitter, const OpInfo& info, 
     else
         func = g.funcs.callattrN;
 
+    union {
+        CallattrFlags flags;
+        char value;
+    } flags_to_int;
+    static_assert(sizeof(CallattrFlags) == sizeof(char), "");
+    flags_to_int.flags = flags;
+
     std::vector<llvm::Value*> other_args;
     other_args.push_back(var->getValue());
     other_args.push_back(embedConstantPtr(attr, g.llvm_str_type_ptr));
-    other_args.push_back(getConstantInt(clsonly, g.i1));
+    other_args.push_back(getConstantInt(flags_to_int.value, g.i8));
 
     llvm::Value* llvm_argspec = llvm::ConstantInt::get(g.i32, argspec.asInt(), false);
     other_args.push_back(llvm_argspec);
@@ -819,11 +826,11 @@ public:
     }
 
     CompilerVariable* callattr(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var,
-                               const std::string* attr, bool clsonly, ArgPassSpec argspec,
+                               const std::string* attr, CallattrFlags flags, ArgPassSpec argspec,
                                const std::vector<CompilerVariable*>& args,
                                const std::vector<const std::string*>* keyword_names) override {
         ConcreteCompilerVariable* converted = var->makeConverted(emitter, BOXED_INT);
-        CompilerVariable* rtn = converted->callattr(emitter, info, attr, clsonly, argspec, args, keyword_names);
+        CompilerVariable* rtn = converted->callattr(emitter, info, attr, flags, argspec, args, keyword_names);
         converted->decvref(emitter);
         return rtn;
     }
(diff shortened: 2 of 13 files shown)


src/analysis/type_analysis.cpp
```diff
@@ -365,8 +365,22 @@ class BasicBlockTypePropagator : public ExprVisitor, public StmtVisitor {
...
+                const std::string& name = "__iter__";
+                CompilerType* attr_type = operand->getattrType(&name, true);
+                CompilerType* iter = attr_type->callType(ArgPassSpec(0), {}, NULL);
+                if (iter != UNKNOWN) {
+                    const std::string& has_next_str = "__hasnext__";
+                    CompilerType* attr_type = iter->getattrType(&has_next_str, true);
+                    if (attr_type == UNDEF)
+                        return UNKNOWN;
+                    return iter;
+                }
```
kmod: What do you think about adding a `getiterType()` function to CompilerType's?  It would at least let us move this functionality near the corresponding `getiter` implementation.

src/codegen/compvars.cpp
```diff
@@ -870,6 +921,13 @@ class IntType : public ConcreteCompilerType {
         return rtn;
     }
 
+    CompilerVariable* getiter(IREmitter& emitter, const OpInfo& info, VAR* var) override {
+        ConcreteCompilerVariable* converted = var->makeConverted(emitter, BOXED_INT);
+        CompilerVariable* rtn = converted->getiter(emitter, info);
+        converted->decvref(emitter);
+        return rtn;
```
kmod: This kind of "defer to boxed/UNKNOWN" pattern is pretty common and is responsible for a lot of the boilerplate in this file :/  It's not important enough to be worth changing code once it's written, but I was happy with my experiment of having a default implementation that explicitly defers to UNKNOWN; see `compvars.h::_ValuedCompilerType<V>::unpack()`.

src/codegen/irgen/irgenerator.cpp
```diff
@@ -487,24 +490,12 @@ class IRGeneratorImpl : public IRGenerator {
...
-
-                ConcreteCompilerVariable* converted_obj = obj->makeConverted(emitter, obj->getBoxType());
+                auto rtn = obj->getiter(emitter, getOpInfoForNode(node, unw_info));
                 obj->decvref(emitter);
-
-                llvm::Value* v = emitter.createCall(unw_info, g.funcs.getPystonIter, { converted_obj->getValue() });
-                assert(v->getType() == g.llvm_value_type_ptr);
-
-                return new ConcreteCompilerVariable(UNKNOWN, v, true);
+                curblock = emitter.getBuilder()->GetInsertBlock();
```
kmod: Hmm so this call worries me because if we wanted to go with this approach (copy the curblock from emitter to curblock), we would need this line a lot more places.

I think the (hacky) current way of trying to address this is that the emitter has a reference to the curblock, so we should use that.  It looks like it only uses it internally, but we could expose that through some sort of appropriate API, and try to not have multiple places that can get out of sync.

src/codegen/compvars.cpp
undingen: Yes, this is really causing a lot of boilerplate code :-(
If I convert to unknown wouldn't I then loose the type info? --> getType()->getattrType always would return unknown?

src/codegen/compvars.cpp
```diff
@@ -330,6 +330,57 @@ class UnknownType : public ConcreteCompilerType {
         return new ConcreteCompilerVariable(UNKNOWN, rtn, true);
     }
 
+    CompilerVariable* getiter(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var) override {
+        static const std::string hasnext_str("__hasnext__");
+        static const std::string iter_str = "__iter__";
+
+        llvm::AllocaInst* rtn_value_memory = emitter.getBuilder()->CreateAlloca(g.llvm_value_type_ptr);
```
kmod: Rather than doing an alloca+store+load, we should use a PHI node (grep for "PHI").  Also, if we want stack memory, we should place the alloca in the function entry block so that it's a static alloca (becomes part of the frame's stack size) rather than put it right with where it's used, since if this is in a for loop we will alloca additional stack space for every iteration.

src/runtime/types.h
```diff
@@ -226,7 +226,7 @@ class BoxedClass : public BoxVar {
     // that we can't rely on for extension classes.
     bool is_pyston_class;
 
-    bool hasGenericGetattr() { return tp_getattr != NULL; }
+    bool hasGenericGetattr() { return tp_getattr == NULL; }
```
kmod: good catch

src/codegen/compvars.cpp
```diff
@@ -330,6 +330,57 @@ class UnknownType : public ConcreteCompilerType {
...
+
+        CallattrFlags flags = {.cls_only = true, .null_on_nonexistent = true };
+        CompilerVariable* iter_call = var->callattr(emitter, info, &iter_str, flags, ArgPassSpec(0), {}, 0);
```
kmod: What do you think about adding something like a `hasattr` CompilerType method (returning yes/no/maybe)?  This could let us skip the dynamic checks for known types.  This could also replace the `hasnext != UNKNOWN && hasnext != UNDEF` -- I think that's equivalent to `hasattr() == YES` but the hasattr version feels more reliable and less likely to break subtly.

Also if we wanted, we could then simplify this function to
```
if (this->hasattr("__iter__") == YES && this->getattrType("__iter__")->callType()->hasattr("__hasnext__") == YES)
    return this->call("__iter__");
return emitter.call(g.funcs.getiter, value);
```
which I think still handles the common cases, though I'm not sure how it otherwise compares to what you've already written.

src/codegen/compvars.cpp
kmod: Oh, yeah you can convert to getBoxType() and then call UNKNOWN->getiter on it.

undingen:
I updated the pull request to address your comments.
- added ```hasattr()``` and ```getPystonIterType()```. (I put Pyston in the name because I want to emphasize that it checks for ```__hasnext__```)
- changed the alloca to a phi instruction
- exposed current basic block
- removed the duplication of ```getiter()``` implementations
- rebased to head

kmod:
Hey Marius, my GCC is giving errors about the changes to CallattrFlags -- I guess it doesn't like using designated initializers when there are unions involved.  Do you have a different GCC hooked in?

undingen:
oh that makes sense, I was already wondering why the travis build is failing...
I will fix it shortly

kmod:
I'm not sure why the travis build is failing; we saw some spurious failures last night and I reran some builds which fixed it, but this one doesn't seem to want to start up again (it just stays "scheduled").  Also I think the travis build only tests under clang.

undingen:
Ok i fixed the gcc build: I moved the union to ```UnknownType::callattr``` and also fixed a problem where I was passing the ```CallattrFlags``` as ```i32``` but looking at the IR generated for ```pyston::callattr``` I noticed that it is expecting a ```i8```.

PR #320, opened 2015-02-23
__name__ for builtin functions

tjhance:


kmod:
RIP "`<function at 0x651234>() takes exactly 2 parameters`"

PR #323, opened 2015-02-24
Fix the delete multiitem issue #294

ssarangi:


ssarangi:
ignore this pull request again.. I should really look and get the right commit.. Sorry :(

undingen:
You can also edit your pull request by making the the fix in a new commit and then squashing the commits together. When you then (force) push to the same branch github will auto update the pull request so that you don't have to create a new pull request. I hope I did not misunderstand the reason why you created a new pull request..

ssarangi:
No you are right.. I didn't put the right status in the py file and Travis failed. So that's why I didn't want to do another commit and send a pull request with 2 of them as Kevin already asked me to squash the commits. Let me see how to do it correctly :)

undingen:
I think you just have to remove the ```# expected: pass``` line.
- run ```make check```
- commit the change
- ```git rebase -i HEAD~2``` (2 is the number of last commits you wan't to modify)
- change ```pick``` in front of the last commit you want to get merged with the one above to ```squash```
- ```git push origin +master```
But be careful I'm also a quite a git newbie... :-D

ssarangi:
Thanks undingen.. Looks like it still showed up with another commit :( Let me just do it from scratch and fix this.

PR #325, opened 2015-02-24
Rebase llvm to r230300

undingen:
Had to change the comdat handling otherwise the executable would crash on startup.
Cause is that publicize is renaming symbols but keeps the old comdat around,
the linker would then replace a call to the symbol with a call to 0.

One may need to run 'make llvm_up' if you are using the cmake build and see a error,
caused by debuginfo beeing renamed to debuginfodwarf.


kmod:
thanks!

PR #329, opened 2015-02-26
Implement delattr()

undingen:


kmod:
cool!

we seem to be having some lingering issues with Travis-CI, but I tested this manually and it passes the tests.

PR #331, opened 2015-02-26
Add sys.getdefaultencoding and sys.getfilesystemencoding

undingen:
I hard coded the file system encoding to be UTF-8 on non windows systems.
Initially I replicated the behavior of CPython and imported the functions but then I thought it's overkill.
As far as I know every linux system we care about uses UTF-8 for file names. The CPython code retrieves on startup with ```nl_langinfo()``` the encoding and then tries to load the corresponding encoder (--> we would run a lot of additional python code shortly after ```setupRuntime()```.
Did I miss something?


kmod:
Yeah, we'll want to eventually support the `LANG` or whatever environment variables, but I think this is great for now :)

PR #332, opened 2015-02-26
Weakrefs

toshok:
Pull in cpython's _weakref module and relevant code.  Also teach our GC to handle them.

Weakrefs are opt-in by type, and the only types currently opting in are functions and instancemethods.  we still fail `import WeakSet` as user types aren't opting in.  I'll add that support (and support for other builtin types) in a further PR.

### DIFF
--- a/from_cpython/Include/weakrefobject.h
+++ b/from_cpython/Include/weakrefobject.h
@@ -1,0 +1,90 @@
+/* Weak references objects for Python. */
+
+#ifndef Py_WEAKREFOBJECT_H
+#define Py_WEAKREFOBJECT_H
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+typedef struct _PyWeakReference PyWeakReference;
+
+/* PyWeakReference is the base struct for the Python ReferenceType, ProxyType,
+ * and CallableProxyType.
+ */
+struct _PyWeakReference {
+    PyObject_HEAD
+
+    /* The object to which this is a weak reference, or Py_None if none.
+     * Note that this is a stealth reference:  wr_object's refcount is
+     * not incremented to reflect this pointer.
+     */
+    PyObject *wr_object;
+
+    /* A callable to invoke when wr_object dies, or NULL if none. */
+    PyObject *wr_callback;
+
+    /* A cache for wr_object's hash code.  As usual for hashes, this is -1
+     * if the hash code isn't known yet.
+     */
+    long hash;
+
+    /* If wr_object is weakly referenced, wr_object has a doubly-linked NULL-
+     * terminated list of weak references to it.  These are the list pointers.
+     * If wr_object goes away, wr_object is set to Py_None, and these pointers
+     * have no meaning then.
+     */
+    PyWeakReference *wr_prev;
+    PyWeakReference *wr_next;
+};
+
+PyAPI_DATA(PyTypeObject) _PyWeakref_RefType;
+PyAPI_DATA(PyTypeObject) _PyWeakref_ProxyType;
+PyAPI_DATA(PyTypeObject) _PyWeakref_CallableProxyType;
+
+#define PyWeakref_CheckRef(op) PyObject_TypeCheck(op, &_PyWeakref_RefType)
+#define PyWeakref_CheckRefExact(op) \
+        (Py_TYPE(op) == &_PyWeakref_RefType)
+#define PyWeakref_CheckProxy(op) \
+        ((Py_TYPE(op) == &_PyWeakref_ProxyType) || \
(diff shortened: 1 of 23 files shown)


from_cpython/Include/weakrefobject.h
```diff
@@ -0,0 +1,90 @@
...
+#define PyWeakref_CheckProxy(op) \
+        ((Py_TYPE(op) == &_PyWeakref_ProxyType) || \
+         (Py_TYPE(op) == &_PyWeakref_CallableProxyType))
+
+#define PyWeakref_Check(op) \
+        (PyWeakref_CheckRef(op) || PyWeakref_CheckProxy(op))
+
+
+PyAPI_FUNC(PyObject *) PyWeakref_NewRef(PyObject *ob,
+                                              PyObject *callback);
```
kmod: Minor, but we've been adding PYSTON_NOEXCEPT to the end of all the C API function declarations -- this resolves to "noexcept" in C++ mode (and to nothing in C mode), and is a nice safeguard + documentation that these are C API functions.

from_cpython/Include/weakrefobject.h~
```diff
@@ -0,0 +1,83 @@
+/* Weak references objects for Python. */
```
kmod: Maybe let's add `*~` to our .gitignore :P

toshok: haha.  wow I can't believe I did that.  :)

On Thu, Feb 26, 2015 at 4:04 PM, Kevin Modzelewski <notifications@github.com

from_cpython/Objects/weakrefobject.c
```diff
@@ -0,0 +1,1000 @@
...
+                if (restore_error)
+                    PyErr_Fetch(&err_type, &err_value, &err_tb);
+                return;
+            }
+
+            for (i = 0; i < count; ++i) {
+                PyWeakReference *next = current->wr_next;
+
+                // Pyston change:
+                //if (current->ob_refcnt > 0)
```
kmod: Hmm you have this conditional resolving the opposite way that the refcnt ones normally do; if this is right, can we add a comment explaining why this one is different?

src/gc/collector.cpp
```diff
@@ -299,6 +301,9 @@ void disableGC() {
     gc_enabled = false;
 }
 
+extern "C" Box* runtimeCall(Box* obj, ArgPassSpec argspec, Box* arg1, Box* arg2, Box* arg3, Box** args,
+                            const std::vector<const std::string*>* keyword_names);
```
kmod: minor, but we can probably just get this from objmodel.h

src/gc/gc_alloc.h
```diff
@@ -77,7 +77,7 @@ extern "C" inline void* gc_alloc(size_t bytes, GCKind kind_id) {
 
     // TODO This doesn't belong here (probably in PythonGCObject?)...
     if (kind_id == GCKind::PYTHON) {
-        ((Box*)r)->cls = NULL;
+        memset(r, 0, bytes);
```
kmod: Hmm I was hoping that we could avoid doing this.  I think in some places we explicitly zero out any extension-defined memory; is that something we could do here?

toshok: Yeah, I can get around it by explicitly null-initializing the weakref list in builtin types that opt-in.

from_cpython/Objects/weakrefobject.c
toshok: Ah yeah, I meant to go back and address this.  Problem is we don't have a reliable refcount.  That said, we shouldn't be calling this function at all from within pyston.

toshok:
Okay, everything should be addressed now.  let me know if things look okay and I'll squash things before you merge.

toshok:
odd.  build works locally and all test pass.

kmod:
You have to update `from_cpython/CMakeLists.txt` so that the new files get added to the cmake build as well.  maybe it's time for us to just switch :)

from_cpython/Modules/_weakref.c
```diff
@@ -96,6 +96,13 @@ init_weakref(void)
     m = Py_InitModule3("_weakref", weakref_functions,
                        "Weak-reference support module.");
     if (m != NULL) {
+        if (PyType_Ready(&_PyWeakref_RefType) < 0)
```
kmod: tiny, but let's add a note that this is a Pyston change.

kmod:
lgtm!

toshok:
136c3b2 snuck in there by mistake.  going to pull it out when I squash and push it.  will open a PR separately for that one, since it's a pretty large (but mechanical) change.

kmod:
cool!

PR #334, opened 2015-02-27
Implement vararg map()

undingen:
in addition I added a small helper function for unpacking our function call argument array.
We could also call ```std::tie(arg1, arg2, arg3, args) =  getTupleFromArgsArray(a, size)``` instead of the ```std::arg<>``` calls.

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -450,11 +450,9 @@ Value ASTInterpreter::visit_jump(AST_Jump* node) {
             }
 
             CompiledFunction* partial_func = compilePartialFuncInternal(&exit);
-            Box* arg1 = arg_array.size() >= 1 ? arg_array[0] : 0;
-            Box* arg2 = arg_array.size() >= 2 ? arg_array[1] : 0;
-            Box* arg3 = arg_array.size() >= 3 ? arg_array[2] : 0;
-            Box** args = arg_array.size() >= 4 ? &arg_array[3] : 0;
-            return partial_func->call(arg1, arg2, arg3, args);
+            auto arg_tuple = getTupleFromArgsArray(&arg_array[0], arg_array.size());
+            return partial_func->call(std::get<0>(arg_tuple), std::get<1>(arg_tuple), std::get<2>(arg_tuple),
+                                      std::get<3>(arg_tuple));
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -559,6 +559,53 @@ Box* map2(Box* f, Box* container) {
...
+
+        if (num_done == num_iterable)
+            break;
+
+        auto v = getTupleFromArgsArray(&current_val[0], num_iterable);
+        listAppendInternal(rtn, runtimeCall(f, ArgPassSpec(num_iterable), std::get<0>(v), std::get<1>(v),
+                                            std::get<2>(v), std::get<3>(v), NULL));
+
+        for (int i = 0; i < num_iterable; ++i) {
+            ++args_it[i];
```
kmod: should we only call this for iterators that haven't finished yet?

undingen:
Good catch

src/codegen/ast_interpreter.cpp
```diff
@@ -450,11 +450,9 @@ Value ASTInterpreter::visit_jump(AST_Jump* node) {
             }
 
             CompiledFunction* partial_func = compilePartialFuncInternal(&exit);
-            Box* arg1 = arg_array.size() >= 1 ? arg_array[0] : 0;
-            Box* arg2 = arg_array.size() >= 2 ? arg_array[1] : 0;
-            Box* arg3 = arg_array.size() >= 3 ? arg_array[2] : 0;
-            Box** args = arg_array.size() >= 4 ? &arg_array[3] : 0;
-            return partial_func->call(arg1, arg2, arg3, args);
+            auto arg_tuple = getTupleFromArgsArray(&arg_array[0], arg_array.size());
+            return partial_func->call(std::get<0>(arg_tuple), std::get<1>(arg_tuple), std::get<2>(arg_tuple),
```
kmod: Ok I wasted some time trying to figure out how to do this with templates :/  We have some template helpers that I thought would help in thread_utils.h, but honestly I can't say I understand when tuples can be turned into parameter packs or not....

If you feel interested, take a look here:
http://stackoverflow.com/questions/7858817/unpacking-a-tuple-to-call-a-matching-function-pointer

undingen:
I prevent increments now when already at the end

The code in thread_utils.h is very clever - took me a while to figure out how this is working...
But I have mixed feelings and would like to keep it as is.

PR #335, opened 2015-02-27
Implement str.replace maxreplaces

undingen:


undingen:
and sneaked in bool support for ```PyInt_AsLong```

kmod:
:+1:

PR #336, opened 2015-02-27
tester.py: canonicalize_stderr: add substitution for object.__new__() error

rntz:


kmod:
Cool!  A small change but should improve the getting-started process.

PR #338, opened 2015-02-28
add weakref support for user defined subtypes (with inheritance) and 'type' instances

toshok:
All user-defined subclasses can be weakly referenced after this change, which is slightly more permissible than [cpython's implementation](https://docs.python.org/2/library/weakref.html).

There are a few holes, though, as the only builtin types that provide for weak references at present are type objects, functions, and instance methods.  We fail loudly (_weakref throws an exception) when an attempt is made to weakly reference something we don't support.

Subtypes of types that define a `tp_weaklistoffset` inherit their parent's (so we don't have multiple weaklists).  Also, if there is no `tp_weaklistoffset` and it's a user defined type, we inflate the object by `sizeof(Box**)` and set its `tp_weaklistoffset` appropriately.

With this change we can successfully `import numbers`.

I also added some more weakref tests, but they fail currently.

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -730,7 +730,8 @@ static BoxedClass* makeBuiltinException(BoxedClass* base, const char* name, int 
     if (size == 0)
         size = base->tp_basicsize;
 
-    BoxedClass* cls = BoxedHeapClass::create(type_cls, base, NULL, offsetof(BoxedException, attrs), size, false, name);
+    BoxedClass* cls
+        = BoxedHeapClass::create(type_cls, base, NULL, offsetof(BoxedException, attrs), 0, size, false, name);
     cls->giveAttr("__module__", boxStrConstant("exceptions"));
 
     if (base == object_cls) {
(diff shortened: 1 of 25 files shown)


toshok:
This does have a side effect of inflating objects by 8 bytes across the board for user-defined types, which is kinda painful.  We could potentially do something like using an `unordered_map<Box*, Box*>` to look up the weak lists, and tp_weaklistoffset would just be a boolean (0 = no weak refs possible, anything else = weak refs possible), but that's a much more invasive change, and requires we write our own weakref impl instead of reusing cpython's.

toshok:
This should be good to go.  Since the last update I've added support for more builtin types, so the following types support WR's:

1. set
2. frozenset
3. generator
4. type
5. file
6. array.array
7. re pattern
8. all user-defined classes
8. subclasses of builtin types (including long/tuple, which cpython disallows.)

I modified `test/tests/weakrefs.py` to enumerate all the types called out in the cpython docs.  we're missing sockets and DBCursors, but we'll still fail loudly (throw an exception) for them.

Also, I noticed that the huge arena was using the same starting address as the generator stack area, and fixed that.

kmod:
Cool!  sorry for not seeing this sooner

PR #340, opened 2015-03-02
add tests for "[] = []", "2 = 2"; currently we core dump on these

rntz:


rntz:
Hey @kmod, this should be mergeable now. The "2 = 2" test has expected: fail, and I've fixed the "[] = []" problem.

PR #341, opened 2015-03-02
Add cStringIO module + basic support for setting member descriptors

undingen:
- the cStringIO module needs support for setting member descriptors
- fixed a bug: when adding the ```readonly``` field I noticed that the get member descriptor rewriting code generates a 64bit mem comparison for the 32bit offset value, thus comparing partially undefined memory.


kmod:
Cool! :star:

PR #343, opened 2015-03-02
Implement `pyston -c command`

rntz:


test/tests/dash_c.py
```diff
@@ -0,0 +1,18 @@
...
+
+me = sys.executable
+
+with open('/dev/null')as ignore:
+    # We don't (yet?) require exact stderr or return code compatibility w/
+    # python. So we just check that we succeed or fail as appropriate.
+    def do(args):
+        print 0 == subprocess.call([me] + args, stderr=ignore)
+
+    run(["-c", "print 2 + 2"])
```
kmod: I assume you meant `do()` (or to call the function `run()`).  Maybe we could add a tester annotation that says that a script is allowed to have a nonzero exit code?  Currently things will pass if CPython and Pyston both fail, which they do here but is unexpected.

rntz: Oops! Yeah, having an annotation for "this file is expected to return nonzero" might be nice. I'll try implementing that.

kmod: Yeah, and I think most of the cases where we have the test crash are from before we supported exception handling, so crashing the program was the only way we could test if the exception got thrown.  Now that we support real exception handling, we could go back and convert most of the crashing tests to just print out what exception they threw, like we do for new tests.

Also, just a logistical thing but I don't get notifications when you update a PR, only when you comment on it, so if you fix things adding a comment to that effect can make sure I see it.  I'm going to try the "assign myself" thing and see if that changes it.

PR #344, opened 2015-03-03
Add should_error directive, for tests that should exit non-0

rntz:
I've added a "# should_error" test directive, for tests which should produce a non-0 exit code. This caught a test that was crashing and wasn't intended to be: `data_descriptors.py` had a variable name typo, `c.dd` instead of `C.dd`.

An unresolved issue before this can be merged: I don't know whether `resource_test.py` is supposed to crash or not. It succeeds when I run it manually or `make run_resource_test` it, but crashes (in both Python & Pyston) when run under tester.py.

The should_error directive is necessary b/c I couldn't change all tests to catch-and-print. There are two reasons I couldn't do this:

1. Some SyntaxErrors can't be caught b/c they happen at parse time
   e.g. continue_outside_loop, return_outside_func, future_non_existent
   (don't see any way to fix this)

2. canonicalize_stderr doesn't work on printed-out exceptions
   e.g. object_new_arguments, class_noctor
   (this could be fixed, but it's simpler to leave it as-is)

I've changed most of the tests that don't fall into one of the above two cases to explicitly catch errors. Some I wasn't sure about (e.g. exceptions_nonreprable.py, init_must_return_none.py, sys_excinfo.py), so I left them.

### DIFF
--- a/src/runtime/import.cpp
+++ b/src/runtime/import.cpp
@@ -149,6 +149,7 @@ static Box* importSub(const std::string& name, const std::string& full_name, Box
         return module;
     }
 
+    // TODO: these are a crude hack to help our tests, find a better way
     if (name == "basic_test")
         return importTestExtension("basic_test");
     if (name == "descr_test")
--- a/test/tests/21.py
+++ b/test/tests/21.py
@@ -1,3 +1,5 @@
+# should_error
+
 # Expected failure: even though we could copy the "definedness" information in the assignment,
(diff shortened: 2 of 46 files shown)


kmod:
Looks great, and I'm glad this found a few actual bad tests!  We can try over time to convert more of them to be non-crashing tests (ex we can catch the SyntaxErrors once we have exec), but I think this is a great step for now.

PR #345, opened 2015-03-03
runtime/inline add header dependencies

dagar:


kmod:
Cool, thanks for taking care of this :)  I just modified some of the header files so I was running into this a bunch.

PR #348, opened 2015-03-04
Add the io/_io Module and make import gzip work

undingen:
This adds quite a few C API functions
and I had to introduce at least two temporary hacks:
- do not depend on the ```ABC``` module which uses weak references
- use a dummy ```PyErr_CheckSignals()``` function


from_cpython/CMakeLists.txt
```diff
@@ -15,10 +15,10 @@ endforeach(STDLIB_FILE)
 add_custom_target(copy_stdlib ALL DEPENDS ${STDLIB_TARGETS})
 
 # compile specified files in from_cpython/Modules
-file(GLOB_RECURSE STDMODULE_SRCS Modules errnomodule.c shamodule.c sha256module.c sha512module.c _math.c mathmodule.c md5.c md5module.c _randommodule.c _sre.c operator.c binascii.c pwdmodule.c posixmodule.c _struct.c datetimemodule.c _functoolsmodule.c _collectionsmodule.c itertoolsmodule.c resource.c signalmodule.c selectmodule.c fcntlmodule.c timemodule.c arraymodule.c zlibmodule.c _codecsmodule.c socketmodule.c unicodedata.c _weakref.c cStringIO.c)
+file(GLOB_RECURSE STDMODULE_SRCS Modules errnomodule.c shamodule.c sha256module.c sha512module.c _math.c mathmodule.c md5.c md5module.c _randommodule.c _sre.c operator.c binascii.c pwdmodule.c posixmodule.c _struct.c datetimemodule.c _functoolsmodule.c _collectionsmodule.c itertoolsmodule.c resource.c signalmodule.c selectmodule.c fcntlmodule.c timemodule.c arraymodule.c zlibmodule.c _codecsmodule.c socketmodule.c unicodedata.c _weakref.c cStringIO.c _io/bufferedio.c _io/bytesio.c _io/fileio.c _io/iobase.c _io/_iomodule.c _io/stringio.c _io/textio.c)
```
kmod: Hmm GLOB_RECURSE seems to not like any of these patterns that have a slash in them, so the cmake build doesn't try to actually use any of the _io/*.c files.  Maybe we can replace the glob with something else?  This seems related:
http://www.cmake.org/pipermail/cmake/2010-March/036061.html

from_cpython/Include/memoryobject.h
```diff
@@ -0,0 +1,76 @@
...
+
+/* Get a pointer to the underlying Py_buffer of a memoryview object. */
+#define PyMemoryView_GET_BUFFER(op) (&((PyMemoryViewObject *)(op))->view)
+/* Get a pointer to the PyObject from which originates a memoryview object. */
+#define PyMemoryView_GET_BASE(op) (((PyMemoryViewObject *)(op))->view.obj)
+
+
+PyAPI_FUNC(PyObject *) PyMemoryView_GetContiguous(PyObject *base, 
+						  int buffertype, 
+						  char fort);
```
kmod: Can we add PYSTON_NOEXCEPT to these?

from_cpython/Lib/io.py
```diff
@@ -79,6 +79,8 @@ class BufferedIOBase(_io._BufferedIOBase, IOBase):
 class TextIOBase(_io._TextIOBase, IOBase):
     __doc__ = _io._TextIOBase.__doc__
 
+# Pyston change: tmp hack until we support weak refs
```
kmod: With Chris's PR I just merged, can we remove these workarounds?

from_cpython/Modules/signalmodule.c
```diff
@@ -893,7 +893,7 @@ int
 PyErr_CheckSignals(void)
 {
     // Pyston change:
-    Py_FatalError("TODO");
+    printf("Warning: PyErr_CheckSignals is not yet implemented!\n");
```
kmod: I think it currently doesn't really matter if we don't do anything in this function, since the only place that can actually set a signal will Py_FatalError().  Can we keep the Py_FatalError() call here too, but move it below the is_tripped check?  that way it doesn't have to be a warning and it will fail more loudly if somehow we do end up setting a signal that should get checked.

src/runtime/capi.cpp
```diff
@@ -86,8 +86,36 @@ Box* BoxedWrapperDescriptor::__get__(BoxedWrapperDescriptor* self, Box* inst, Bo
...
+    if ((*pb->bf_getsegcount)(obj, NULL) != 1) {
+        PyErr_SetString(PyExc_TypeError, "expected a single-segment buffer object");
+        return -1;
+    }
+    len = (*pb->bf_getcharbuffer)(obj, 0, &pp);
+    if (len < 0)
+        return -1;
+    *buffer = pp;
+    *buffer_len = len;
+    return 0;
```
kmod: Let's move this function to capi/abstract.cpp since that's where we've been copying this kind of code (ex there's already a null_error function there).  Sorry that this got put here in the first place :/

kmod:
cool!  just a few comments

undingen:
I updated the pull request:
- fixed the cmake build by just removing the ```_io``` directory prefix it. The prefix is not needed because it recurses into every dir.
- added ```PYSTON_NOEXCEPT```
- I moved the ```PyNumber_*``` and ```PyBuffer_*``` funcs to ```abstract.cpp``` (cpython implements them in ```abstract.c```)
- removed the weak ref hack (thanks @toshok :-)). Sadly the ABC module does not work out of the box. It triggers ```RELEASE_ASSERT(0, "untested");``` in ```objmodel.cpp``` and the following code does not work. It calls into type new instead of calling ```ABCMeta.__new__```. I added a change as a separate commit (```37ee869```) which fixes the issue for me but I'm really unsure if that is the right fix because I didn't fully understand what it changes. Please have a look at it.

undingen:
Oh just noticed that the ```objmodel.cpp``` problem is already fixed... (at least we fixed it exactly the same way :+1:)
I rebased and removed my change and updated the pull request again.

kmod:
awesome :)

PR #349, opened 2015-03-05
basic implement of exec

tjhance:


tjhance:
I wasn't sure how to handle the implementations of `visit_exec` in `function_analysis` or in `type_analysis`... `exec` kinda mucks up any analysis you try to make, and I wasn't sure what the guarantees of those functions were supposed to be.

kmod:
This looks pretty good to me, but I think we do need a version that can modify the scopes that it's given.  Here's the namedtuple code, which is one of the cases we're hitting:
```python
    namespace = dict(_itemgetter=_itemgetter, __name__='namedtuple_%s' % typename,
                     OrderedDict=OrderedDict, _property=property, _tuple=tuple)
    try:
        exec class_definition in namespace
    except SyntaxError as e:
        raise SyntaxError(e.message + ':\n' + class_definition)
    result = namespace[typename]
```

I think I was pretty wrong about the urgency of this -- we can definitely do a temporary patch to get namedtuple working to give you more time to implement the dict-backed locals version.  Maybe we can chat on IM/gitter?

kmod:
Also, re function_analysis / type_analysis -- those analyses seem to be mutually exclusive with functions with exec in them.  My hope would be that the only types of names that query type/function analysis are LOAD_FAST and STORE_FAST, so in a function with an exec there wouldn't be any calls (I would guess that this is how it currently works but it might not).  Maybe we could do something like not computing those analyses, and throw an assert if we somehow do hit a case where we try to query them and they're NULL?

tjhance:
As a first step towards the proper LOAD_NAME/STORE_NAME scoping rules, I implemented a `SyntaxError` for when an `exec` or `import *` statement (which force the NAME scoping rules) are in a function with free variables.

The build is failing, but it's failing with

The command "if [ "$CXX" = "clang++" ]; then sudo apt-get install -qq clang-3.4 libstdc++-4.8-dev; fi" failed and exited with 100 during .

which seems like a problem with the ci system?

dagar:
@tjhance The build is failing because ppa.launchpad.net is timing out. The error is hidden in the collapsed apt-get update output.

W: Failed to fetch http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/precise/Release.gpg  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

undingen:
I just tried this out on the exec in the virtualenv site.py file, sadly it does not work yet:  
```SyntaxError: unqualified exec is not allowed in function 'addpackage' it is a nested function```

```python
def addpackage(sitedir, name, known_paths):
    """Add a new path to known_paths by combining sitedir and 'name' or execute
    sitedir if it starts with 'import'"""
    if known_paths is None:
        _init_pathinfo()
        reset = 1
    else:
        reset = 0
    fullname = os.path.join(sitedir, name)
    try:
        f = open(fullname, "rU")
    except IOError:
        return
    try:
        for line in f:
            if line.startswith("#"):
                continue
            if line.startswith("import"):
                exec(line)
                continue
            line = line.rstrip()
            dir, dircase = makepath(sitedir, line)
            if not dircase in known_paths and os.path.exists(dir):
                sys.path.append(dir)
                known_paths.add(dircase)
    finally:
        f.close()
    if reset:
        known_paths = None
    return known_paths
```
the strings to execs are:
```python
import sys; sys.__plen = len(sys.path)
```
```python
import sys; new=sys.path[sys.__plen:]; del sys.path[sys.__plen:]; p=getattr(sys,'__egginsert',0); sys.path[p:p]=new; sys.__egginsert = p+len(new)
```

Is it a lot of work to implement this use case in ```exec```?

tjhance:
If it's giving the syntax error and it shouldn't be, then that's just a bug with my syntax error logic.

So how much work it takes to support really depends on what `line` is. In this case it looks like `line` starts with `import`. If the goal is just to import the module, without adding the name to the scope, then it should just be a matter of fixing the syntax error bug.

In fact, you may even be able to get it to work by just dropping the second commit.

tjhance:
Just realized you actually gave examples of `line`. So I can say with greater confidence that I think they should be supported without needing the scoping stuff.

undingen:
cool, I removed the syntax error raising and it works :-)

tjhance:
OK, I think I fixed the issue in the question.

tjhance:
OK, I got `exec` scoping to work: you can now set local variables with `exec`.

I probably have some cleaning up to do, but the tests are passing, so I'll put this up here for feedback.

How do I make sure this doesn't have a performance regression? If I did everything right, there shouldn't be one, but it would be good to double-check. (The "unoptimized" scopes like ClassDefs and scopes with `exec` statements use the NAME lookup rules should be a lot slower, but there should be very few such scopes.)

src/analysis/scoping_analysis.cpp
```diff
@@ -462,16 +490,27 @@ class NameCollectorVisitor : public ASTVisitor {
...
         for (int i = 0; i < node->names.size(); i++) {
             AST_alias* alias = node->names[i];
-            mangleNameInPlace(alias->asname, cur->private_name, scoping->getInternedStrings());
-            mangleNameInPlace(alias->name, cur->private_name, scoping->getInternedStrings());
-            if (alias->asname.str().size())
-                doWrite(alias->asname);
-            else
-                doWrite(alias->name);
+            if (alias->name.str() == std::string("*")) {
+                doImportStar(node);
```
kmod: super random, but it looks like the name mangling should still happen on this path:
```python
class C(object):
    from __foo import *
```
throws `No module named _C__foo`

src/analysis/scoping_analysis.cpp
```diff
@@ -130,8 +133,18 @@ struct ScopingAnalysis::ScopeNameUsage {
...
     StrSet passthrough_accesses; // what names a child scope accesses a name from a parent scope
 
+    // `import *` and `exec` both force the scope to use the NAME lookup
+    bool hasNameForcingSyntax;
+    AST_ImportFrom* nameForcingNodeImportStar;
+    AST_Exec* nameForcingNodeBareExec;
+
+    // If it has a free variable / if any child has a free variable
+    bool free;
+    bool child_free;
```
kmod: Are these equivalent to `(bool)got_from_closure.size()` and `(bool)passthrough_accesses.size()`?

src/analysis/function_analysis.cpp
```diff
@@ -376,7 +378,8 @@ DefinednessAnalysis::DefinednessAnalysis(const ParamNames& arg_names, CFG* cfg,
     for (const auto& p : results) {
         RequiredSet required;
         for (const auto& p2 : p.second) {
-            if (scope_info->refersToGlobal(p2.first))
+            if (scope_info->refersToGlobal(p2.first)
+                || scope_info->getScopeTypeOfName(p2.first) == ScopeInfo::VarScopeType::NAME)
```
kmod: I'm still finding it confusing to mix the old functions (refersToGlobal) and new functions (getScopeTypeOfName)... is this the same as `scope_type == GLOBAL || scope_type == NAME`?  And should the logic here be `scope_type != FAST`?

src/analysis/scoping_analysis.cpp
```diff
@@ -130,8 +133,18 @@ struct ScopingAnalysis::ScopeNameUsage {
     StrSet got_from_closure;
     StrSet passthrough_accesses; // what names a child scope accesses a name from a parent scope
 
+    // `import *` and `exec` both force the scope to use the NAME lookup
+    bool hasNameForcingSyntax;
```
kmod: minory, but maybe we could have this as a function if it's just a simple derivation of other fields (`nameForcingNodeImportStar != NULL || nameForcingNodeBareExec != NULL`).

src/codegen/ast_interpreter.cpp
```diff
@@ -1010,33 +1054,21 @@ Value ASTInterpreter::visit_str(AST_Str* node) {
...
-                if (is_old_local)
-                    return getGlobal(source_info->parent_module, &node->id.str());
-
-                assertNameDefined(0, node->id.c_str(), UnboundLocalError, true);
-                return Value();
-            }
+    if (node->lookup_type == AST_Name::UNKNOWN) {
+        ScopeInfo::VarScopeType vst = scope_info->getScopeTypeOfName(node->id);
+        if (vst == ScopeInfo::VarScopeType::GLOBAL) {
+            node->lookup_type = AST_Name::GLOBAL;
```
kmod: If you feel like it, it'd be nice to get rid of the duplicate (and differently-named) AST_Name lookup types, and just use a VarScopeType there instead.

src/codegen/ast_interpreter.cpp
```diff
@@ -143,16 +144,39 @@ class ASTInterpreter {
...
-
 void ASTInterpreter::addSymbol(InternedString name, Box* value, bool allow_duplicates) {
     if (!allow_duplicates)
         assert(sym_table.count(name) == 0);
     sym_table[name] = value;
 }
 
+void ASTInterpreter::setGenerator(Box* gen) {
+    assert(gen->cls == generator_cls);
+    this->generator = static_cast<BoxedGenerator*>(gen);
```
kmod: I still think an `assert(!this->generator)` here would be nice :)

src/codegen/ast_interpreter.cpp
```diff
@@ -1115,38 +1152,41 @@ Box* astInterpretFunction(CompiledFunction* cf, int nargs, Box* closure, Box* ge
...
+    RELEASE_ASSERT(boxedLocals->cls == dict_cls, "we don't support non-dicts here yet");
+    interpreter.setBoxedLocals(boxedLocals);
     Value v = ASTInterpreter::execute(interpreter);
 
     return v.o ? v.o : None;
 }
 
 Box* astInterpretFrom(CompiledFunction* cf, AST_expr* after_expr, AST_stmt* enclosing_stmt, Box* expr_val,
-                      BoxedDict* locals) {
+                      BoxedDict* stackLocals) {
```
kmod: hmm why don't we need to pass boxedLocals here?

src/codegen/unwinding.cpp
```diff
@@ -542,16 +543,20 @@ BoxedModule* getCurrentModule() {
     return compiledFunction->clfunc->source->parent_module;
 }
 
-BoxedDict* getLocals(bool only_user_visible, bool includeClosure) {
-    for (PythonFrameIterator& frame_info : unwindPythonFrames()) {
+// TODO factor getStackLoclasIncludingUserHidden and fastLocalsToBoxedLocals
+// because they are pretty ugly but have a pretty repetitive pattern.
+
+// TODO no reason for this to be a BoxedDict over an ordinary dict, I think
```
kmod: Using a BoxedDict gives us GC support; we could probably also use a `std::unordered_map<STLCompatAllocator>` which I guess would let us not have to box the strings going in and out.

tjhance:
> Using a BoxedDict gives us GC support; we could probably also use a std::unordered_map<STLCompatAllocator> which I guess would let us not have to box the strings going in and out.

Actually, we should probably just use whatever ASTInterpreter uses for its symbol table, since that what it ends up becoming anyway.

src/codegen/unwinding.cpp
```diff
@@ -572,7 +577,83 @@ BoxedDict* getLocals(bool only_user_visible, bool includeClosure) {
...
+                        if ((v & 1) == 0)
+                            is_undefined.insert(p.first.substr(12));
+
+                        break;
+                    }
+                }
+            }
+
+            for (const auto& p : cf->location_map->names) {
+                if (p.first[0] == '!'
```
kmod: Ok, I pushed a change to make these names start with a '#' which I think should be more consistent -- could you try rebasing and seeing if you can just remove the checks for these names?

kmod:
Well, the interpreter uses a non-GC-aware data structure and has special GC support (`gatherInterpreterRoots`) that traverses it.

src/codegen/ast_interpreter.cpp
kmod: Maybe we should pass the entire FrameInfo object?  Feels like we need to propagate the exc_info as well.

kmod:
Ok, missed your comment about planning on doing more cleanup :P  I think things look pretty good.  You can check performance by doing

```shell
$ cd ~; git clone https://github.com/dropbox/pyston-perf
$ cd ~/pyston; git checkout BASELINE; make pyston_release
$ cd ~pyston-perf/benchmarking; python measure_perf.py --save=baseline
$ cd ~/pyston; git checkout TEST; make pyston_release
$ cd ~pyston-perf/benchmarking; python measure_perf.py --save=modified --compare=baseline
```
(doing this in two shells is easier)

Also, it looks like namedtuple does do `exec class_definition in namespace` :/

src/analysis/scoping_analysis.cpp
tjhance: `free` maybe. `child_free` is actually different, because this triggers, for example, because `h` has a free variable:

```python
def f():
    exec "a = 5"
    def g():
        b = 2
        def h():
            print b
```

I'm not sure if there's any reason for this to be invalid, but it is.

src/analysis/function_analysis.cpp
tjhance: Yeah I can clean this up.

As for whether it should `scope_type != FAST`, well, the `if` statement excluded closure variables before, and it still does. If that's correct, I don't know... I think you would know better than I would.

src/codegen/ast_interpreter.cpp
tjhance: That's probably a good idea, but right now there's not really a one-to-one correspondence

```
    enum LookupType {
        UNKNOWN,
        GLOBAL, // GLOBAL
        CLOSURE, // DEREF
        FAST_LOCAL, // FAST
        LOCAL, // NAME
    } lookup_type;
```

CLOSURE doesn't actually correspond to anything (I guess it's folded into FAST_LOCAL) so it's a little confusing

src/codegen/ast_interpreter.cpp
tjhance: Yeah seems like an issue.

src/codegen/unwinding.cpp
tjhance: Ah, I think that's a bit weird too, since since now they're the only '#' names that don't get in the ASTInterpreter's symbol table?

tjhance: Let's chat about this in-person tomorrow

tjhance:
Well I checked the performance and got this the first time:

```
              pyston (calibration)                      :    0.6s baseline: 0.6 (+1.2%)
              pyston interp2.py                         :    2.5s baseline: 2.6 (-2.0%)
              pyston raytrace.py                        :    4.9s baseline: 5.0 (-1.3%)
              pyston nbody.py                           :    6.7s baseline: 6.7 (+0.7%)
              pyston fannkuch.py                        :    5.1s baseline: 5.2 (-1.4%)
              pyston chaos.py                           :   11.2s baseline: 11.7 (-4.1%)
              pyston spectral_norm.py                   :   20.8s baseline: 21.6 (-3.7%)
              pyston pidigits.py                        :    3.8s baseline: 3.9 (-2.8%)
              pyston richards.py                        :    1.0s baseline: 1.0 (-2.2%)
              pyston deltablue.py                       :    1.7s baseline: 1.7 (-0.1%)
```

and the second time I got this

```
              pyston (calibration)                      :    0.6s baseline: 0.6 (-0.2%)
              pyston interp2.py                         :    2.6s baseline: 2.7 (-1.5%)
              pyston raytrace.py                        :    4.9s baseline: 5.3 (-9.3%)
              pyston nbody.py                           :    6.7s baseline: 6.9 (-3.1%)
              pyston fannkuch.py                        :    5.3s baseline: 5.3 (+0.8%)
              pyston chaos.py                           :   11.0s baseline: 11.6 (-4.9%)
              pyston spectral_norm.py                   :   20.5s baseline: 20.7 (-1.2%)
              pyston pidigits.py                        :    3.9s baseline: 3.6 (+7.6%)
              pyston richards.py                        :    1.0s baseline: 0.9 (+8.5%)
              pyston deltablue.py                       :    1.8s baseline: 1.6 (+7.5%)
```

It's noise as far as I can tell.

src/codegen/unwinding.cpp
kmod: Hmm the way I was thinking about it is that it's just an optimization to store them in the ASTInterpreter instead of in the symbol table -- not sure how much it gains us but it seems reasonable?  And I think we do ok about making sure to manually include those any time we want to iterate over the symbol table.

kmod:
Yeah, if you're running this on a machine with a desktop environment the results can have a lot of variance -- the `(calibration)` "benchmark" is a fixed amount of work, so the 1.2% change in that hints at the variance in the numbers.  I sent you an email about getting access to our development server.

For the LookupType vs VarScopeType, I think CLOSURE is missing since the LookupType is just a cache and it's only implemented for loads but not stores (ie `doStore()` calls `scope_info->refersToGlobal()` and then `scope_info->saveInClosure()` every time).

tjhance:
Okay, I implemented most of kmod's suggestions; in particular, I made sure to preserver the `frame_info`
during a deopt.

There are some more corner cases in `misc_scoping.py`, and it sounds like we need for `exec` to support passing in dictionaries as arguments. But I think this is enough for one pull-request.

tjhance:
Oh, the benchmarks are a bit disappointing, though :(


The benchmarks are a bit disappointing, though :(

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.2%)
              pyston interp2.py                         :    5.3s baseline: 5.1 (+4.0%)
              pyston raytrace.py                        :    7.5s baseline: 7.4 (+1.9%)
              pyston nbody.py                           :    8.9s baseline: 8.9 (+0.7%)
              pyston fannkuch.py                        :    6.9s baseline: 6.9 (+0.2%)
              pyston chaos.py                           :   21.3s baseline: 20.7 (+2.9%)
              pyston spectral_norm.py                   :   24.6s baseline: 24.9 (-1.1%)
              pyston pidigits.py                        :    4.8s baseline: 4.8 (+1.7%)
              pyston richards.py                        :    2.6s baseline: 2.6 (+2.4%)
              pyston deltablue.py                       :    2.3s baseline: 2.3 (+0.3%)
```

kmod:
lgtm, though one of the tests is timing out (due to the slow parsing of exec statements, I think) -- you could probably just reduce the iteration count on that one now that we can set the OSR limits manually.

tjhance:
I
- cleaned up the ScopeInfo API, completely removing the old one so in Python spirit there should be and only one obvious way to use it
- fixed the way the syntax error works (I tried to simplify it, but it's inherently complicated and the conditions don't really follow from the primitives that were already in scoping analysis)
- fixed it so that `locals()` will show variables that are passed-through the closure

kmod:
Hmm I think there's an issue that we're not passing the boxedlocals out when we OSR or reopt -- maybe we should just disable those for now?

Here are two tests that crash:
```python
def f1(x):
    exec """
for i in xrange(x):
    pass
print x
    """
f1(10000)
```
```python
def f3():
    exec """
def f2(x):
    def inner():
        return x
    return inner
    """
    g = f2(10000)
    for i in xrange(10000):
        g()
    print g()
f3()
```

sorry to drag this PR on, but after this I think it's good to go :)

tjhance:
ACK I totally forgot about OSRing! Will fix~

tjhance:
I now pass the FrameInfo through to the OSRs.

src/analysis/scoping_analysis.cpp
```diff
@@ -90,24 +91,25 @@ class ModuleScopeInfo : public ScopeInfo {
     ScopeInfo* getParent() override { return NULL; }
 
     bool createsClosure() override { return false; }
-
     bool takesClosure() override { return false; }
-
     bool passesThroughClosure() override { return false; }
 
-    bool refersToGlobal(InternedString name) override {
+    bool refersToGlobal(InternedString name) {
```
kmod: can we kill this?

src/analysis/scoping_analysis.cpp
```diff
@@ -130,8 +132,18 @@ struct ScopingAnalysis::ScopeNameUsage {
     StrSet got_from_closure;
     StrSet passthrough_accesses; // what names a child scope accesses a name from a parent scope
 
+    // `import *` and `exec` both force the scope to use the NAME lookup
+    AST_ImportFrom* nameForcingNodeImportStar;
+    AST_Exec* nameForcingNodeBareExec;
+    bool hasNameForcingSyntax() { return nameForcingNodeImportStar != NULL || nameForcingNodeBareExec != NULL; }
+
+    // If it has a free variable / if any child has a free variable
```
kmod: Can you add a comment with your explanation of why these are needed and how they are not derivable from the other attributes?

src/analysis/scoping_analysis.cpp
```diff
@@ -200,26 +210,26 @@ class ScopeInfoBase : public ScopeInfo {
...
         return usage->got_from_closure.count(name) != 0;
     }
-    bool saveInClosure(InternedString name) override {
+    bool saveInClosure(InternedString name) {
         // HAX
-        if (isCompilerCreatedName(name) || usesNameLookup)
+        if (isCompilerCreatedName(name) || usesNameLookup_)
             return false;
         return usage->referenced_from_nested.count(name) != 0;
     }
```
kmod: Can we kill these functions?  It looks like the only use is in getScopeTypeOfName, but we could just inline these in there and it would probably be more clear exactly what checks are happening (a lot of the paths that would result from inlining these functions into getScopeTypeOfName would be dead).

test/tests/misc_scoping.py
```diff
@@ -0,0 +1,199 @@
+# expected: fail
```
kmod: Could we move the tests that work out of here and leave a comment as to what's not working?

src/analysis/scoping_analysis.h
```diff
@@ -62,12 +63,24 @@ class ScopeInfo {
...
 
+    // Returns true if the variable should be passed via a closure to this scope.
+    // Note that:
+    //      (a) This can be false even if there is an entry in the closure object
+    //          passed to the scope, if the variable is not actually used in this
+    //          scope or any child scopes.
+    //      (b) This can be true even if it is not used in this scope, if it
+    //          is used in a child scope.
+
+    virtual bool isPassedToViaClosure(InternedString name) = 0;
```
kmod: Could this comment say something about why this function exists (that these names show up in locals even though they might never be referenced in the block)?  It's hard to look at this description and understand how exactly this is different from the other functions -- ex would this be the same as `getScopeTypeOfName() == DEREF || CLOSURE`?  Saying that it's used for a different purpose makes it more clear that we're not necessarily exposing different functionality here, just serving a slightly different purpose.  (Similarly it might be more clear to call it something like `closureNameShowsInLocals`.)

src/codegen/ast_interpreter.cpp
kmod: Sorry to harp on this, but I really think we can get rid of the duplicate system and that that would reduce the mental burden.  There's still a mapping from `VarScopeType` to the way that the AST interpreter retrieves the value for the name; the only thing is that we have the optimization that closure variables also got set in the local symbol table so we can receive them from there.  So we could just have the `FAST and `CLOSURE` cases fall through to the same code that accesses them in the symbol table.

src/codegen/ast_interpreter.cpp
```diff
@@ -1013,38 +1067,26 @@ Value ASTInterpreter::visit_str(AST_Str* node) {
...
-                assertNameDefined(0, node->id.c_str(), UnboundLocalError, true);
-                return Value();
-            }
+    if (node->lookup_type == AST_Name::UNKNOWN) {
+        ScopeInfo::VarScopeType vst = scope_info->getScopeTypeOfName(node->id);
+        if (vst == ScopeInfo::VarScopeType::GLOBAL) {
+            node->lookup_type = AST_Name::GLOBAL;
+        } else if (vst == ScopeInfo::VarScopeType::DEREF) {
+            node->lookup_type = AST_Name::DEREF;
+            return getattr(passed_closure, node->id.c_str());
```
kmod: ok this is really tiny, but this inconsistency (having only one branch immediately fetch the name) makes me think that something was different about this path, but then I looked further down and it looks like they're the same.

src/analysis/scoping_analysis.h
kmod: also, do we have any (active) tests for this feature?

tjhance: just something in `misc_scoping.py`

src/codegen/patchpoints.h
```diff
@@ -46,7 +46,7 @@ struct PatchpointInfo {
         CompilerType* type;
     };
 
-private:
+public:
```
kmod: revert?

tjhance: ACK thanks!

kmod:
sorry to harp on these things, but you're touching a bunch of code that's already confusing to people, so I really want us to do what we can to keep it understandable.  most of my comments correspond to places that I had to spend some time to understand what was going on.

can you also squash the changes down?  so we can hopefully forget about the saga that this functionality has been :P

libunwind
```diff
@@ -1 +1 @@
-Subproject commit 65ac86741606e1d87aef75755c699e4fa6884230
+Subproject commit 6e3254ea6e6fa6c433cd38f9716a0b3b4110de64
```
kmod: I think this is accidental?

kmod:
It's in!  Quite a bit more involved than we thought it would be :P  thanks for taking this on.

:star2:

PR #350, opened 2015-03-05
add support for relative imports, and bring our import api closer to cpython

toshok:
most of the changes are to support `from .foo import bar`, but there were other fixes along the way (absolute imports from within a package).  I also added the path parameter to imp.find_module while I was at it.

With this change `import json` gets past its relative imports.

### DIFF
--- a/src/runtime/import.cpp
+++ b/src/runtime/import.cpp
@@ -14,6 +14,8 @@
 
 #include "runtime/import.h"
 
+#include <limits.h>
+
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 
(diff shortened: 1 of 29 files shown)


src/runtime/import.cpp
```diff
@@ -140,19 +147,119 @@ SearchResult findModule(const std::string& name, const std::string& full_name, B
...
+   from the module foo.bar.bat (not itself a package), this returns the
+   sys.modules entry for foo.bar.  If globals is from a package's __init__.py,
+   the package's entry in sys.modules is returned, as a borrowed reference.
+
+   The *name* of the returned package is returned in buf.
+
+   If globals doesn't come from a package or a module in a package, or a
+   corresponding entry is not found in sys.modules, Py_None is returned.
+*/
+static Box* get_parent(Box* globals, int level, std::string& buf) {
```
kmod: What do you think about naming this `getParent()`?  It looks like this was mostly copied from CPython, but if it's a Pyston-style function now (throws Pyston exceptions), using the Pyston-style naming could help make that clear to callers.

I wish we had a better way of annotating this distinction :/

kmod:
lgtm, though I would advocate renaming the functions to camelCase to make them look more Pyston-y now that they throw Pyston exceptions.  I think that's also where the merge issue is coming from, since I renamed ensure_fromlist to ensureFromlist

toshok:
Yup, will rename

PR #351, opened 2015-03-05
copy _PyInt_Format implementation from cpython

toshok:
after #350 is merged, this is all that's required to get `import json` succeed.

### DIFF
--- a/src/runtime/int.cpp
+++ b/src/runtime/int.cpp
@@ -67,8 +67,87 @@ extern "C" PyObject* PyInt_FromLong(long n) noexcept {
     return boxInt(n);
 }
 
+/* Convert an integer to a decimal string.  On many platforms, this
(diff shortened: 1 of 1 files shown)


src/runtime/int.cpp
```diff
@@ -67,8 +67,87 @@ extern "C" PyObject* PyInt_FromLong(long n) noexcept {
     return boxInt(n);
 }
 
+/* Convert an integer to a decimal string.  On many platforms, this
+   will be significantly faster than the general arbitrary-base
+   conversion machinery in _PyInt_Format, thanks to optimization
+   opportunities offered by division by a compile-time constant. */
+static Box* int_to_decimal_string(BoxedInt* v) {
```
kmod: can we add noexcept here?  otherwise lgtm.

toshok:
added the noexcept.

PR #352, opened 2015-03-05
WIP: Virtualenv with hacks

undingen:
Sorry for this definitely not merge ready patch, but I'm looking for comments. While I try get more of the package management working....

With this changes one can do:
```
$./pyston_dbg virtualenv.py --no-setuptools  ~/pyston_env/
or
$./virtualenv.py -p ./pyston_dbg --no-setuptools  ~/pyston_env/
$source ~/pyston_env/bin/activate
$python
Pyston v0.3 (rev 423b82dc5656), targeting Python 2.7.6
>> import optparse
Warning: import level 0 will be treated as -1!
```

- The first commit adds some ```sys.prefix``` support. (I plan to upload that one soon as a separate pull request)
- Sadly we also have to load the modified the complex```site.py``` on startup because it sets up some paths. But this makes pyston startup time really slow... (commit 2, +enable pypa)
- The third commit adds my modified ```virtualenv.py``` with the modified ```site.py```

What I want to ask is: 
- what are your thoughts concerning the ```site.py``` should we just import it every time and pay with slower startup time because we have to support it. or is there another way? I'm not sure how important this file is... (I also commented out some parts to make it run)
- how much of virtualenv should we try implement for now? (pip support?)

I will clean up this mess soon and put up a virtualenv rep too make keeping keep track of my modifications possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,7 +292,7 @@ STDLIB_OBJS := stdlib.bc.o stdlib.stripped.bc.o
 STDLIB_RELEASE_OBJS := stdlib.release.bc.o
 ASM_SRCS := $(wildcard src/runtime/*.S)
 
-STDMODULE_SRCS := errnomodule.c shamodule.c sha256module.c sha512module.c _math.c mathmodule.c md5.c md5module.c _randommodule.c _sre.c operator.c binascii.c pwdmodule.c posixmodule.c _struct.c datetimemodule.c _functoolsmodule.c _collectionsmodule.c itertoolsmodule.c resource.c signalmodule.c selectmodule.c fcntlmodule.c timemodule.c arraymodule.c zlibmodule.c _codecsmodule.c socketmodule.c unicodedata.c _weakref.c cStringIO.c $(EXTRA_STDMODULE_SRCS)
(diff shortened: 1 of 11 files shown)


kmod:
I think we have to support site.py, and just deal with our startup time issues :/  Django is also slow to start up as well, so I don't think it's something we're going to be able to dodge.

I think we also want to support pip, since that+virtualenv will make testing out packages much easier.  I'm not sure what else is in virtualenv, but the workflow that'd be great to support is something like:
```
$ virtualenv -p pyston pyston-env
$ cd pyston-env; . bin/activate
$ pip install foo_library
$ python # invokes pyston
>> import foo_library
>> foo_library.foo()
```

kmod:
it sounds like we're getting a lot closer though!

undingen:
ok, that's what I thought and work on.

In order to make our existing stat checking tests still work, should I add the stats generated during the ```site.py``` import to the _init_ or should I create a new prefix?

kmod:
Oh shoot, yeah it will affect all the tests too.  Maybe we could add '-S' (aka don't import site.py) to the list of standard args we use when running the tests?  otherwise adding to the init counts sounds reasonable.

undingen:
ok this wip pull request served it's purpose, I'll close it

PR #353, opened 2015-03-06
add a json test and get it running

toshok:
test is basically the contents of this page: https://docs.python.org/2/library/json.html


kmod:
nice!

PR #354, opened 2015-03-06
Add sys.prefix, sys.exec_prefix, site.py and $PYTHONPATH support

undingen:
I'm not 100% happy with this change but I think it's for now quite reasonable behavior and it helps getting virtualenv running.
It not exactly the same behavior as cpython but we use a different directory structure so it can't be.
(PyPy is also different then cpython in this regard).

The ```site.py``` script will now get imported on startup except ```-S``` is specified. I added the flag to all our regressions test runs --> ```site.py``` loading won't get tested by our test-suite (because otherwise several tests which check the stats will fail and also ```make check``` would be much slower). We probably need to adopt our site.py script in the future.

I removed printing out ```sys.prefix``` because it won't be the same as cpython. (It depens on the install location,...)


undingen:
Rebased to fix the merge conflict and added the CMakeList.txt change I forgot.

Issue #258, opened 2015-01-04
Improve Travis CI testing speed


kmod:
@dagar First of all, thanks for getting it working, it's definitely a huge help :)

I wonder if there's something we can do about the testing time though, which I assume comes mostly from having to compile LLVM.  I don't think we can easily get rid of our source dependency on LLVM, but we end up building the same stuff every time so maybe there's some way to cache it?  We could probably come up with some hash of {llvm_rev, patchset}, and map that to the final build products that we use.  A simpler option could be to enable ccache for the Travis build and have Travis cache the .ccache directory.

What do you think?  Not hugely important at this stage but could be nice.

dagar:
I looked into this initially. I was worried about exceeding the travis-ci.org fair use. The current travis free tier doesn't support caching between builds. Their new docker based infrastructure does, but doesn't let you install packages or use your own containers yet.

The travis build could try to retrieve a prebuilt llvm+clang from something like s3 using the llvm rev + patchset hash. At the end of the build after testing it could upload to s3 if it built from source. CMake needs a few small changes to work with either a prebuilt llvm or from source.

Another option you may want to consider is providing a machine. I believe it's fairly easy to replicate the github + travis integration with jenkins. Builds would be significantly faster with ccache plus not having to install packages and full clone every time. The same jenkins build could also update speed.pyston.org.

dagar:
Another plus for the bring your own machine idea is that we could easily extend the testing matrix to include Release, Release+Asserts, Debug with both clang and gcc or different modes (GRWL, GIL, etc). We could even add builds that run under valgrind or with the sanitizers (asan, tsan, etc).

marscher:
at least receiving the sources of llvm via a tarball or zip saves ~1.5 min despite cloning their whole (really huge) history ;)

dagar:
@kmod - if you're interested in the bring your own hardware I could take care of the setup and integration. You'd basically just need to provide a machine with ubuntu + jenkins + pyston packages and make the jenkins web interface publically accessible (probably github auth).

[coreclr](https://github.com/dotnet/coreclr/blob/master/README.md) is using jenkins with github integration like travis.

kmod:
I'm not very psyched about sysadmining another machine :)  The 20 minute Travis-CI testing time is kind of annoying right now but isn't too high up on the priority list; I'm sure we'll set something like this up eventually but it might be a while before it makes sense to learn about jenkins and what not.

kmod:
[...]

PR #355, opened 2015-03-06
travis try enabling apt cache

dagar:


dagar:
Overall saves about 80 seconds per build. Doesn't exactly resolve PR #258, but it's something...

undingen:
Maybe we could enable ccache and enable the travis ci cache for it? (to speed up the llvm build)

dagar:
All caching is supposed to be only available for the paid tiers. Based on their documentation I didn't expect this to even work.

undingen:
Oh yeah, I should have read the the whole page instead of only a sentence... :-1: (or at least the first one :bowtie:)

dagar:
Let's try anyway. #357

kmod:
cool, thanks :)

PR #356, opened 2015-03-06
Add the zipimport, csv and ssl module

undingen:
Add the zipimport module and add sys.path_hook support:
With this changes we are able to import packages contained in a zip file when the filename is in ```sys.path```. (With #354 we should in addition be able to import stuff from archives set via $PYSTONPATH)
- I did not port the ```marshal``` module, I just ifdefed out everything which is not needed by the ```zipimport``` module. 
- I also removed support for bytecode archives inside ```zipimport``` and instead of passing around code objects I pass around the source code as string. I think for a first version this should be fine.


undingen:
Rebased the changes to resolve the merge conflict.

undingen:
I have added the ```csv``` and ```ssl``` module import to this pull request because separate pull requests would cause merge conflicts. Hope this doesn't make it more difficult to review.
- I changed two classes in the stdlib to inherit from object because it looks like we don't yet support iterable old style classes.
- ```getCurrentModule()```  will return NULL when we can't find a known python stack frame, this happens when a C API module imports another module during init. Not sure if this is the right behavior.

from_cpython/Modules/zipimport.c
```diff
@@ -4,6 +4,8 @@
 #include "marshal.h"
 #include <time.h>
 
+// Pyston change:
+#include <sys/stat.h>
```
kmod: minor (I merged the PR without it), but this should be part of pyport.h

kmod: (same for getpath.c)

PR #357, opened 2015-03-06
travis try enabling ccache

dagar:


dagar:
Definitely does not work.

PR #359, opened 2015-03-07
virtualenv fixes

undingen:
Add imp.load_module() and next() support
The ```imp.load_module()``` just supports a specific type virtualenv needs for now.


undingen:
also added: ```int << long``` support for ```uuid```

undingen:
Ok I sneaked in even more commits. I think the changes are not controversial and the number of pull requests is getting out of control :-D so added them to that one.

kmod:
Awesome, thanks for all the fixes :)  Let's talk at the next meeting about a more scalable code review model...

PR #360, opened 2015-03-07
fixed deopt bug involving closures

tjhance:
I noticed that we weren't setting `passed_closure` or `created_closure` in the `ASTInterpreter` after a deopt. I fixed the issue and added a regression test in `deopt_tests.py`.

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -143,6 +143,10 @@ public:
     const ScopeInfo* getScopeInfo() { return scope_info; }
 
     void addSymbol(InternedString name, Box* value, bool allow_duplicates);
+    void setGenerator(Box* gen);
+    void setPassedClosure(Box* closure);
+    void setCreatedClosure(Box* closure);
+
     void gcVisit(GCVisitor* visitor);
 };
 
@@ -151,6 +155,21 @@ void ASTInterpreter::addSymbol(InternedString name, Box* value, bool allow_dupli
     if (!allow_duplicates)
         assert(sym_table.count(name) == 0);
     sym_table[name] = value;
+}
+
+void ASTInterpreter::setGenerator(Box* gen) {
+    assert(gen->cls == generator_cls);
+    this->generator = static_cast<BoxedGenerator*>(gen);
+}
+
+void ASTInterpreter::setPassedClosure(Box* closure) {
+    assert(closure->cls == closure_cls);
+    this->passed_closure = static_cast<BoxedClosure*>(closure);
+}
+
+void ASTInterpreter::setCreatedClosure(Box* closure) {
+    assert(closure->cls == closure_cls);
+    this->created_closure = static_cast<BoxedClosure*>(closure);
 }
 
 void ASTInterpreter::gcVisit(GCVisitor* visitor) {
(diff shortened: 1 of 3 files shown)


src/codegen/ast_interpreter.cpp
```diff
@@ -1144,9 +1163,17 @@ Box* astInterpretFrom(CompiledFunction* cf, AST_expr* after_expr, AST_stmt* encl
 
     for (const auto& p : locals->d) {
         assert(p.first->cls == str_cls);
-        auto name = static_cast<BoxedString*>(p.first)->s;
-        InternedString interned = cf->clfunc->source->getInternedStrings().get(name);
-        interpreter.addSymbol(interned, p.second, false);
+        std::string name = static_cast<BoxedString*>(p.first)->s;
```
kmod: minor nit -- this will make a copy (I guess the original version did too), could you have this be a `const std::string&`?

src/codegen/ast_interpreter.cpp
```diff
@@ -153,6 +157,21 @@ void ASTInterpreter::addSymbol(InternedString name, Box* value, bool allow_dupli
     sym_table[name] = value;
 }
 
+void ASTInterpreter::setGenerator(Box* gen) {
+    assert(gen->cls == generator_cls);
+    this->generator = static_cast<BoxedGenerator*>(gen);
```
kmod: Could you have these assert that these values weren't already set, just to make it clear that these are just for initialization?

src/codegen/unwinding.cpp
```diff
@@ -582,7 +582,9 @@ BoxedDict* getLocals(bool only_user_visible, bool includeClosure) {
             }
 
             for (const auto& p : cf->location_map->names) {
-                if (p.first[0] == '!')
+                if (p.first[0] == '!' && (only_user_visible
+                                          || !(p.first == PASSED_GENERATOR_NAME || p.first == PASSED_CLOSURE_NAME
+                                               || p.first == CREATED_CLOSURE_NAME)))
```
kmod: I utterly failed to parse this conditional :(  I think this could be made easier to understand, but I think this is also evidence that these names should start with `#` and not `!` -- `!` are supposed to be irgen-internal names.  Would you mind trying switching the names and seeing if anything breaks?  If something does it's probably not worth hunting down unless you're curious.

src/codegen/ast_interpreter.cpp
```diff
@@ -1144,9 +1163,17 @@ Box* astInterpretFrom(CompiledFunction* cf, AST_expr* after_expr, AST_stmt* encl
...
-        auto name = static_cast<BoxedString*>(p.first)->s;
-        InternedString interned = cf->clfunc->source->getInternedStrings().get(name);
-        interpreter.addSymbol(interned, p.second, false);
+        std::string name = static_cast<BoxedString*>(p.first)->s;
+        if (name == PASSED_GENERATOR_NAME) {
+            interpreter.setGenerator(p.second);
+        } else if (name == PASSED_CLOSURE_NAME) {
+            interpreter.setPassedClosure(p.second);
+        } else if (name == CREATED_CLOSURE_NAME) {
+            interpreter.setCreatedClosure(p.second);
```
kmod: Hmm it worries me that we have two places that we can store these names, either in the symbol table or the special slots for them.  This is made worse by the fact that irgen just stores the names as normal names in the symbol table and not in special slots -- if we want to keep slots for perf reasons, maybe we could add an assert somewhere that we're not trying to store/load these special names from the symbol table?

tjhance:
So, I actually think this function `getLocals` is getting pretty ugly as it serves two pretty distinct purposes

- get the stack locals for deopting purposes
- get the `locals()` dict

In my work on `exec` (which will be ready soon, I think), the two purposes diverged every further and I ended up splitting this function up into two.

Once it's split up, we can clean up the interface quite a bit, and instead of a `BoxedDict` (there's not really any reason to use a Python object)

```C++
struct {
    your_favorite_map_data_structure<std::string, Box*>
    BoxedClosure* createdClosure;
    BoxedClosure* passedClosure;
    BoxedGenerator* generator;
}
```

or something like that. But it would be less work to just do this refactor after I get in my `exec` stuff which as I said already touches this a lot.

tjhance:
This ended up in #349 so I'm closing it

PR #361, opened 2015-03-08
Add bytearray and add a memoryview test

undingen:


kmod:
I am very surprised these work!

PR #363, opened 2015-03-09
make tester produce useful help messages

rntz:
I was frustrated by having to read tester.py's source whenever I wanted to know how to use it, so now it's self-describing.

Unfortunately, the parsing of `-a` changes slightly; instead of `-a -n` (or some other option), you have to invoke `-a=-n`, to avoid `argparse` interpreting `-n` as another option. Seems a small price to pay, but I can probably find a way to do this better if you want to keep the old parsing.


kmod:
cool, thanks for the cleanup

PR #364, opened 2015-03-09
don't let AttributeError exceptions bubble out of softspace.  fixes fasta.py

toshok:
PyFile_Softspace (according to cpython docs) returns 0 and clears all errors if there was a problem getting or setting "softspace".  It's pretty clear that the rest of the codebase assumes as much from softspace as well.  We might be able to also wrap isSubclass in the try{} block and add noexcept, but I'm not sure if the isSubclass check would cause cpython to fail.

without this change we abort because of the catch block in PyFile_Softspace.  fixing that to return 0 bubbles Attribute errors from a few places (ast_interpreter.cpp, jitted code, etc.)

### DIFF
--- a/src/runtime/file.cpp
+++ b/src/runtime/file.cpp
@@ -974,7 +974,7 @@ extern "C" int PyFile_SoftSpace(PyObject* f, int newflag) noexcept {
     try {
         return softspace(f, newflag);
     } catch (ExcInfo e) {
-        abort();
+        return 0;
     }
 }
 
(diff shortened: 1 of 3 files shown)


undingen:
thanks for fixing the bug I introduced, haven't run fast.py lately

toshok:
no worries at all.  I hadn't run the benchmarks either lately.  Kinda
surprised that was the only failure.

kmod:
cool, good catch :)

PR #365, opened 2015-03-10
__setattr_ and thread._local

toshok:
Don't feel like this is ready to merge yet, mostly looking for comments. I'm dissatisfied with the reentrancy involving setattrInternal+PyObject_GenericSetAttr.

### DIFF
--- a/src/capi/object.cpp
+++ b/src/capi/object.cpp
@@ -158,8 +158,6 @@ extern "C" int PyObject_SetAttr(PyObject* v, PyObject* name, PyObject* value) no
 }
 
 extern "C" int PyObject_SetAttrString(PyObject* v, const char* name, PyObject* w) noexcept {
-    if (Py_TYPE(v)->tp_setattr != NULL)
-        return (*Py_TYPE(v)->tp_setattr)(v, const_cast<char*>(name), w);
     try {
         setattr(v, name, w);
     } catch (ExcInfo e) {
--- a/src/capi/object.cpp
+++ b/src/capi/object.cpp
@@ -144,7 +144,13 @@ extern "C" PyObject* PyObject_SelfIter(PyObject* obj) noexcept {
 }
 
 extern "C" int PyObject_GenericSetAttr(PyObject* obj, PyObject* name, PyObject* value) noexcept {
-    Py_FatalError("unimplemented");
+    try {
+        setattr(obj, static_cast<BoxedString*>(name)->s.c_str(), value);
+    } catch (ExcInfo e) {
+        setCAPIException(e);
+        return -1;
+    }
+    return 0;
 }
 
 extern "C" int PyObject_SetAttr(PyObject* v, PyObject* name, PyObject* value) noexcept {
@@ -152,6 +158,8 @@ extern "C" int PyObject_SetAttr(PyObject* v, PyObject* name, PyObject* value) no
 }
 
 extern "C" int PyObject_SetAttrString(PyObject* v, const char* name, PyObject* w) noexcept {
+    if (Py_TYPE(v)->tp_setattr != NULL)
+        return (*Py_TYPE(v)->tp_setattr)(v, const_cast<char*>(name), w);
     try {
         setattr(v, name, w);
     } catch (ExcInfo e) {
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -678,11 +678,6 @@ void Box::setattr(const std::string& attr, Box* val, SetattrRewriteArgs* rewrite
         return;
     }
 
-    if (cls->tp_setattro) {
-        cls->tp_setattro(this, boxString(attr), val);
-        return;
-    }
-
     // Unreachable
     abort();
 }
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -81,6 +81,7 @@ static const std::string iter_str("__iter__");
 static const std::string new_str("__new__");
 static const std::string none_str("None");
 static const std::string repr_str("__repr__");
+static const std::string setattr_str("__setattr__");
 static const std::string setitem_str("__setitem__");
 static const std::string set_str("__set__");
 static const std::string str_str("__str__");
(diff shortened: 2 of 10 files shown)


src/runtime/objmodel.cpp
```diff
@@ -677,6 +678,11 @@ void Box::setattr(const std::string& attr, Box* val, SetattrRewriteArgs* rewrite
         return;
     }
 
+    if (cls->tp_setattro) {
+        cls->tp_setattro(this, boxString(attr), val);
+        return;
+    }
+
```
kmod: I don't think we should be checking this here?

src/capi/object.cpp
```diff
@@ -152,6 +158,8 @@ extern "C" int PyObject_SetAttr(PyObject* v, PyObject* name, PyObject* value) no
 }
 
 extern "C" int PyObject_SetAttrString(PyObject* v, const char* name, PyObject* w) noexcept {
+    if (Py_TYPE(v)->tp_setattr != NULL)
+        return (*Py_TYPE(v)->tp_setattr)(v, const_cast<char*>(name), w);
```
kmod: Can't this rely on `setattr` to check for tp_setattr?

src/runtime/objmodel.cpp
```diff
@@ -1705,8 +1711,37 @@ void setattrInternal(Box* obj, const std::string& attr, Box* val, SetattrRewrite
         // We don't need to to the invalidation stuff in this case.
         return;
     } else {
+        // Finally, check __setattr__
+        if (obj->cls->tp_setattr) {
+            rewrite_args = NULL;
+            REWRITE_ABORTED("");
+
+            int rtn = obj->cls->tp_setattr(obj, const_cast<char*>(attr.c_str()), val);
```
kmod: We've been changing the C API headers to use `const char*`s rather than casting away the constness -- maybe we could do that for tp_setattr?  I guess that would require changing a number of extension modules though.

src/runtime/objmodel.cpp
```diff
@@ -1705,8 +1711,37 @@ void setattrInternal(Box* obj, const std::string& attr, Box* val, SetattrRewrite
...
+            return;
+        }
+        Box* setattr = typeLookup(obj->cls, setattr_str, NULL);
+        if (setattr) {
+            rewrite_args = NULL;
+            REWRITE_ABORTED("");
+
+            // if we're dealing with a BoxedWrapperDescriptor wrapping
+            // PyObject_GenericSetAttr, skip calling __setattr__, as
+            // that will just re-enter us.
```
kmod: Hmm yeah we've been running into this with a number of the tp slots....  When do we generate a class where both versions delegate to the other?   Maybe we can approach this by saying that that those cases are wrong and we won't produce things of that form?

kmod:
Looks pretty reasonable :)  Could we add some tests for the new slots though? (in slots_tester.c)

src/runtime/objmodel.cpp
toshok: quite right.  Didn't I do this exact same thing with the `__getattr__` changes?  removed.

src/capi/object.cpp
toshok: yup

src/runtime/objmodel.cpp
toshok: yeah, for apis that we implement it's an easy switch.

maybe we could #ifdef the typedef for setattrfunc to be const for c++, non-const for C?

src/runtime/objmodel.cpp
toshok: The problem arises when an extension exposes a type that has its `tp_setattro` set to `PyObject_GenericSetAttr`.  This is totally valid unfortunately, and `_functools.c` does it for the `partial` type.  Without the check, `socket_test.py` fails because of this.

src/runtime/objmodel.cpp
kmod: Hmm I'd prefer to leave the const_cast explicit.  Taking a look at the standard library, any module that sets tp_setattr explicitly casts it to setattrfunc, so in theory we could try changing the setattrfunc typedef without the ifdef, but for a reason that defeats the type checking anyway.

I guess let's just keep this for now even though it's worrisome...

src/runtime/objmodel.cpp
kmod: Ah interesting... I think the issue is that PyObject_GenericSetAttr is not supposed to check for `__setattr__` (or `tp_setattr[o]`) at all.  This check covers the common case that PyObject_GenericSetAttr is called on an object where it is also the value of tp_setattro, but it's possible to call PyObject_GenericSetAttr on an object with a different tp_setattro, and then we'd have the same issue.  This could happen if an extension class inherits from an object and wants to add their own setattr that handles some but not all attribute names, and does something like:

```C++
int mysetattr(PyObject* obj, char* name, PyObject* value) {
  if (strcmp(name, "myattrname") == 0)
    return 0;
  return basecls->tp_setattr(obj, name, value);
}
```

I don't think this happens anywhere, but it did happen for for some other fields -- I ran into this with 79b2e9ccd835fb2cfeb47756a92295d4d17ef91b.  I hacked around it there, so I think the check you have here is good enough for now, but we'll have to fix these issues for real at some point :/

src/runtime/objmodel.cpp
```diff
@@ -1705,8 +1706,37 @@ void setattrInternal(Box* obj, const std::string& attr, Box* val, SetattrRewrite
...
+            if (setattr->cls != wrapperdescr_cls
+                || ((BoxedWrapperDescriptor*)setattr)->wrapped != PyObject_GenericSetAttr) {
+                Box* boxstr = boxString(attr);
+                Box* rtn = callattrInternal2(obj, &setattr_str, CLASS_ONLY, NULL, ArgPassSpec(2), boxstr, val);
+                if (rtn == NULL)
+                    runtimeCall3(setattr, ArgPassSpec(3), obj, boxstr, val);
```
kmod: Sorry, just saw this -- I'm confused by these couple lines.  Shouldn't rtn always be non-NULL here, since we already verified that the attribute exists?  But maybe we're better off just skipping to calling runtimeCall3 on the setattr object?

toshok: you're right about the fallback not being necessary.  that check predates the addition of the typeLookup above above.  Going to switch from runtimeCall3 to runtimeCallInternal to match the rest of the code in this method.

kmod:
Hey, sorry, is this ready to merge?

toshok:
No worries.  It is indeed.

PR #366, opened 2015-03-10
Add a few 'long' methods, support tuple sublasses, add uuid test and fix a _struct GC bug

undingen:
This are fixes for virtualenv...


kmod:
cool :)

PR #369, opened 2015-03-11
Implement set comprehensions, fix dict comprehension scope

rntz:


src/analysis/scoping_analysis.cpp
```diff
@@ -401,11 +397,21 @@ class NameCollectorVisitor : public ASTVisitor {
...
+    void visit_comp_values(AST_GeneratorExp* node) { node->elt->accept(this); }
+    void visit_comp_values(AST_SetComp* node) { node->elt->accept(this); }
+    void visit_comp_values(AST_DictComp* node) {
+        node->key->accept(this);
+        node->value->accept(this);
+    }
+
+    template <typename CompType> bool visit_comp(CompType* node) {
+        // NB. GeneratorExps evaluate their first for-subject's expression outside of the function scope they create.
+        // dictcomps & setcomps don't.
```
kmod: Hmm...

```python
x = "before"
{i for i in [x for x in xrange(5)]}
print x
```
CPython prints 4, this PR prints "before".  It's a lot more esoteric than the generatorexp version though (with generator expressions it affects not just scoping but the order in which things get evaluated).

test/tests/setcomp.py
```diff
@@ -0,0 +1,75 @@
+# this file is adapted from dictcomp.py
+# TODO: this shouldn't be necessary, but our set implementation sucks & doesn't keep things in order
```
kmod: Not true -- maybe it sucks, but the lack of ordering isn't why :P  Python sets and dicts are well-specified to be hash tables and to have undefined iteration order:
https://docs.python.org/2/library/stdtypes.html#types-set

I think it's still somewhat common for code to rely on set/dict ordering, but I think there's a decent consensus that that's a bug on the user's part.

src/analysis/scoping_analysis.cpp
kmod: (we should add a test for this too)

kmod:
looks good, just a few minor comments.

src/analysis/scoping_analysis.cpp
rntz: ooh, good catch.

rntz:
Ok @kmod, this should be mergeable now, I hope.

kmod:
cool!

PR #370, opened 2015-03-11
Document that we need libssl-dev

rntz:


kmod:
Do we need to add this to the cmake instructions too?  I guess it's already included in the travis build but I'm not sure how.

rntz:
I actually don't know anything about building with cmake. When I try following the instructions it breaks on my machine.

rntz:
@kmod Ok, this should be mergeable now.

corona10:
I experienced build-broken with 'can not find -lcrypto -lssl' message. 
I resolved it by installing libssl-dev

kmod:
Cool, thanks :)

PR #371, opened 2015-03-12
quick fix for support long type in range() function

corona10:
quick fix for support long type in range() function

but it needs some re-factoring..

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -61,7 +61,6 @@ extern "C" Box* trap() {
 
     return None;
 }
-
 /* Helper for PyObject_Dir.
    Merge the __dict__ of aclass into dict, and recursively also all
    the __dict__s of aclass's base classes.  The order of merging isn't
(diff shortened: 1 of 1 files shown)


kmod:
Hmm, maybe you could try using `PyLong_AsLong()`?  This function handles both longs and ints and converts them to an unboxed long.  Since this is a CAPI function being called from a Pyston function, you'd have to call `checkAndThrowCAPIException()` after each call to it.

corona10:
hmm i used  asSignedLong(BoxedLong* para);

for example

 if(isSubclass(start->cls, long_cls))
 {
      istop = asSignedLong((BoxedLong*)start);
 }

corona10:
okay i will try it

corona10:
but is there any way to get pyobject from box to use PyLong_AsLong() function?
I know that i am less famillar with pyston source code.
but i can not find the way

kmod:
PyObject* is a typedef for Box*, so you can just pass it through :)

corona10:
I re-factored the code after your reviewing it looks more clearly and cool

kmod:
Looks good :)  Can you squash the changes and undo the extraneous whitespace change?  Also, let's not call RELEASE_ASSERT on the types any more -- PyLong_AsLong does the proper error checking and will raise any issues as a Python error instead of crashing the process :P

corona10:
okay i fix it

kmod:
Merged in 0e111b5f967b8c8a0a50f052eb16407800d08171

kmod:
thanks for the contribution!

PR #374, opened 2015-03-13
expose module __dict__ as an attr wrapper

toshok:
warnings.py access sys.__dict__ in warn().

before:

```
>> import warnings
>> warnings.warn("hi")
  File "<stdin>", line 1, in <module>:
  File "/home/vagrant/pyston-build/from_cpython/Lib/warnings.py", line 194, in warn:
    globals = sys.__dict__
AttributeError: 'module' object has no attribute '__dict__'
```

after:

```
>> import warnings
>> warnings.warn("hi")
/home/vagrant/pyston-build/__builtin__:1: UserWarning: hi
None
```

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -2025,6 +2025,15 @@ BoxedModule* createModule(const std::string& name, const std::string& fn) {
     d->d[b_name] = module;
 
     module->giveAttr("__doc__", None);
+
+    // not strictly correct, as the value of __dict__ is not a
+    // BoxedDict, and also doing it this way makes the following work
+    // in pyston but not cpython:
+    //
+    // import sys
+    // sys.__dict__['__dict__']
+    module->giveAttr("__dict__", makeAttrWrapper(module));
+
     return module;
 }
 


kmod:
erf sorry this is fallout from [my change](https://github.com/dropbox/pyston/commit/bc62488ac3bc44d4b900719a1e435bd1a2015ca2) to the way `__dict__` works that I guess I didn't see through all the way.  That change made it so we can add `__dict__` the way CPython does (minus the fact that it's an attrwrapper instead of a dict), which is by adding a descriptor on the module object:

```python
>>> type(sys).__dict__['__dict__']
<member '__dict__' of 'module' objects>
>>> type(sys).__dict__['__dict__'].__get__(sys)['exit']
<built-in function exit>
```

I hopefully fixed this in https://github.com/dropbox/pyston/commit/f80d8bb6d5e19cf470de905e590a9091c28a5feb, though that particular snippet doesn't quite work.

toshok:
no worries at all.  f80d8bb fixes the warnings.py problem as well.  cool.

PR #375, opened 2015-03-13
INSTALLING.md: add instructions on getting gdb pretty-printers to work

rntz:


kmod:
cool :)  is there a way to tell which gdb versions correspond with which gcc versions?  we could try bumping one of their versions up or down.

Or migrate to gcc 4.9 if anyone volunteers :)

PR #378, opened 2015-03-14
Improvements: Deploy deb/rpm

denji:
The prototype skeleton to build the deb/rpm/etc:

cmake:
- `DEPLOY[=[ON|1]]` (Enable building package)
- `*_GENERATOR=[DEB;RPM;PackageMaker;DragNDrop;TBZ2;TGZ;ZIP;ZIP]` (optional, default: auto-detect)
- `*_SOURCE_GENERATOR=[TBZ2;TGZ;ZIP;ZIP]` (optional, default: auto-detect)
- `NIGHTLY[=[ON|1]]` (nightly-version filename git-suffix)
  
  ```
  # [make|ninja] package
  # cpack ..
  ```
  
  ```
  Generators
  7Z                          = 7-Zip file format
  Bundle                      = Mac OSX bundle
  DragNDrop                   = Mac OSX Drag And Drop
  IFW                         = Qt Installer Framework
  NSIS                        = Null Soft Installer
  NSIS64                      = Null Soft Installer (64-bit)
  OSXX11                      = Mac OSX X11 bundle
  PackageMaker                = Mac OSX Package Maker installer
  STGZ                        = Self extracting Tar GZip compression
  TBZ2                        = Tar BZip2 compression
  TGZ                         = Tar GZip compression
  TXZ                         = Tar XZ compression
  TZ                          = Tar Compress compression
  ZIP                         = ZIP file format
  ```


cmake/Deploy.cmake
```diff
@@ -0,0 +1,70 @@
+if(DEPLOY)
+  set(CPACK_PACKAGE_NAME "pyston")
+  set(CPACK_PACKAGE_VERSION "0.3")
+  set(CPACK_PACKAGE_VENDOR "Dropbox")
+  set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Python implementation built using LLVM and JIT techniques")
+
+  set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "Pyston targets Python 2.7, only runs on x86_64 platforms.")
+
+  set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_LIST_DIR}/LICENSE")
```
dagar: Should be top level? ${CMAKE_SOURCE_DIR}

cmake/Deploy.cmake
```diff
@@ -0,0 +1,70 @@
...
+  set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_LIST_DIR}/README.md")
```
dagar: ${CMAKE_SOURCE_DIR}/README.md

dagar:
We'd need to setup cmake install for this to work.
http://www.cmake.org/cmake/help/v3.0/command/install.html

dagar:
On my Ubuntu system the default filename ends with a dash (Pyston-0.3-.deb) and it also tries to create an rpm, but fails.

Once this is working travis-ci can upload the packages to github releases whenever pyston is tagged (eg releases). http://docs.travis-ci.com/user/deployment/releases/

kmod:
Cool, I got it to make a .deb (I had to pass `-DCPACK_GENERATOR=DEB`), but from the looks of it this is only packaging libpypa?

Also, I realized this after the latest github release, but we need to package lib_pyston/ and from_cpython/Lib/ with our releases.

dagar:
Cpack needs cmake INSTALL rules for Pyston to build the packages. We'd need to decide where from_cpython (stdlib), lib_pyston, and parse_ast.py should be installed and how Pyston finds them (probably setting PYTHONPATH).

kmod:
@undingen has been looking into these kinds of things with his work on virtualenv.  I'm not sure how much demand there will be for a .deb or .rpm though at this point -- I think the two main use cases for now are Dropbox's internal use and for distributing releases for people to play with, the first which uses some internal tools, and the second might be better served by distributing tarballs?  I guess we'll deal with similar issues building those though.

PR #379, opened 2015-03-14
add cpython buffer + sqlite3 module, and a couple tests.

toshok:
also add some capi functions (unimplemented) to get sqlite3 module to link.

the tests are in no way exhaustive, but sqlite is importable and works for simple tables.

also, the buffer type is missing it's tp_compare (we assert in PyType_Ready if tp_compare is not null), so we'll need to deal with that.


kmod:
that's awesome!

I'm going to add an issue for running the tests in a temporary directory so that we don't have to worry so much about whether `example.db` already exists or if we have to clean it up afterwards.

Issue #337, opened 2015-02-27
use llvm::SmallVector in more places


toshok:
we should use `llvm::SmallVector` instead of `std::vector` in more places (potentially everywhere.)

It provides for stack space to be used as backing store for the data, instead of the `std::vector` method of always allocating.

Reducing these calls to malloc (especially in hot paths) should have a pretty significant affect on performance.

Issue #253, opened 2014-12-28
__eq__ overload not working


ghost:
If I have the following code:

```
class A:
    def __eq__(self, other):
        return True

a0 = A()
a1 = A()
print a0 == a1
```

Pyston will print `False` while CPython will print `True`, why?

kmod:
Ok I opened #254 to track this and the other missing ones, thanks for bringing it up!

PR #380, opened 2015-03-14
In order to avoid die with SIGINT during REPL mode and fix bug '__eq__ overload not working #253'

corona10:
In order to avoid die with SIGINT during REPL mode.
write a function on jit.cpp.
and change data structure of free_chunks from std::vector to llvm::SmallVector
because of performance issue on vector of fixed size
same issue was discussed on discuss #337.

and for the last fix bug '__eq__ overload not working #253'

that's it!!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - sudo add-apt-repository --yes ppa:ubuntu-toolchain-r/test
   - sudo add-apt-repository --yes ppa:kubuntu-ppa/backports
   - sudo apt-get -qq update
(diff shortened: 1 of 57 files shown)


kmod:
I would recommend creating multiple branches in your github repo so that you can have multiple pull requests outstanding (I think the commands I gave on the other PR would achieve this).

I think the instance `__eq__` stuff mostly looks good, but we definitely would want some tests for this.  Can you add some code to test/tests/oldstyle_classes.py that tests the behavior for both whether or not `__eq__` is defined?  I think your current code might have an issue with trying to call a NULL function if `__eq__` is not defined (in which case it should fall back to an object pointer comparison).

PR #376, opened 2015-03-13
Implement further changes required for virtualenv

undingen:

PR #381, opened 2015-03-16
get iobench.py working

toshok:
a few small additions to get iobench.py working.

The `__doc__` change (46773ef) is wrong for a couple reasons, the biggest of which is that it's only set in the interpreter: The `-n` command line flag results in no doc strings.  Where should this go instead?  `wrapFunction`?

### DIFF
--- a/from_cpython/Include/pyconfig.h
+++ b/from_cpython/Include/pyconfig.h
@@ -54,6 +54,7 @@
 #define HAVE_SOCKADDR_STORAGE 1
 #define HAVE_SOCKETPAIR 1
 #define HAVE_GETPEERNAME 1
+#define HAVE_TIMES 1
 
 #define PY_FORMAT_LONG_LONG "ll"
 #define PY_FORMAT_SIZE_T "z"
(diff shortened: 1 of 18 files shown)


undingen:
Concerning the file.seek() and file.tell(): I have locally added the cpython implementation of this two functions, (besides being much more complicated) it looks like they handle ```f_skipnextlf```. I'm not sure if this is something we need to support?(https://gist.github.com/undingen/627b09dd6f908638bcc2 )

toshok:
I'd prefer the cpython implementations, definitely.  I'll remove that
src/runtime change from my PR, and we can gate my PR on when your change
makes it in.

undingen:
Ok I have added the change (slightly modified) to #376.

kmod:
Ok, I merged #376

toshok:
this change should be ready to go other than the fact that iobench.py dirties up the pyston/ directory with temp files.  similar to 4e9effb, maybe we need makefile bits to run minibenchmarks in tempdirs.

PR #384, opened 2015-03-17
fix bug 'dying with SIGINT when REPL mode'

corona10:
fix bug 'dying with SIGINT when REPL mode'

### DIFF
--- a/src/runtime/dict.cpp
+++ b/src/runtime/dict.cpp
@@ -411,11 +411,7 @@ Box* dictNonzero(BoxedDict* self) {
     return boxBool(self->d.size());
 }
 
-Box* dictFromkeys(BoxedDict* self, Box* iterable, Box* default_value) {
-    if (!isSubclass(self->cls, dict_cls))
-        raiseExcHelper(TypeError, "descriptor 'fromkeys' requires a 'dict' object but received a '%s'",
-                       getTypeName(self));
-
+Box* dictFromkeys(Box* cls, Box* iterable, Box* default_value) {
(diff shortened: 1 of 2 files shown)


kmod:
Hi @corona10, we definitely need to clean up the commit history on this :)  There are better git tutorials on the internet than I'll be able to provide here, but personally I would recommend doing this:
```
git checkout -b more_smallvector upstream/master
git checkout master .
git commit
```
This creates a new branch starting at upstream/master, and then updates your files to be the same as on the master branch (but without including any of the history of that branch).  This has the effect of squashing all of the changes together into a single commit.  (If you want a more complicated result, such as multiple commits, I personally use and would recommend the `git rebase -i` command.)

After doing this, you can do `git push -u origin/master more_smallvector` which will push the changes to github and you can update the pull request.

kmod:
But as for the content of the pull request, we already have a [SIGINT handler](https://github.com/dropbox/pyston/blob/master/src/codegen/entry.cpp#L176).  The existing one is definitely over-simplified (just calls abort()), but the correct handling of sigint is actually more complicated than either what we have or what you added: it's supposed to raise a KeyboardInterrupt exception *on the main thread*.

This is mostly handled by the fact that we pulled in the signalmodule.c code from CPython, which takes care of this, but we're missing a few key pieces of functionality before we can start using that.  Until then we might as well keep the broken handler we currently have :)

The SmallVector change looks good!  We can merge that if you want to separate that out and squash the commits.

corona10:
Thank you i got some lots of tips about git
I did as you explained it :-)

PR #388, opened 2015-03-18
Add support for unicode literals when using pypa

undingen:
We support the \u, \U and \N escape sequences and
the unicode_literals option.
In addition updates pypa to latest version with unicode support.


kmod:
Nice!  It might be time for us to switch to pypa as the default now.

PR #392, opened 2015-03-19
split the getattr/setattr of softspace into two try blocks

toshok:
cpython can be observed setting softspace on an object even if it doesn't start with it.
the getattrInternal in softspace() throws an exception the first time through since the
attribute isn't present.  if we don't turn around and set it on the object, we throw
every time we print, which causes a pretty large perf regression in fasta (from ~6 seconds to ~17 seconds.)


kmod:
good find :)

PR #393, opened 2015-03-19
update INSTALLING.md, and libsqlite-dev in cmake section

aisk:
I created a new virtualbox ubuntu14.04 machine, and found `libsqlite3-dev` is needed, but not been installed on the installing document.


toshok:
Darn - sorry about that, I added it to the wrong section (Clang Compiler) and not to the host specific configs below.  Mind removing it from that section and updating this PR so we can roll everything into 1 change?

toshok:
thanks!

PR #395, opened 2015-03-19
Cache pypa parsed AST to disk

undingen:
before ```import pip``` -I took about 2.5sec on the second run, with this change it takes about 1.5sec

While the pypa parser is faster then our current cpython approach, caching the parse result is a significant speed improvement for large scripts. (import pip without cached files takes 7.5sec (cpython parser) vs. 2.5secs (pypa))
- I used the same binary format but choose a different magic string otherwise our pypa/cpython parser test would be useless because the result of the first run would be used instead of calling the parser.
- I'm unhappy with the code duplication :-( on the other-hand this is probably something we don't touch often. I could not use the normal AST recursive visit pattern because it does not visit the nodes in the order which our cache format requires.

### DIFF
--- a/src/codegen/parser.cpp
+++ b/src/codegen/parser.cpp
@@ -25,6 +25,7 @@
 #include "llvm/Support/Path.h"
 
 #include "codegen/pypa-parser.h"
+#include "codegen/serialize_ast.h"
 #include "core/ast.h"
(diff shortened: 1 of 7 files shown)


toshok:
:+1:

PR #398, opened 2015-03-20
Handling exceptions in with blocks

rntz:
We still have that bug having to do with invoke statements and definedness-analysis, but that bug was around even before I got with-blocks to handle exceptions, it just wasn't triggered as much. All the old tests pass, exceptions_test is now passing, and I've added several more tests for bugs I've encountered or introduced & fixed.

I'm still going to hunt down that bug with definedness analysis, but I figure we might as well merge this now and treat that as a separate bug.


rntz:
Hm, the Travis CI build is failing due to unexpected successes in some of the new tests I added that I marked as failing due to the definedness-analysis-bug. I'll look into it.

kmod:
Can we squash some of these commits?  Also it seems like the tests you're mentioning are probably pretty specific to the particular tiering arguments, so you can try playing around with `skip-if` or `fail-if` annotations.

src/codegen/irgen.cpp
```diff
@@ -785,14 +859,13 @@ static void emitBBs(IRGenState* irstate, TypeAnalysis* types, const OSREntryDesc
...
             }
-
-            InternedString is_defined_name = getIsDefinedName(it->first, source->getInternedStrings());
         }
         for (auto t : phi_args) {
             std::get<0>(t)->addIncoming(std::get<1>(t), std::get<2>(t));
         }
     }
 
+    // TODO(rntz): what are we doing in this for-loop?
```
kmod: This is just trying to clean up some of the codegen memory -- `decvrefNodrop()` just drops a "vref" which will eventually delete the object once it goes to zero.  The memory management can be improved a lot but hopefully that's not what you're trying to get at with this question :)

src/core/ast.h
```diff
@@ -1021,14 +1021,14 @@ class AST_LangPrimitive : public AST_expr {
...
-        LANDINGPAD,
+        LANDINGPAD, // grabs the info about the last raised exception
         LOCALS,
         GET_ITER,
         IMPORT_FROM,
         IMPORT_NAME,
         IMPORT_STAR,
         NONE,
-        NONZERO,
+        NONZERO, // Python's __nonzero__ magic method
```
kmod: This is close but not quite true :P  (it's more analogous to the `bool()` function)  if you're curious you can check out objmodel.cpp::nonzero() to see how the semantics are slightly different.

src/core/cfg.cpp
```diff
@@ -2274,105 +2222,197 @@ class CFGVisitor : public ASTVisitor {
...
+        CFGBlock* exit_block = cfg->addDeferredBlock();
+        exit_block->info = "with_exit";
 
         pushAssign(ctxmgrname, remapExpr(node->context_expr));
 
-        AST_expr* enter
-            = makeLoadAttribute(makeName(ctxmgrname, AST_TYPE::Load, node->lineno), internString("__enter__"), true);
-        AST_expr* exit
-            = makeLoadAttribute(makeName(ctxmgrname, AST_TYPE::Load, node->lineno), internString("__exit__"), true);
+        // TODO(rntz): for some reason this acts like it's "mgr.__exit__" instead
```
kmod: So the pseudocode they gave is not quite right in this corner case:
```python
class M(type):
    def __exit__(*args):
        print args
        return False

class C(object):
    __metaclass__ = M

def f(o):
    print hasattr(type(o), "__exit__")
    # type(o).__exit__ exists but this with statement will still throw an "AttributeError: __exit__":
    with o:
        pass

f(C())
```

Even though type(o) has an `__exit__` method, the lookup is actually happening using Python's "special method lookup rules", which is that it looks it up on the class but not the instance *or the metaclass*.  We usually represent this as a "class-only attribute lookup" (what the last argument as `true` means).  I'm not sure there's any way to represent this kind of lookup from Python code, which is why we added a special "ClsAttribute" type of ast node.

src/core/cfg.cpp
```diff
@@ -2274,105 +2222,197 @@ class CFGVisitor : public ASTVisitor {
...
+            pushJump(finally_block);
+        }
+
+        // This tells us whether our exit_block will have multiple incoming edges (one from exc_block, one from
+        // finally_block) or not (if exc_block and/or finally_block is unneeded, or if finally_block doesn't jump to
+        // exit_block because it's never entered except via continue, break, and/or return). If exit_block has multiple
+        // incoming edges, then anybody *branching* into it (rather than jumping directly) needs to do so through a
+        // "trampoline" block to avoid critical edges.
+        // bool exit_multi_incoming = exc_block->predecessors.size() && (finally_did_why & (1 << Why::FALLTHROUGH));
+        bool exit_multi_incoming = true; // FIXME FIXME FIXME(rntz)
```
kmod: We also have a unneccessary-block-cleanup phase at the end, so while it's nice if we can avoid generating them in the first place, I think it might be ok to just always generate the trampolines?

src/core/cfg.cpp
```diff
@@ -2274,105 +2222,197 @@ class CFGVisitor : public ASTVisitor {
...
+            // TODO(rntz): test for this case
+            delete exc_block;
+        } else {
+            cfg->placeBlock(exc_block);
+            curblock = exc_block;
+
+            // call the context-manager's exit method
+            InternedString suppressname = nodeName(node, "suppress");
+            pushAssign(suppressname, makeCall(makeLoad(exitname, node), makeLoad(exc_type_name, node),
+                                              makeLoad(exc_value_name, node), makeLoad(exc_traceback_name, node)));
```
kmod: I think you may need to call `callNonzero()` on the result of makeCall -- it looks like `__exit__` is allowed to return a non-boolean type, and at some point we have to coerce that to an unboxed bool, which is what the NONZERO ast node is for.

src/runtime/types.cpp
```diff
@@ -382,6 +382,9 @@ std::string BoxedModule::name() {
     }
 }
 
+// This mustn't throw; our IR generator generates calls to it without "invoke" even when there are exception handlers /
+// finally-blocks in scope.
+// TODO: should we use C++11 `noexcept' here?
```
kmod: good idea

src/core/ast.h
rntz: Is it accurate to say that NONZERO determines whether something is "true" for purposes of executing `if` statements?

src/core/cfg.cpp
rntz: `pushBranch` calls `makeBranch` which calls `callNonzero`.

src/core/ast.h
kmod: Yeah I guess so, and other similar cases (while loops, boolean expressions, whether the `__exit__` method caught the exception).

rntz:
@kmod This is failing `weakref4.py` on `pyston_defaults`. I'm not sure why - I haven't touched any of the GC code. Other than that it looks good to go.

undingen:
I think @toshok had a similar problem when running the test with the interpreter?!?

src/core/ast.h
```diff
@@ -951,6 +953,31 @@ class AST_Yield : public AST_expr {
...
+    AST_FunctionDef* function_def;
+
+    virtual void accept(ASTVisitor* v);
+    virtual void* accept_expr(ExprVisitor* v);
+
+    AST_MakeFunction(AST_FunctionDef* fd)
+        : AST_expr(AST_TYPE::MakeFunction, fd->lineno, fd->col_offset), function_def(fd) {}
+
+    static const AST_TYPE::AST_TYPE TYPE = AST_TYPE::MakeFunction;
+};
```
kmod: I like this approach for getting these statements into assigns :)

kmod:
Cool, I think I figured out why weakref4 was failing, so I merged this in (it's passing for me locally).  thanks for all the fixes!

PR #399, opened 2015-03-21
cmake automatically handle llvm_up

dagar:


dagar:
Rolling this into PR #400

PR #400, opened 2015-03-22
[WIP] cmake use prebuilt llvm

dagar:
This PR modifies cmake to allow Pyston to use a previously built LLVM (with patches).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "libunwind"]
 	path = libunwind
 	url = git://git.sv.gnu.org/libunwind.git
+	ignore = dirty
 [submodule "libpypa"]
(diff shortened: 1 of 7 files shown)


dagar:
Using a prebuilt LLVM saves travis about 15 minutes of compile time + 4 minutes not having to clone and configure LLVM.

@kmod Could you provide access to something like s3 for uploading LLVM binaries? Travis allows you to encrypt the credentials. http://docs.travis-ci.com/user/encryption-keys/

dagar:
Closing this because it doesn't save much time on top of just using ccache.

The cmake changes in this PR will be worth revisiting if it ever becomes possible to use LLVM without applying pyston patches.

PR #401, opened 2015-03-23
improve performace of finding chunk

corona10:
i think it might better is there anything can be problem??


corona10:
i think that recent code will be better when thinking of mark and sweep phase

i cancel my pull request

PR #405, opened 2015-03-23
cmake fix libunwind flags

dagar:
-don't pass empty CFLAGS='' when buliding libunwind in release mode
-revert libunwind submodule to older 65ac867416 (same as INSTALLING.md)


kmod:
I'm not sure how I feel about reverting what's supposed to be a fix, but I guess if we're already running with this rev for the makefile build it doesn't make things any worse to make cmake consistent :)

kmod:
(talking about changing the libunwind rev)

PR #407, opened 2015-03-24
Some eval and exec improvements

tjhance:
Depends on #349 (*cough*)

This PR
- Implements `locals()` (more) correctly for module-level scope by just returning the same things as `globals()` (the `attrwrapper` thing)
- Since this is passed to `eval` or `exec`, I have to add support for `boxedLocals` being an arbitrary object, not necessarily a `dict`
- Fix the scope rules for the top-level of `eval` and `exec`--previously, these were treated like a function, but they should be treated more like a module scope and independently descending into the child scopes.

All these changes together let us uncomment everything in `eval_test.py` and have it pass.

(This has all of #349 in it, so just look at the last 2 commits)

### DIFF
--- a/src/analysis/function_analysis.cpp
+++ b/src/analysis/function_analysis.cpp
@@ -340,6 +340,8 @@ public:
         return true;
     }
 
+    virtual bool visit_exec(AST_Exec* node) { return true; }
+
     friend class DefinednessBBAnalyzer;
 };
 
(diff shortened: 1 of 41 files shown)


kmod:
Github didn't automatically rebase the PR history with the new changes so it's hard to tell what just got merged here, but I merged and checked it on the command line and lgtm.

tjhance:
It's the last 2 commits, c3b3f6a and d3e045a, for the curious

PR #408, opened 2015-03-24
Use cpythons isinstance and issubclass implementation

undingen:
This commit removes our implementation of this functions,
because they could not handle mixing of old and new style classes.
And having two implementations with very similar name but different
results is very confusing.

Issues with this patch:
- in order to not have to change all our ```isSubclass()``` checks to the fast version, added if checks for the fast versions. They will not call ```__subclasscheck__```....
- I left the 3args ```isinstance``` function as is, because I'm not sure what exactly the excepected behaviour is. Maybe I should rename the ```AST_LangPrimitive::ISINSTANCE``` to ```AST_LangPrimitive::EXCEPTION_MATCHES``` and repalce the implementation with one which calls ```PyErr_GivenExceptionMatches()``` internally?
- The additional ```__subclasschecks__``` and ```__instancecheck__``` lookups decrease performance.

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -212,18 +212,17 @@ extern "C" void my_assert(bool b) {
 }
 
 bool isInstance(Box* obj, BoxedClass* cls) {
-    return isSubclass(obj->cls, cls);
+    int rtn = _PyObject_RealIsInstance(obj, cls);
+    if (rtn < 0)
+        checkAndThrowCAPIException();
+    return rtn;
 }
 
 extern "C" bool isSubclass(BoxedClass* child, BoxedClass* parent) {
-    // TODO the class is allowed to override this using __subclasscheck__
-    assert(child->tp_mro);
-    assert(child->tp_mro->cls == tuple_cls);
-    for (auto b : static_cast<BoxedTuple*>(child->tp_mro)->elts) {
-        if (b == parent)
-            return true;
-    }
-    return false;
+    int rtn = _PyObject_RealIsSubclass(child, parent);
+    if (rtn < 0)
+        checkAndThrowCAPIException();
+    return rtn;
 }
 
 extern "C" void assertFail(BoxedModule* inModule, Box* msg) {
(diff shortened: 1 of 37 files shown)


src/runtime/objmodel.cpp
```diff
@@ -212,18 +212,47 @@ extern "C" void my_assert(bool b) {
...
+    FAST_CLS_CHECK(int_cls, Py_TPFLAGS_INT_SUBCLASS);
+    FAST_CLS_CHECK(long_cls, Py_TPFLAGS_LONG_SUBCLASS);
+    FAST_CLS_CHECK(str_cls, Py_TPFLAGS_STRING_SUBCLASS);
+    FAST_CLS_CHECK(unicode_cls, Py_TPFLAGS_UNICODE_SUBCLASS);
+    FAST_CLS_CHECK(tuple_cls, Py_TPFLAGS_TUPLE_SUBCLASS);
+    FAST_CLS_CHECK(list_cls, Py_TPFLAGS_LIST_SUBCLASS);
+    FAST_CLS_CHECK(dict_cls, Py_TPFLAGS_DICT_SUBCLASS);
+#undef FAST_CLS_CHECK
+
+    int rtn = PyObject_IsSubclass(child, parent);
```
kmod: hmm I think the intention of isSubclass is more similar to PyType_IsSubtype (only checks new style classes, doesn't check `__subclasscheck__`) than to PyObject_IsSubclass.

kmod:
So the use of `__subclasscheck__` is extremely limited in CPython (at least in 2.7) -- I believe it's only used for the `issubclass()` builtin and for checking whether exceptions match the specifier.  So almost all of the uses can just be the faster, new-class-only, non-overridable version :)

I think the 3-arg `isinstance()` is because while the normal isinstance() throws an exception if you don't pass a class, you can do something like

```python
try:
  raise Exception()
except 1:
  pass
```

which won't catch the exception but also won't throw an exception from the exception matching.  But it's not that exception-matching doesn't throw exceptions, it's only for this part I believe.

undingen:
- I'm now calling ```_PyObject_RealIsInstance``` inside ```isInstance``` (similar change to ```isSubclass```) which does the check without calling ```__instancecheck__```.
- I switched our ```AST_LangPrimitive::ISINSTANCE``` to ```AST_LangPrimitive::CHECK_EXC_MATCH``` and modified it to generate a ```PyErr_GivenExceptionMatches``` call. With this change we now pass most of the ```exception_subckasscheck.py```, I moved the small part which is not passing to a new file. (The issue is that CPython calls ```PyErr_NormalizeException``` when setting the exception and this causes a second ```__subclasscheck__``` to happen.)

kmod:
Cool :)  Is there a reason to call `_PyObject_RealIsSubclass` and not `PyType_IsSubtype` from `isSubclass()`?

undingen:
To be honest I'm not really sure, I thought I would use ```_PyObject_RealIsSubclass``` because it calls into ```PyType_IsSubtype``` and ```PyClass_IsSubclass``` and therefore handles old and new style classes. But I'm not sure if we even want that it to also supporting old style classes?

kmod:
I think the signature of `isSubclass` already enforces that the arguments are new-style classes?

PR #411, opened 2015-03-25
colocate string and tuple data

toshok:
There's a lot of commits here, but most of them are squashable.  The crux of the change is both BoxedTuple and BoxedString have specialized `operator new`s that calculate the size of the Boxed structure as well as any additional data (Box*'s for the BoxedTuple, chars for BoxedString).

This should reduce the number of calls to our allocator by at least a half for these types.  In order to keep more of the BoxedString code the same, I added `pyston::String` which can mostly be described as "an llvm::StringRef that exposes a std::string api".  One difference is that `pyston::String`'s data can be owned (malloced/freed by pyston::String), which happens using `operator+`, etc.  It's normal mode is unowned (its use in BoxedString).

Also some effort goes to calculating the number of elements (tuple elements, characters) in the result before doing the allocation for the box, so we don't have to realloc.  This means we need non-ctor allocation paths (we already had them for BoxedString, in the boxString* family).  For BoxedTuple, I added static BoxedTuple::Create() overloads.

performance numbers initially were pretty crazy (-10% for interp2), but changes on master have brought them down to a more modest level (I haven't investigated to see which change(s) on master caused this.)  Current comparison shows:

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.0%)
              pyston interp2.py                         :    3.8s baseline: 3.9 (-2.5%)
              pyston raytrace.py                        :    6.4s baseline: 6.3 (+1.0%)
              pyston nbody.py                           :    8.7s baseline: 8.5 (+2.7%)
              pyston fannkuch.py                        :    6.6s baseline: 6.6 (+0.4%)
              pyston chaos.py                           :   20.1s baseline: 19.9 (+1.4%)
              pyston spectral_norm.py                   :   21.1s baseline: 22.5 (-6.2%)
              pyston fasta.py                           :    4.9s baseline: 5.3 (-6.9%)
              pyston pidigits.py                        :    5.1s baseline: 5.0 (+1.6%)
              pyston richards.py                        :    1.9s baseline: 1.8 (+1.0%)
              pyston deltablue.py                       :    2.0s baseline: 2.0 (-0.2%)
              pyston (geomean-0b9f)                     :    6.0s baseline: 6.0 (-0.8%)
```

### DIFF
--- a/src/capi/object.cpp
+++ b/src/capi/object.cpp
@@ -421,9 +421,9 @@ extern "C" PyObject* PyObject_SelfIter(PyObject* obj) noexcept {
 extern "C" int PyObject_GenericSetAttr(PyObject* obj, PyObject* name, PyObject* value) noexcept {
     try {
         if (value == NULL)
-            delattrGeneric(obj, static_cast<BoxedString*>(name)->s, NULL);
+            delattrGeneric(obj, std::string(static_cast<BoxedString*>(name)->s), NULL);
         else
-            setattrGeneric(obj, static_cast<BoxedString*>(name)->s.c_str(), value, NULL);
+            setattrGeneric(obj, std::string(static_cast<BoxedString*>(name)->s), value, NULL);
     } catch (ExcInfo e) {
         setCAPIException(e);
         return -1;
--- a/src/runtime/import.cpp
+++ b/src/runtime/import.cpp
@@ -166,7 +166,7 @@ SearchResult findModule(const std::string& name, const std::string& full_name, B
         auto path_pass = path_list ? path_list : None;
         Box* loader
             = callattr(finder, &find_module_str, CallattrFlags({.cls_only = false, .null_on_nonexistent = false }),
-                       ArgPassSpec(2), new BoxedString(full_name), path_pass, NULL, NULL, NULL);
+                       ArgPassSpec(2), boxString(full_name), path_pass, NULL, NULL, NULL);
 
         if (loader != None)
             return SearchResult(loader);
@@ -195,7 +195,7 @@ SearchResult findModule(const std::string& name, const std::string& full_name, B
         BoxedString* p = static_cast<BoxedString*>(_p);
 
         joined_path.clear();
-        llvm::sys::path::append(joined_path, p->s, name);
+        llvm::sys::path::append(joined_path, std::string(p->s), name);
         std::string dn(joined_path.str());
 
         llvm::sys::path::append(joined_path, "__init__.py");
@@ -209,7 +209,7 @@ SearchResult findModule(const std::string& name, const std::string& full_name, B
             auto path_pass = path_list ? path_list : None;
             Box* loader = callattr(importer, &find_module_str,
                                    CallattrFlags({.cls_only = false, .null_on_nonexistent = false }), ArgPassSpec(2),
-                                   new BoxedString(full_name), path_pass, NULL, NULL, NULL);
+                                   boxString(full_name), path_pass, NULL, NULL, NULL);
             if (loader != None)
                 return SearchResult(loader);
         }
@@ -218,7 +218,7 @@ SearchResult findModule(const std::string& name, const std::string& full_name, B
             return SearchResult(std::move(dn), SearchResult::PKG_DIRECTORY);
 
         joined_path.clear();
-        llvm::sys::path::append(joined_path, p->s, name + ".py");
+        llvm::sys::path::append(joined_path, std::string(p->s), name + ".py");
         fn = joined_path.str();
 
         if (pathExists(fn))
@@ -250,7 +250,7 @@ static Box* getParent(Box* globals, int level, std::string& buf) {
         if (pkgname->cls != str_cls) {
             raiseExcHelper(ValueError, "__package__ set to non-string");
         }
-        size_t len = pkgname->s.size();
+        size_t len = pkgname->size();
         if (len == 0) {
             if (level > 0) {
                 raiseExcHelper(ValueError, "Attempted relative import in non-package");
@@ -270,7 +270,7 @@ static Box* getParent(Box* globals, int level, std::string& buf) {
         Box* modpath = getattrInternal(globals, path_str, NULL);
         if (modpath != NULL) {
             /* __path__ is set, so modname is already the package name */
-            if (modname->s.size() > PATH_MAX) {
+            if (modname->size() > PATH_MAX) {
                 raiseExcHelper(ValueError, "Module name too long");
             }
             buf += modname->s;
@@ -357,7 +357,7 @@ static Box* importSub(const std::string& name, const std::string& full_name, Box
             const static std::string load_module_str("load_module");
             module = callattr(sr.loader, &load_module_str,
                               CallattrFlags({.cls_only = false, .null_on_nonexistent = false }), ArgPassSpec(1),
-                              new BoxedString(full_name), NULL, NULL, NULL, NULL);
+                              boxString(full_name), NULL, NULL, NULL, NULL);
         } else
             RELEASE_ASSERT(0, "%d", sr.type);
 
@@ -441,9 +441,8 @@ static bool loadNext(Box* mod, Box* altmod, std::string& name, std::string& buf,
 }
 
 static void ensureFromlist(Box* module, Box* fromlist, std::string& buf, bool recursive);
-Box* importModuleLevel(std::string* name, Box* globals, Box* from_imports, int level) {
+Box* importModuleLevel(const std::string& name, Box* globals, Box* from_imports, int level) {
     bool return_first = from_imports == None;
-    assert(name);
 
     static StatCounter slowpath_import("slowpath_import");
     slowpath_import.log();
@@ -453,7 +452,9 @@ Box* importModuleLevel(std::string* name, Box* globals, Box* from_imports, int l
     if (!parent)
         return NULL;
 
-    std::string _name = *name;
+    // make a copy of the string we pass in, since loadNext while
+    // modify the string as we load the modules along the path.
+    std::string _name = name;
 
     Box* head;
     bool again = loadNext(parent, level < 0 ? NULL : parent, _name, buf, &head);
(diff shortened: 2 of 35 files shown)


src/capi/object.cpp
```diff
@@ -146,7 +146,7 @@ extern "C" PyObject* PyObject_SelfIter(PyObject* obj) noexcept {
 extern "C" int PyObject_GenericSetAttr(PyObject* obj, PyObject* name, PyObject* value) noexcept {
     try {
         if (value == NULL)
-            delattrGeneric(obj, static_cast<BoxedString*>(name)->s, NULL);
```
kmod: It'd be great to have functions start taking llvm::StringRef or Pyston::String rather than allocating a fresh std::string (doesn't have to be in this PR).

src/runtime/types.h
```diff
@@ -435,14 +455,63 @@ class BoxedList : public Box {
...
+        return rtn;
+    }
+
+    void* operator new(size_t size, BoxedClass* cls, size_t nelts) __attribute__((visibility("default"))) {
+        Box* rtn = static_cast<Box*>(cls->tp_alloc(cls, nelts));
+        rtn->cls = cls;
+        return rtn;
+    }
+
+    static BoxedTuple* Create(int64_t size) { return new (size) BoxedTuple(size); }
```
kmod: We've usually been using camelCase for function names (ie this would be `create`).

src/core/string.h
```diff
@@ -0,0 +1,231 @@
...
+#ifndef PYSTON_CORE_STRING_H
+#define PYSTON_CORE_STRING_H
+
+#include <stdexcept>
+
+namespace pyston {
+
+class String {
+public:
+    explicit String(const std::string& s, char* _ptr = NULL) : ptr(_ptr), own(_ptr == NULL) {
```
kmod: Can we get a comment about the ownership strategy?  I was somewhat confused by the way this works, since I was assuming it would work similarly to llvm::StringRef in the non-owning case (ie would just reference the original data rather than copy it).

src/capi/object.cpp
toshok: agreed.  we could also support llvm::Twine's, though those aren't as handy if callers aren't calling with concatenated strings.

src/core/string.h
toshok: I found and fixed the cases where we were creating temporaries (operator+, operator<<), and switched those instances to using llvm::Twines and llvm::raw_string_ostream.  The latter lets us use llvm::StringRef directly, and the former delays the concatenation until the place where the value is used (in the new `boxStringTwine` method).  Now there are no cases where pyston::String owns the buffer, so it shrinks by a bool, the ctors get simpler, and there's no dtor.

toshok:
this will definitely need squashing, but I think everything's been addressed/rebased off master.

toshok:
hrm, actually after the last rebase the django migrate test is crashing... ugh.

toshok:
actually it appears to be the switch to -x by default that's crashing the django migrate test.

undingen:
I could reproduce the libpypa bug with django on current head, I will look into it.

src/runtime/util.cpp
```diff
@@ -39,32 +39,32 @@ void parseSlice(BoxedSlice* slice, int size, i64* out_start, i64* out_stop, i64*
     int64_t istop;
     int64_t istep = 1;
 
-    if (isSubclass(step->cls, int_cls)) {
+    if (step->cls != none_cls && (step->cls == int_cls || isSubclass(step->cls, int_cls))) {
```
kmod: ?

test/tests/weakref1.py
```diff
@@ -1,3 +1,4 @@
+# skip-if: '-n' not in EXTRA_JIT_ARGS
```
kmod: Are these tests working now with the changes I made recently?

src/core/string.h
```diff
@@ -0,0 +1,169 @@
...
+#ifndef PYSTON_CORE_STRING_H
+#define PYSTON_CORE_STRING_H
+
+#include <stdexcept>
+
+namespace pyston {
+
+class String {
+public:
+    explicit String(const std::string& s, char* _ptr) : ptr(_ptr) {
```
kmod: I still think it's a weird mix of concerns that the storage pointer must be provided by the caller, but the constructor takes care of the copying.  It kind of feels like we just took some of the BoxedString functionality and put it in another class... I guess I'm glad we tried to make String reusable, but I just have a hard time seeing where this would be used other than BoxedString.

What if we have the BoxedString constructor do this copying, and give it a llvm_str function that returns a llvm::StringRef?

src/runtime/util.cpp
toshok: ah I'd added that just to reduce calls to isSubclass (which were rather hot in `make perf` output while I was working on this.)  given @undingen's changes this should be unnecessary.

test/tests/weakref1.py
toshok: I thought I'd removed those.  they're working.

src/core/string.h
toshok: I can definitely move the copying from pyston::String to BoxedString.  I'll look at how large the changes are to removing pyston::String and replacing it with the llvm alternatives.

src/runtime/str.cpp
```diff
@@ -46,16 +46,33 @@ extern "C" PyObject* string_splitlines(PyStringObject* self, PyObject* args) noe
...
-BoxedString::BoxedString(const char* s, size_t n) : s(s, n) {
-    gc::registerGCManagedBytes(this->s.size());
+BoxedString::BoxedString(const char* s, size_t n) : s(storage(), n) {
+    if (s) {
+        memmove(storage(), s, n);
+        storage()[n] = 0;
+    } else {
+        memset(storage(), 0, n + 1);
+    }
+    gc::registerGCManagedBytes(cls->tp_basicsize + this->s.size() + 1);
```
kmod: I don't think we have to call this any more; it was there since the gc wouldn't see the allocation that the s constructor (as a std::string) would do.  I think this should get seen now by the allocation for the object in operator new?

src/runtime/import.cpp
```diff
@@ -441,9 +441,8 @@ static bool loadNext(Box* mod, Box* altmod, std::string& name, std::string& buf,
 }
 
 static void ensureFromlist(Box* module, Box* fromlist, std::string& buf, bool recursive);
-Box* importModuleLevel(std::string* name, Box* globals, Box* from_imports, int level) {
+Box* importModuleLevel(std::string& name, Box* globals, Box* from_imports, int level) {
```
kmod: should probably be a const reference?

src/runtime/import.cpp
```diff
@@ -453,7 +452,7 @@ Box* importModuleLevel(std::string* name, Box* globals, Box* from_imports, int l
     if (!parent)
         return NULL;
 
-    std::string _name = *name;
+    std::string _name = name;
```
kmod: heh let's kill this, not sure why we have it.

src/runtime/types.h
```diff
@@ -371,14 +375,31 @@ class BoxedBool : public BoxedInt {
 
 class BoxedString : public Box {
 public:
-    // const std::basic_string<char, std::char_traits<char>, StlCompatAllocator<char> > s;
-    std::string s;
+    llvm::StringRef s;
+
+    char* storage() { return (char*)this + cls->tp_basicsize; }
```
kmod: Hmm so some places call storage(), and some look at `->s.data()`.  I guess if we're going to store the pointer we should prefer to use it?  Maybe we could do something like have a private function that calculates the storage offset, and a public one that just returns s.data?

src/runtime/types.h
```diff
@@ -371,14 +375,31 @@ class BoxedBool : public BoxedInt {
...
+
+    void* operator new(size_t size, size_t ssize) __attribute__((visibility("default"))) {
+        Box* rtn = static_cast<Box*>(gc_alloc(str_cls->tp_basicsize + ssize + 1, gc::GCKind::PYTHON));
+        rtn->cls = str_cls;
+        return rtn;
+    }
 
+    void* operator new(size_t size, BoxedClass* cls, size_t ssize) __attribute__((visibility("default"))) {
+        // XXX this is likely broken...
```
kmod: (why?)

src/runtime/util.cpp
```diff
@@ -46,25 +46,25 @@ void parseSlice(BoxedSlice* slice, int size, i64* out_start, i64* out_stop, i64*
         }
     }
 
-    if (isSubclass(start->cls, int_cls)) {
-        istart = static_cast<BoxedInt*>(start)->n;
-        if (istart < 0)
-            istart = size + istart;
-    } else {
+    if (start->cls == none_cls || (start->cls != int_cls && !isSubclass(start->cls, int_cls))) {
```
kmod: Do we need these changes?  it looks like it still has the same behavior but I hate touching the slice parsing code...

kmod:
lgtm!  lots of opportunities to reduce string copies after this :)

could you squash before pushing?  also if it's not too hard, implementing _PyTuple_Resize would be great as well (I needed an implementation for my branch but it will break with this change).

src/runtime/util.cpp
toshok: I'd like to spend some time with them in the future (once verifying that they're still needed after the `isSubclass` work), but definitely not something needed for this change.

src/runtime/str.cpp
toshok: yup, the various arena `alloc` functions already register.  so with these lines we double count string memory.

src/runtime/import.cpp
toshok: this one is necessary (we need the copy) because loadNext actually modifies _name.

src/runtime/types.h
toshok: ah, because I hadn't verified that `tp_itemsize == sizeof(char)` in cpython.  just did that, and removing comment.

toshok:
I've squashed everything except the final `->storage()` change.  will squash again before merging.

we can't directly use `->s.data()` everywhere (at least without `const_cast`ing it), so I've made the following changes:

1. `::storage()` is private, and only used in ctors when initializing the `llvm::StringRef`
2. added `::data()` which returns `s.data()`, but casted to `char*`
3. added `::size()` for consistency with `::data()` - just returns `s.size()`

PR #413, opened 2015-03-26
Use the pypa parser as the default parser.

undingen:
Switches the meaning of -x.
In addition update to latest pypa version

@tjhance: I haven't tried it with your ```exec``` support patch. But I hope I fixed the missing newline issue you run into.  Please let me know if you still encounter problems.
@vinzenz: Thanks for writing the parser!


kmod:
Cool!  Thanks @vinzenz :)

PR #415, opened 2015-03-27
statically laid out closures

tjhance:
This needs a lot of cleaning up but I wanted to solicit feedback on the approach first...

Right now we handle closure as normal objects with `HCAttrs`, and we de-reference variables from closures by walking up the closures until we find one with the variable name. This is overly-dynamic to the point of being wrong (see my new `static_closure_locations.py` test case):

```python
# The use of c makes sure a closure gets passed through all 4 functions.
# The use of a in g makes sure that a is in f's closure.
# The a in j should refer to the a in h, thus throwing an exception since
# it is undefined (that is, it should *not* access the a from f even
# though it is accessible via the closure).

def f():
    c = 0 
    a = 0 
    def g():
        print c
        print a
        def h():
            print c
            def j():
                print c
                print a    # should throw a NameError

            j() 
            a = 1 
        h() 
    g() 
f()
```

Since the the set of names is a closure is fixed, in this diff I statically determine the size of the closure and the mapping of names to offsets (letting undefined variables be `NULL`). Then the IR for a closure lookup just consists of some pointer-dereferences with constant offsets. I allocate the array of `Box *`s inside the `BoxedClosure` following @toshok 's lead with tuples in #411.

The offset information is computed statically in the `ScopeInfo`, and for any name you can look up a `DerefInfo`, which tells you exactly how many parent closures you need to walk up and the offset within that closure the desired variable lies. (There are a few other functions which return related information.) Here's some example IR:

```
  %8 = getelementptr inbounds %"class.pyston::BoxedClosure"* %0, i64 0, i32 1, !dbg !7
  %9 = load %"class.pyston::BoxedClosure"** %8, align 8, !dbg !7
  %10 = getelementptr %"class.pyston::BoxedClosure"* %9, i64 0, i32 3, i64 0, !dbg !7
  %11 = load %"class.pyston::Box"** %10, align 8, !dbg !7
  call void (i64, i32, i8*, i32, ...)* @llvm.experimental.patchpoint.void(i64 45780160, i32 95, i8* @assertDerefNameDefined, i32 2, %"class.pyston::Box"* %11, i8* inttoptr (i64 45231744 to i8*), i8* %scratch1.sub, %"struct.pyston::FrameInfo"* %frame_info, i64 45205664,   %"class.pyston::Box"* %7, %"class.pyston::BoxedClosure"* %0, %"class.pyston::Box"* %1), !dbg !7
```

This fixes the test, which is good. I was optimistic about it giving a slight performance boost for closures, but it turned out a lot crappier than I had hoped:

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.0%)
              pyston interp2.py                         :    6.5s baseline: 6.5 (+0.4%)
              pyston raytrace.py                        :    7.1s baseline: 6.8 (+4.5%)
              pyston nbody.py                           :    8.9s baseline: 8.8 (+1.4%)
              pyston fannkuch.py                        :    8.8s baseline: 8.6 (+2.3%)
              pyston chaos.py                           :   21.8s baseline: 20.9 (+4.2%)
              pyston spectral_norm.py                   :   28.0s baseline: 24.9 (+12.3%)
              pyston clos.py                            :    0.7s baseline: 0.7 (+8.8%)
              pyston fasta.py                           :    5.6s baseline: 5.9 (-5.4%)
              pyston pidigits.py                        :    5.4s baseline: 5.3 (+1.3%)
              pyston richards.py                        :    2.0s baseline: 2.0 (-2.5%)
              pyston deltablue.py                       :    2.1s baseline: 2.1 (+0.1%)
              pyston (geomean-6d4e)                     :    5.7s baseline: 5.6 (+2.4%)
```

I can't figure why so many of the benchmarks would be slower, notably `spectral_norm.py`, even though as far as I can tell they don't use closures at all. (If they don't use closures, this diff shouldn't have any effect, right?) Even `clos.py` (a contrived benchmark I wrote just now that uses a lot of closures) is slower. However, I think the slowness is due to the `assertDerefNameDefined` function call, so I'm hoping the speed will improve when I inline it. (The IC under the old method doesn't have a function call like that, just an inline guard.)

Anyway... thoughts?

### DIFF
--- a/src/analysis/scoping_analysis.h
+++ b/src/analysis/scoping_analysis.h
@@ -25,6 +25,11 @@ class AST_Module;
 class AST_Expression;
 class AST_Suite;
 
+struct DerefInfo {
+    size_t num_parents_from_passed_closure;
+    size_t offset;
+};
+
 class ScopeInfo {
 public:
     ScopeInfo() {}
@@ -75,39 +80,6 @@ public:
     };
     virtual VarScopeType getScopeTypeOfName(InternedString name) = 0;
 
-    // Returns true if the variable should be passed via a closure to this scope.
-    // Note that:
-    //      (a) This can be false even if there is an entry in the closure object
-    //          passed to the scope, if the variable is not actually used in this
-    //          scope or any child scopes. This can happen, because the variable
-    //          could be in the closure to be accessed by a different function, e.g.
-    //
-    //                  def f();
-    //                      a = 0
-    //                      b = 0
-    //                      def g():
-    //                           print a
-    //                      def h():
-    //                           print b
-    //                           # locals() should not contain `a` even though `h` is
-    //                           # passed a closure object with `a` in it
-    //                           print locals()
-    //
-    //      (b) This can be true even if it is not used in this scope, if it
-    //          is used in a child scope. For example:
-    //
-    //                  def f():
-    //                       a = 0
-    //                       def g():
-    //                           def h():
-    //                               print a
-    //                           print locals() # should contain `a`
-    //
-    // This is useful because it determines whether a variable from a closure
-    // into the locals() dictionary.
-
-    virtual bool isPassedToViaClosure(InternedString name) = 0;
-
     // Returns true if the scope may contain NAME variables.
     // In particular, it returns true for ClassDef scope, for any scope
     // with an `exec` statement or `import *` statement in it, or for any
@@ -115,6 +87,11 @@ public:
     virtual bool usesNameLookup() = 0;
 
     virtual bool areLocalsFromModule() = 0;
+
+    virtual DerefInfo getDerefInfo(InternedString name) = 0;
+    virtual const std::vector<std::pair<InternedString, DerefInfo>>& getAllDerefVarsAndInfo() = 0;
+    virtual size_t getClosureOffset(InternedString name) = 0;
+    virtual size_t getClosureSize() = 0;
 
     virtual InternedString mangleName(InternedString id) = 0;
     virtual InternedString internString(llvm::StringRef) = 0;
--- a/src/analysis/scoping_analysis.h
+++ b/src/analysis/scoping_analysis.h
@@ -25,6 +25,11 @@ class AST_Module;
 class AST_Expression;
 class AST_Suite;
 
+// Each closure has an array (fixed-size for that particular scope) of variables
+// and a parent pointer to a parent closure. To look up a variable from the passed-in
+// closure (i.e., DEREF), you just need to know (i) how many parents up to go and
+// (ii) what offset into the array to find the variable. This struct stores that
+// information. You can query the ScopeInfo with a name to get this info.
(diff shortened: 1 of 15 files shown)


toshok:
:+1: on the idea.  aside from the assert call, is there any increase in the analysis time?

maybe further down the line the analysis could also determine when it doesn't need to use `assertDerefNameDefined` all (if a given use is dominated by the def)?

kmod:
sgtm too.  spectral_norm is such a simple+short benchmark that I can't imagine this change affecting that... I tried removing the call to assertDerefNameDefined, and it got 5% slower, and then I put it back and it got 3% faster.  I tried this again on jitdev, which has usually been more consistent than my desktop, and it said that your change made spectral_norm 2% faster.

I think spectral_norm is an unreliable benchmark.  We've had issues with it in the past: it's so tight and exercises so little, that any tiny changes could influence what gets inlined and all of a sudden there's a huge performance change (we've seen +-50% jumps from unrelated changes).  I would definitely be in favor of removing it from the suite especially if we can replace it with something like the django / virtualenv benchmarks.

tjhance:
I inlined the definedness check, and it seemed to help. I also went through and cleaned it up a bit, so this should be ready for review.

> maybe further down the line the analysis could also determine when it doesn't need to use assertDerefNameDefined (if a given use is dominated by the def)?

This may actually be relatively easy compared to other definedness analysis; Python has a rule that you can't `del` a variable stored in a closure.

test/tests/static_closure_locations.py
```diff
@@ -0,0 +1,25 @@
+# should_error
```
kmod: Can you catch the exception that gets thrown and then print it?  #should_error is undesirable because then we can't detect a mistake in the test itself (ex a typo).

src/runtime/objmodel.cpp
```diff
@@ -1274,39 +1278,7 @@ Box* getattrInternalGeneric(Box* obj, const std::string& attr, GetattrRewriteArg
     }
 
     // TODO this should be a custom getattr
```
kmod: can probably remove this comment now?

src/analysis/scoping_analysis.h
```diff
@@ -75,11 +85,28 @@ class ScopeInfo {
...
+    virtual bool usesNameLookup() = 0;
+
+    virtual bool areLocalsFromModule() = 0;
+
+    // For a variable with DEREF lookup, return the DerefInfo used to lookup
+    // the variable in a passed closure.
+    virtual DerefInfo getDerefInfo(InternedString name) = 0;
+
+    // Gets the DerefInfo for each DEREF variable accessible in the scope.
+    // The returned vector is in SORTED ORDER by the `num_parents_from_passed_closure` field.
```
kmod: minor, but I would be explicit that it's ascending order.

kmod:
lgtm!

PR #416, opened 2015-03-27
Misc fixes for virtualenv / pip

undingen:
Hopefully one of the last sets of patches required.
- Update libpypa to fix another missing newline at end of file bug.
- Set distutils package install directory
- zipimporter: Ignore bytecode files inside archives (This let's them still load if the souce file can get found)

### DIFF
--- a/from_cpython/Lib/distutils/command/install.py
+++ b/from_cpython/Lib/distutils/command/install.py
@@ -41,8 +41,11 @@ else:
 
 INSTALL_SCHEMES = {
     'unix_prefix': {
-        'purelib': '$base/lib/python$py_version_short/site-packages',
-        'platlib': '$platbase/lib/python$py_version_short/site-packages',
+        # Pyston change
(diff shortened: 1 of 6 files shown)


from_cpython/Lib/distutils/command/install.py
```diff
@@ -41,8 +41,8 @@
 
 INSTALL_SCHEMES = {
     'unix_prefix': {
-        'purelib': '$base/lib/python$py_version_short/site-packages',
-        'platlib': '$platbase/lib/python$py_version_short/site-packages',
+        'purelib': '$base/site-packages',
+        'platlib': '$base/site-packages',
```
kmod: can we add a "Pyston change" comment here?

undingen:
- I added the ```Pyston changed``` comment
- Cleaned up yesterdays commit #408 
- Changed a error message slightly to output same string as my ubuntu cpython 2.7.8

kmod:
cool thanks :)

PR #417, opened 2015-03-27
add StackRoot<> and #defines for DECREF

toshok:
no usage of StackRoot<> yet, just getting the definition in for the coming work.

### DIFF
--- a/from_cpython/Include/object.h
+++ b/from_cpython/Include/object.h
@@ -826,7 +826,7 @@ PyAPI_FUNC(void) _Py_AddToAllObjects(PyObject *, int force) PYSTON_NOEXCEPT;
 
 
 #define Py_INCREF(op) ((void)(op))
-#define Py_DECREF(op) ((void)(op))
+#define Py_DECREF(op) asm volatile("" : : "X"(op))
 
 /* Safely decref `op` and set `op` to NULL, especially useful in tp_clear
  * and tp_dealloc implementatons.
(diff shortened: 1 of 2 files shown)


src/gc/roots.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace pyston {
+
+#define GC_KEEP_ALIVE(t) asm volatile("" : : "x"(t))
+
+template <class T> class StackRoot {
+public:
+    explicit StackRoot(T* t) : t(t) {}
+    StackRoot(const StackRoot& other) : t(other.t) {}
+    template <class... Args> StackRoot(Args&&... args) : t(new T(args...)) {}
```
undingen: I think this could use a ```t(new T(std::forward<Args>(args)...))``` but not sure if it makes a difference for the types we use it on.

from_cpython/Include/object.h
```diff
@@ -826,7 +826,7 @@ PyAPI_FUNC(void) _Py_AddToAllObjects(PyObject *, int force) PYSTON_NOEXCEPT;
 
 
 #define Py_INCREF(op) ((void)(op))
-#define Py_DECREF(op) ((void)(op))
+#define Py_DECREF(op) asm volatile("" : : "x"(op))
```
kmod: It looks like this should be a `X` (capital x) constraint; lower case `x` apparently means "SSE register" :P

kmod:
Cool :)  lgtm with marius's and my comments.

from_cpython/Include/object.h
toshok: hah, thanks for catching that.  The page detailing constraints that I'd looked at indeed had a capital X.  it just looked exactly like a lowercase x.

src/gc/roots.h
toshok: done.

PR #419, opened 2015-03-30
through syntax error when you to `del` a var stored in a closure

tjhance:


kmod:
Cool, looks like exec support makes testing this stuff way nicer :)

PR #421, opened 2015-03-30
Update Makefile

arbipher:
"@clang" is not installed globally according to the previous instructions. use $(CLANG_EXE) instead.
"cmake_check" is used in "make pyston_gcc", which is in one of the steps of "make check".
@clang will trigger an error interrupting "make check"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -859,7 +859,7 @@ pyston_profile: $(PROFILE_OBJS) $(LLVM_PROFILE_DEPS)
 	$(VERB) $(CXX) $(PROFILE_OBJS) $(LDFLAGS_PROFILE) -o $@
 
 cmake_check:
-	@clang --version >/dev/null || (echo "clang not available"; false)
+	@$(CLANG_EXE) --version >/dev/null || (echo "clang not available"; false)
 	@cmake --version >/dev/null || (echo "cmake not available"; false)
 	@ninja --version >/dev/null || (echo "ninja not available"; false)
 


kmod:
Ah, good point, though I think the way you fixed it might cause issues for people who only use the USE_CMAKE shim, since $(CLANG_EXE) doesn't get built or used in that mode (it just assumes that you installed clang globally).  How about, could you split `cmake_check` into a separate `cmake_check` (checks for cmake and ninja) and `clang_check` (checks for `clang`), and have the clang cmake recipes (the first two) call both, but the gcc one just calls cmake_check?

arbipher:
Spend some time reading the Makefile.
I agree with using $(CLANG_EXE) is error here.

I didn't use CMake last time (in fact I didn't install CMake before). It's strange to find the error in "cmake_check".

See line 866-893. There is a "if-else-statement" wrapper. If USE_CMAKE isn't 1 (USE_CMAKE is 0 by default), there is nothing to do with "cmake_check" when compiling pyston_dbg/pyston_gcc/pyston_release.

I don't know this bug happens. I run "make check" "make pyston_dbg" several times these days. It works fine.
I manually change USE_CMAKE to 1, it tell me "clang not available". It's not a bug for default(non-cmake) user.

I will try to fix it as what you say.

arbipher:
see #430 
I split cmake_check to cmake_check and clang_check
I read CMake section in INSTALLING, finding clang-3.5 is installed globally...

kmod:
I think this was superseded by #430, but feel free to reopen if I'm misunderstanding.

PR #424, opened 2015-03-31
Set encoding on streams

jvkersch:
This PR ports some code from CPython to Pyston to set an encoding on std{in, out, err}. @kmod This is in response to your suggestion on #246.

While this seems to work well (most of the code was copied almost verbatim from CPython), there is one thing that I don't understand yet: in `test/tests/unicode_test.py` there is a gc-related piece of code that essentially does
```python
 for i in xrange(100):
     print repr(BaseException().__unicode__())
     gc.collect()
     # do some allocations:
    for j in xrange(100):
        [None] * j
```
For some reason, this seems to reset the stream encodings and causes all subsequent attempts to print unicode characters to fail with a `UnicodeEncodeError`. In the PR, I've commented out the inner loop, that seems to be sufficient to make the tests pass. I haven't looked into what the problem could be.


kmod:
I think the GC aspect is a red herring here -- I think this issue is due to the loop causing us to OSR up to the llvm tier.  You can force the LLVM tier by passing the `-n` argument to pyston; the test is failing for me when I do `make run_unicode_test ARGS=-n` (or you can uncomment those lines, force the interpreter with the `-I` flag, and it works).  It looks like you changed the ast interpreter but not the IR generator, so I think this is just an issue of having to make the analogous change to src/codegen/irgen/irgenerator.cpp::doPrint().  It might be easier to put the logic into a new runtime function rather than trying to have the irgenerator emit the code that you used in the ast interpreter.

jvkersch:
Thanks, I completely overlooked that part of the compiler. I added a new runtime function to do this (it took me a while to understand how this works, hence the delay, but it was worth figuring this out). There are still some problems e.g. with file redirection, so this isn't good to go yet, but I have a good grasp on how to proceed.

jvkersch:
Hmmm, after a few false starts I got the unicode tests to pass, and the only thing standing between this PR and a successful Travis run is a failed statcheck in `26.py`. Unfortunately I don't really understand why this fails in the first place: the slowpath statistics get a little worse (with e.g. the most significant statistic, `slowpath_box_getattr`, going from 47761 to 51081) but I don't know if this is sufficient to explain why that test fails.  @kmod Would you mind taking a look at the `strOrUnicode` function that I've implemented to see if this could be done differently?

kmod:
If you do `make check_26`, you get: `Exception: ('noninit_count("slowpath_runtimecall") <= 5', 'slowpath_runtimecall', 6)`.  So we were saying that there should be at most 5 slowpath calls but we saw 6; usually low numbers like that just mean that we're trying to assert that the slowpaths were only hit a constant number of times and not per loop iteration, and when they get exceeded a small amount it's usually fine.  So in this case I think it's fine to just increase the check value from 5 to something larger (say, 10).  We could also increase the loop iteration count and loosen the check further.

jvkersch:
@kmod Thanks for the help! I bumped the slowpath count, merged master and resolved merge conflicts ... and all hell broke loose ;) I kept hitting an assert in `PyString_AsString`, and further inspection showed that `f->f_encoding` changes from being a BoxedString to a BoxedList when we go up to the llvm tier. I set a watchpoint on `(f->f_encoding)->cls` and I get the following traceback when it hits:
```
Hardware watchpoint 10: -location (f->f_encoding)->cls

Old value = (pyston::BoxedClass *) 0x1270009008
New value = (pyston::BoxedClass *) 0x0
gc_alloc (bytes=40, kind_id=pyston::gc::PYTHON) at src/gc/gc_alloc.h:95
95          return r;
(gdb) list
90      // if (VERBOSITY()) printf("Allocated %ld bytes at [%p, %p)\n", bytes, r, (char*)r + bytes);
91      #endif
92          // printf("Allocated %p\n", r);
93
94
95          return r;
96      }
97
98      extern "C" inline void* gc_realloc(void* ptr, size_t bytes) {
99          // Normal realloc() supports receiving a NULL pointer, but we need to know what the GCKind is:
(gdb) bt
#0  gc_alloc (bytes=40, kind_id=pyston::gc::PYTHON) at src/gc/gc_alloc.h:95
#1  0x000000000068732d in pyston::BoxedList::operator new (size=40) at src/runtime/types.h:485
#2  0x00000000008693af in pyston::listMul (self=0x12701dd6e8, rhs=0x127001c888) at src/runtime/list.cpp:407
#3  0x00007ffff7f52964 in <module>_e1_osr8_0 () at test/tests/unicode_test.py:55
#4  0x000000000067f3db in pyston::(anonymous namespace)::ASTInterpreter::visit_jump (this=0x7fffffffd820, node=0x2c29eb0) at src/codegen/ast_interpreter.cpp:526
#5  0x000000000067d05c in pyston::(anonymous namespace)::ASTInterpreter::visit_stmt (this=0x7fffffffd820, node=0x2c29eb0) at src/codegen/ast_interpreter.cpp:691
#6  0x00000000006796d6 in pyston::(anonymous namespace)::ASTInterpreter::execute (interpreter=..., start_block=0x2c09ee0, start_at=0x2c1b510) at src/codegen/ast_interpreter.cpp:316
#7  0x0000000000679ba7 in pyston::astInterpretFunction (cf=0x2c1b0f0, nargs=0, closure=0x0, generator=0x0, globals=0x0, arg1=0x0, arg2=0x0, arg3=0x0, args=0x0) at src/codegen/ast_interpreter.cpp:1211
#8  0x0000000000767ea6 in pyston::compileAndRunModule (m=0x2c096d0, bm=0x1270180d28) at src/codegen/irgen/hooks.cpp:326
#9  0x000000000090e114 in pyston::main (argc=2, argv=0x7fffffffe598) at src/jit.cpp:222
#10 0x000000000090d432 in main (argc=2, argv=0x7fffffffe598) at src/jit.cpp:307
```
The listMul is probably due to the allocations in `unicode_test.py`, but I don't see how this would affect `f->f_encoding`. 

I'll try to see if this is due to a change I made (very likely), and if I can't rule that out I'll dive a little deeper into the generated IR. If all else fails, I'll do a `git bisect`. In the meantime, thanks for your patience!

kmod:
Ah ok, that kind of thing (object mysteriously changes to something else) is usually due to a gc issue, where we forget to scan an object and let it get collected, and only notice once something else was allocated in that space.  In this case it looks like we're not scanning any of the file fields -- I double checked when you sent the PR, and we have a gc visitor set up (BoxedFile::gcHandler), but now that I look again we never attach it to the file class!  We need to pass that as the gc_visitor argument when we create file_cls.

jvkersch:
Ahh, that makes a lot of sense! I can create a PR to attach the visitor, but probably not before the end of the week.

jvkersch:
Hurray, that seems to have solved the remaining problems. @kmod Thanks for your continuing input, I think this is finally ready for review.

kmod:
awesome, thanks for the contribution!

PR #422, opened 2015-03-30
a few changes which help out GC performance

toshok:
A few changes (mostly dealing with scanning) that help GC perf.

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (-0.4%)
              pyston django_migrate.py                  :    5.9s baseline: 10.7 (-45.1%)
              pyston interp2.py                         :    6.0s baseline: 5.9 (+1.5%)
              pyston raytrace.py                        :    6.7s baseline: 6.8 (-1.1%)
              pyston nbody.py                           :    8.9s baseline: 9.3 (-4.0%)
              pyston fannkuch.py                        :    8.1s baseline: 8.0 (+0.8%)
              pyston chaos.py                           :   21.3s baseline: 20.5 (+4.2%)
              pyston fasta.py                           :    6.0s baseline: 5.6 (+8.2%)
              pyston pidigits.py                        :    5.4s baseline: 5.5 (-0.6%)
              pyston richards.py                        :    2.0s baseline: 2.0 (-0.9%)
              pyston deltablue.py                       :    2.1s baseline: 2.1 (+0.8%)
              pyston (geomean-8999)                     :    5.9s baseline: 6.2 (-5.0%)
```

PR #425, opened 2015-03-31
free generator stacks both in the destructor and when a generator exits.

toshok:
also, don't register MAX_STACK_SIZE bytes with the GC, register INITIAL_STACK_SIZE.

----

splitting this out from #422 to get it in sooner, since it's more likely to have a large impact with the big-app work than the other fixes.

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (-0.3%)
              pyston django_migrate.py                  :    5.9s baseline: 10.7 (-45.1%)
              pyston interp2.py                         :    6.0s baseline: 5.9 (+1.6%)
              pyston raytrace.py                        :    6.6s baseline: 6.8 (-1.7%)
              pyston nbody.py                           :    8.9s baseline: 9.3 (-4.0%)
              pyston fannkuch.py                        :    8.1s baseline: 8.0 (+0.7%)
              pyston chaos.py                           :   21.5s baseline: 20.5 (+5.2%)
              pyston fasta.py                           :    6.0s baseline: 5.6 (+8.3%)
              pyston pidigits.py                        :    5.4s baseline: 5.5 (-0.4%)
              pyston richards.py                        :    2.0s baseline: 2.0 (-1.0%)
              pyston deltablue.py                       :    2.1s baseline: 2.1 (+0.8%)
              pyston (geomean-8999)                     :    5.9s baseline: 6.2 (-5.0%)
```

### DIFF
--- a/src/runtime/generator.cpp
+++ b/src/runtime/generator.cpp
@@ -71,13 +71,6 @@ static void freeGeneratorStack(BoxedGenerator* g) {
         available_addrs.pop_front();
         int r = munmap((void*)(addr - MAX_STACK_SIZE), MAX_STACK_SIZE);
         assert(r == 0);
-// also unmap the redzone
-#if STACK_GROWS_DOWN
-        r = munmap((void*)addr, STACK_REDZONE_SIZE);
-        assert(r == 0);
-#else
-#error "implement me"
-#endif
     }
 
     g->stack_begin = NULL;
@@ -111,7 +104,6 @@ void generatorEntry(BoxedGenerator* g) {
     // we returned from the body of the generator. next/send/throw will notify the caller
     g->entryExited = true;
     threading::popGenerator();
-    freeGeneratorStack(g);
     swapContext(&g->context, g->returnContext, 0);
 }
 
@@ -127,8 +119,10 @@ Box* generatorSend(Box* s, Box* v) {
         raiseExcHelper(ValueError, "generator already executing");
 
     // check if the generator already exited
-    if (self->entryExited)
+    if (self->entryExited) {
+        freeGeneratorStack(self);
         raiseExcHelper(StopIteration, "");
+    }
 
     self->returnValue = v;
     self->running = true;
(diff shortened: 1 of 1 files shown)


src/runtime/generator.cpp
```diff
@@ -60,6 +60,29 @@ class RegisterHelper {
...
+    // Limit the number of generator stacks we keep around:
+    if (available_addrs.size() > 5) {
+        uint64_t addr = available_addrs.front();
+        available_addrs.pop_front();
+        int r = munmap((void*)(addr - MAX_STACK_SIZE), MAX_STACK_SIZE);
+        assert(r == 0);
+// also unmap the redzone
+#if STACK_GROWS_DOWN
+        r = munmap((void*)addr, STACK_REDZONE_SIZE);
+        assert(r == 0);
```
kmod: hmm you should double-check my math, but I thought this shouldn't be necessary since the redzone starts at `addr - MAX_STACK_SIZE`?

src/runtime/generator.cpp
```diff
@@ -88,6 +111,7 @@ void generatorEntry(BoxedGenerator* g) {
     // we returned from the body of the generator. next/send/throw will notify the caller
     g->entryExited = true;
     threading::popGenerator();
+    freeGeneratorStack(g);
```
kmod: I think this works out as long as we always keep at least 1 generator stack around, but it feels risky to call freeGeneratorStack while we're on the stack that we're freeing.  Do you think it works to put this on the other side somewhere? (somewhere in generatorSend I guess)

kmod:
lgtm, a pretty nice improvement :)

src/runtime/generator.cpp
toshok: yeah, I don't know what I was thinking there.  My initial reading of the code was that the redzone was outside of the MAX_STACK_SIZE range, and even if that were true (which it isn't), the munmap's args are wrong.

src/runtime/generator.cpp
toshok: Honestly I was kinda surprised this worked, but I suppose the register allocator must be stashing `g` in a callee-save register before the call to `freeGeneratorStack`.

It'll work to put it on the other side (and will be safer).  Given my limited understanding of python's iterator protocol, it seems like the only downside would be that the stack could stick around a little longer.

kmod: I think it's because we end up saving the stack in available_addrs, and even if we hit the max number of saved stacks (5) we free a different one (the one that had been in there the longest) to try to encourage cache reuse.  I tried changing it to saving 0 stacks and was able to get a segfault.

PR #426, opened 2015-04-01
sys._getframe + frame objects

toshok:
implement sys._getframe + frame objects, and add co_name/co_filename properties to code object

also rework a bit of the logging module to use only sys._getframe, not f_back.

### DIFF
--- a/from_cpython/Lib/logging/__init__.py
+++ b/from_cpython/Lib/logging/__init__.py
@@ -71,15 +71,9 @@ else:
     _srcfile = __file__
 _srcfile = os.path.normcase(_srcfile)
 
-# next bit filched from 1.5.2's inspect.py
-def currentframe():
-    """Return the frame object for the caller's stack frame."""
-    try:
-        raise Exception
-    except:
-        return sys.exc_info()[2].tb_frame.f_back
-
-if hasattr(sys, '_getframe'): currentframe = lambda: sys._getframe(3)
+# pyston changes: we don't support tb_frame or f_back, so always use sys._getframe
+currentframe = lambda: sys._getframe(4)
+start_getframe = 4
 # done filching
 
 # _srcfile is only used in conjunction with sys._getframe().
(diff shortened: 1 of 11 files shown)


kmod:
Man, I can't see anyone relying on that behavior... is there some way we can at least prevent people from trying to do this?  ex somehow detect if the Frame object corresponds to a frame that has exited / been returned to, and then throw an exception/error out if any of the fields get accessed.

If we can't do that (so we would really have to make `_getframe` support these updating semantics), maybe we could add a new function that explicitly makes no promises about how that will behave, and then we can migrate code over?  Right now we have no way of replicating the behavior of something like `sys._getframe(4).f_locals`, but I think it could be pretty reasonable to change all the callers if it was just a matter of changing it to `sys._getframeThatIPromiseWontExecute(4).f_locals`. 

On an unrelated note, I think we can get rid of the "Pyston change" in warnings.py now as well.

tjhance:
(I'm not actually worried about the update semantics when the frame execute mores - I don't think this is difficult to implement - I'm just worried about what happens when the frame terminates.)

I'm usually the last to suggest breaking compatibility, but we can probably get away with having the frame objects just break with the frame ends, but work fine up until that point. My biggest worry would be if someone tries to make a "helper function" to get a frame

```python
def get_frame():
    return sys._getframe(0)
```

or something like that

kmod:
Ok just to close the loop on this -- I made some changes on top of this PR and merged it.  It ended up being a bit more involved than I thought it would since it triggered some libunwind issues.

But now we have real f_locals and f_lineno, which you can access even if the referenced frame executes further, but not if it exits.  I think the has-the-frame-exited logic is robust, but it's hard to be sure, since it uses stack introspection at the point that you try to use the frame object, rather than doing invalidation when you exit the frame.  I tried doing the latter by overwriting the return IP with a trampoline, which works decently well except that it breaks exception handling so I gave up on that.

PR #427, opened 2015-04-01
use the same hash function for both unicode and string objects

toshok:
keeps from allocating a BoxedString every time we need a unicode object's hash().

saves ~1% on django_migrate.py.


kmod:
lgtm.  I just noticed though that we only test `hash(u'') == hash('')`, which is not a very robust check... could you add/update it to hash a non-empty string?

toshok:
done, will merge once travis clears.

PR #428, opened 2015-04-02
add some CPython tests

rntz:


kmod:
The github web interface chokes on such a large diff... in the future I guess separating the mass-import into a different PR would make that easier (I don't think we have to do that for this one).  If you feel like it, splitting up the import commit so that the imports and the tester changes are separate would be nice.

As for the lookup methods:
- `Box::getattr` is the equivalent of `PyDict_GetItem(obj->dict, attr)`, ie it just looks up the attribute value in that object's attribute storage.  It doesn't look at other objects or do any descriptor logic.
- `typeLookup` is the equivalent of `_PyType_Lookup`, which calls `Box::getattr` on each item in the MRO in order.  It doesn't do any descriptor logic either
- `getattrInternalGeneric` is the equivalent of PyObject_GenericGetAttr, which is the default lookup rules (check for a data descriptor, check for instance attribute, check for non-data descriptor).  It does not check for `__getattr__` or `__getattribute__`.
- `getattrInternal` is the equivalent of PyObject_GetAttr, which checks for custom getattr methods and then calls them.

Couple other thoughts:
- Could you make the new tester options not specific to/named for CPython?  I'm thinking of adding some integration tests which would want pretty much the exact same behavior.
- What do you think about adding these in test/cpython instead of test/tests/cpython?
- are you having issues with sys_test/io_test/longargs_stackusage?  and also `from _bisect import *`?  we might want to make that into a test

kmod:
Sorry, just noticed that there was more non-import code in the big commit... definitely please split that :P

Can you make sure our edits to things in from_cpython/Lib/ have the string "Pyston" in there somewhere?  (Usually we mark things as a "Pyston change".)  We have this info via the source control history, but it's nice to also have it be easily-greppable.

otherwise lgtm!

rntz:
Should I annotate the "# expected: fail" lines at the top of various test files in from_cpython/Lib/ as "Pyston changes"? That seems like it would produce a lot of noise.

rntz:
Oh, I put "skip-if: True" into longargs_stackusage and tuple_depth because they were previously listed in TO_SKIP in tester.py, and I thought it was cleaner to just have "#skip-if" annotations in the files rather than having two ways to skip files.

kmod:
Hmm maybe the ideal would be to not have our annotations in the Lib/ directory, but for now I think it's fine and I agree that we can skip the Pyston change notes (I guess the Lib/test directory isn't really the same as the rest of the Lib directory).

rntz:
Hm, the failing Travis CI is because the return_selfreferential test will fail only if asserts are turned on, which they aren't on Travis, and I can't detect this using the `IMAGE` variable because in the CMake build the executable is always called `pyston`.

kmod:
I would recommend removing the assert :)

rntz:
@kmod Okay, this should be mergeable now.

PR #429, opened 2015-04-02
basic implementation of `exec code in dict`

tjhance:
This lets you do `exec code in dict`, sort of

It doesn't pass in the globals, but it should still work a lot of the time (as long as there's no global variable reference from within a function in the exec, and if there is nothing explicitly declared `global`)

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -888,12 +888,12 @@ Value ASTInterpreter::visit_print(AST_Print* node) {
 }
 
 Value ASTInterpreter::visit_exec(AST_Exec* node) {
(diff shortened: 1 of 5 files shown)


tjhance:
I'm not sure why this is failing--it runs fine on my machine, and I don't know how to get better information about the test failure from the travis-ci machine

tjhance:
Never mind, this was merely a case of me pushing the wrong thing. Should work now

kmod:
Cool!  Let's hold off on merging this into master until it's working fully, but I'll personally use this in my metaserver branch :)

As we get to running larger and larger codebases, I only get more wary of introducing "allow the wrong thing to happen" behavior.  It's been super annoying to track down bugs where the symptoms are extremely far away from the cause (ex the django session management issue).

PR #430, opened 2015-04-02
split cmake_check to cmake_check and clang_check

arbipher:


Makefile
```diff
@@ -894,7 +898,7 @@ endif
 CMAKE_DIR_GCC := $(HOME)/pyston-build-gcc
 CMAKE_SETUP_GCC := $(CMAKE_DIR_GCC)/build.ninja
 $(CMAKE_SETUP_GCC):
-	@$(MAKE) cmake_check
+	@$(MAKE) clang_check
```
kmod: should this be `cmake_check`?

arbipher: You are right. Sorry for the silly error.

Not quite familiar with github workflow.
Should I add a new commit after your view or change the commit to the correct one? (I choose the latter, but it's troublesome than the former)

kmod:
cool, thanks!

I usually prefer amending the PR rather than adding a new "oops" commit -- `git commit -a --amend` can be a huge time saver with that.

arbipher:
"Oops" Got it.

PR #432, opened 2015-04-03
use a binary search to find/insert CompiledFunction*'s into the registry

toshok:
with this we drop from 50ms to 20ms for django requests.  I expect it'll help other things as well, with bigger impact the more functions we jit.

### DIFF
--- a/src/codegen/unwinding.cpp
+++ b/src/codegen/unwinding.cpp
@@ -88,19 +88,55 @@ void parseEhFrame(uint64_t start_addr, uint64_t size, uint64_t* out_data, uint64
 
 class CFRegistry {
 private:
-    // TODO use a binary search tree
     std::vector<CompiledFunction*> cfs;
 
-public:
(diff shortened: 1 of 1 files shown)


kmod:
Oh wow didn't expect that to be such a performance hit at this point (I'm curious why we call this so many times).  In the future we might be able to use std::set and lower_bound()?

toshok:
It actually might have been due to the exception I fixed in str.decode().
Will do a perf run without this fix and with the str.decode fix and see if
it's as hot (it was like 40% of runtime iirc with the bulk of it coming
from frame iteration)

On Apr 3, 2015, at 4:47 PM, Kevin Modzelewski <notifications@github.com>
wrote:

Oh wow didn't expect that to be such a performance hit at this point (I'm
curious why we call this so many times). In the future we might be able to
use std::set and lower_bound()?

—
Reply to this email directly or view it on GitHub
<https://github.com/dropbox/pyston/pull/432#issuecomment-89458684>.

PR #434, opened 2015-04-03
strip U out of the mode passed to fileNew before calling fopen

toshok:


kmod:
Can we just copy over CPython's `_PyFile_SanitizeMode()`?  It looks your change does right thing, but it's easier to see + trust it when it's directly CPython's code.

toshok:
turns out we already had it :)

kmod:
Can we use a unique_ptr to make sure that the `file_mode` allocation gets freed?  Otherwise lgtm.

PR #435, opened 2015-04-03
support 'linux2', 'darwin', and 'unknown' for sys.platform

toshok:
implement Py_GetPlatform along the way.  CPython's implementation of this function just returns PLATFORM (which gets #defined at build time from the configure script.)

### DIFF
--- a/src/runtime/builtin_modules/sys.cpp
+++ b/src/runtime/builtin_modules/sys.cpp
@@ -231,6 +231,20 @@ static bool isLittleEndian() {
     return s[0] != 0;
 }
 
+extern "C" const char* Py_GetPlatform() noexcept {
+// cpython does this check in their configure script
+#if defined(__linux__)
+    return "linux2";
+#elif defined(__APPLE__) && defined(__MACH__)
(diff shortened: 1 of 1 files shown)


src/runtime/builtin_modules/sys.cpp
```diff
@@ -260,7 +274,7 @@ void setupSys() {
     sys_module->giveAttr("py3kwarning", False);
     sys_module->giveAttr("byteorder", boxStrConstant(isLittleEndian() ? "little" : "big"));
 
-    sys_module->giveAttr("platform", boxStrConstant("unknown")); // seems like a reasonable, if poor, default
+    sys_module->giveAttr("platform", boxStrConstant(Py_GetPlatform())); // seems like a reasonable, if poor, default
```
kmod: the comment is obsolete now?

kmod:
lgtm!

src/runtime/builtin_modules/sys.cpp
toshok: ah, yeah oops

PR #438, opened 2015-04-07
Django fixes

toshok:


CMakeLists.txt
```diff
@@ -155,6 +155,7 @@ add_definitions(-DLLVMREV=${LLVMREV})
 
 include_directories(${LLVM_INCLUDE_DIRS})
 include_directories(from_cpython/Include)
+include_directories(from_cpython/Modules)
```
kmod: Hmm what do you think about doing `#include "../Modules/socketmodule.h"` instead of adding Modules/ to the search path?  I know that it's ugly but CPython does something similar as well, and it might be risky to add Modules/ to the search path if it wasn't intended to be on it.

from_cpython/Include/Python.h
```diff
@@ -76,6 +76,7 @@
 #include "descrobject.h"
 #include "warnings.h"
 #include "weakrefobject.h"
+#include "socketmodule.h"
```
kmod: This should probably get included in the files that need it rather than in Python.h, since it's not part of CPython's Python.h.

from_cpython/Modules/socketmodule.c
```diff
@@ -3119,7 +3119,8 @@ sock_dealloc(PySocketSockObject *s)
 {
     if (s->sock_fd != -1)
         (void) SOCKETCLOSE(s->sock_fd);
-    Py_TYPE(s)->tp_free((PyObject *)s);
+    // Pyston change: the gc takes care of this
+    //Py_TYPE(s)->tp_free((PyObject *)s);
```
kmod: We should be able to support this kind of code in general, right?  Does the issue here come from the changes that we make to the socket class?

src/runtime/builtin_modules/thread.cpp
```diff
@@ -171,16 +171,19 @@ class BoxedThreadLocal : public Box {
...
+
+    static Box* getattr(Box* obj, char* name) {
+        Box* rtn = BoxedThreadLocal::getattrOrNull(obj, name);
+        if (!rtn)
             raiseExcHelper(AttributeError, "'%.50s' object has no attribute '%.400s'", obj->cls->tp_name, name);
-        }
+        return rtn;
     }
 
     static Box* hash(Box* obj) { return boxInt(PyThread_get_thread_ident()); }
```
kmod: unrelated to this PR, but I just saw this -- should this object really have a different hash on different threads?  looks like CPython doesn't set a custom hash function.

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -407,11 +407,15 @@ Box* getattrFunc(Box* obj, Box* _str, Box* default_value) {
     BoxedString* str = static_cast<BoxedString*>(_str);
 
     Box* rtn = NULL;
-    try {
-        rtn = getattr(obj, str->s.data());
-    } catch (ExcInfo e) {
-        if (!e.matches(AttributeError))
-            throw e;
+    if (obj->cls == thread_local_cls) {
+        rtn = getThreadLocal(obj, const_cast<char*>(str->s.data()));
```
kmod: I don't mind having this in here for now, but let's try to think about ways to not have to special case this...

I think one option is to change our getattr* functions to use CAPI-style exceptions.  We kind of do this by returning NULL in some (but not all) cases that the attribute doesn't exist, so it wouldn't be a huge break from what we're currently doing.

Anyway, we don't have to do that now, but it'd be nice to have a note here saying why or why not thread.local deserves special treatment here.

from_cpython/Include/Python.h
toshok: oh whoops, that was actually already added to the file that needed it (`types.cpp`).  will remove.

src/runtime/builtin_modules/thread.cpp
toshok: oh yeah, that's definitely wrong

toshok:
okay, with these changes I was able to add/remove users and groups, edit them, and generally navigate all pages having to do with the admin interface.

I was also able to flesh out the tutorial from here: http://www.creativebloq.com/netmag/get-started-django-7132932

and add the blog models/templates/etc to my local testsite, and actually make a post:

screen shot 2015-04-14 at 6 06 11 pm

I still need to address the code duplication in str.cpp (between the tp_as_sequence functions and the ones we add to str_cls), and I'd like to figure out a way to use the str_iterator_cls instead of a seq iterator for strings.

Makefile
```diff
@@ -225,7 +225,7 @@ endif
 CLANGFLAGS := $(CXXFLAGS_DBG) $(CLANG_EXTRA_FLAGS)
 CLANGFLAGS_RELEASE := $(CXXFLAGS_RELEASE) $(CLANG_EXTRA_FLAGS)
 
-EXT_CFLAGS := $(COMMON_CFLAGS) -fPIC -Wimplicit -O2 -Ifrom_cpython/Include
+EXT_CFLAGS := $(COMMON_CFLAGS) -fPIC -Wimplicit -O2 -Ifrom_cpython/Include -Ifrom_cpython/Modules
```
kmod: Do we still need this?

src/runtime/list.cpp
```diff
@@ -664,8 +664,7 @@ extern "C" void listIteratorGCHandler(GCVisitor* v, Box* b) {
     v->visit(it->l);
 }
 
-Box* listNew(BoxedClass* cls, Box* container) {
-    assert(cls->cls == type_cls);
+Box* listNew(BoxedClass* cls) {
```
kmod: I think we should still have the `args` argument to this even if it doesn't get used.

src/runtime/str.cpp
```diff
@@ -23,6 +23,10 @@
 
 #include "Python.h"
 
+#include "../../from_cpython/Objects/stringlib/stringdefs.h"
+#include "../../from_cpython/Objects/stringlib/fastsearch.h"
+#include "../../from_cpython/Objects/stringlib/find.h"
```
kmod: could these be "../Objects/stringlib/find.h"?

src/runtime/super.cpp
```diff
@@ -128,10 +129,10 @@ Box* superGetattribute(Box* _s, Box* _attr) {
...
     }
 
-    Box* r = typeLookup(s->cls, std::string(attr->s), NULL);
-    // TODO implement this
-    RELEASE_ASSERT(r, "should call the equivalent of objectGetattr here");
-    return processDescriptor(r, s, s->cls);
+    Box* r = typeLookup(s->cls, attr_s, NULL);
+    if (r)
+        return processDescriptor(r, s, s->cls);
+    raiseExcHelper(AttributeError, "'super' object has has no attribute '%s'", attr->data());
```
kmod: Hmm based on the original comment I think we need to call something like getattrInternalGeneric in case this is an instance attribute.

src/runtime/types.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include "llvm/Support/raw_ostream.h"
 
+#include "../../from_cpython/Modules/socketmodule.h"
```
kmod: ../Modules/socketmodule.h?

src/runtime/super.cpp
toshok: meaning getattrInternalGeneric instead of the `typeLookup` + `processDescriptor`?

kmod: Oh, sorry yeah that's what CPython does.  I think the code here is the same as getattrInternalGeneric when there aren't any instance attributes, but I'm not sure if either of us are very confident in that.

from_cpython/Modules/socketmodule.c
toshok: yeah I think we should be able to deal with this.  the problem was that socket's had tp_free == PyObject_Del, which was failing for gc allocated (not untracked) allocations.  maybe we can just have it return without doing anything in the case of a gc-tracked allocation.

toshok:
rebased this off upstream master after addressing the rest of the review comments.  need to squash this down before any merging happens.

currently with this branch the best I'm seeing for a given 100 requests (once everything is warmed up/osr'ed) of `http://localhost:8000/admin/auth/group/1` (after I created a group):

```
Percentage of the requests served within a certain time (ms)
  50%      6
...
 100%      6 (longest request)
```

if I disable tracebacks (by setting `ENABLE_FRAME_INTROSPECTION` to `0`), we drop down to:

```
Percentage of the requests served within a certain time (ms)
  50%      3
...
 100%      3 (longest request)
```

cpython is at 1ms for this page.

kmod:
Hey Chris sorry, I lost track of the state of this PR... is it obsoleted by the other stuff you've gotten in lately?  Are there things that we should try to separate out so we can get them merged?

toshok:
Yeah, let me close this.  It has some changes that are worth bringing in,
but I've been creating smaller PR's (like #578) for those this.

On Wed, Jun 3, 2015 at 12:20 AM, Kevin Modzelewski <notifications@github.com

PR #441, opened 2015-04-12
[WIP] CMake default / Ubuntu 14.10+ compatibility

dagar:
In order to support Ubuntu >= 14.10 and other distros we need a configure script which means moving to CMake.

@kmod Are there any remaining issues or missing features that would prevent moving to CMake at this point?

- [ ] update README.md
- [ ] update docs/INSTALLING.md
- [ ] make check use cmake check-pyston
- [ ] ~~include config.h in src/runtime/types.h (find a new home for sizeof asserts)~~ do later
- [x] cmake stop globbing source

Should the instructions be for the Makefile (cmake shim) or using cmake directly? For some of the testing helpers the shim is still necessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ ETAGS := ctags-exuberant -e
 # to disk space concerns.
 FORCE_TRUNK_BINARIES := 0
 
-USE_CMAKE := 0
+USE_CMAKE := 1
 NINJA := ninja
 
 # Put any overrides in here:
--- a/docs/INSTALLING.md
+++ b/docs/INSTALLING.md
@@ -35,112 +35,6 @@ make check
 make install
 ```
 
-### ccache
-
-ccache is a build tool that can help speed up redundant compilations.  It's not strictly necessary but it's useful enough to be enabled by default; you can disable it by adding `USE_CCACHE := 0` to your Makefile.local.  To get it, run:
-```
-sudo apt-get install ccache
-```
-
-### LLVM dependencies
-```
-sudo apt-get install libncurses5-dev zlib1g-dev liblzma-dev
-```
-
-### LLVM + clang
-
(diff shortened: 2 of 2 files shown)


kmod:
I think we should definitely go with the Makefile->cmake shim; I use those tester helpers all the time and switch between debug and release builds, and I have no idea what the right cmake configuration args are :P  I think a Makefile is really helpful as a kind of specialized and powerful shell and works well as a simple development environment.

There are some straggling things I feel could be better (I sometimes find myself running `touch CMakeLists.txt` to get things to work), but I've been using the cmake shim exclusively for a while so I'm all for making it the default.  @toshok @undingen @tjhance what do you guys think?

dagar:
I'll update all the documentation to refer only to the Makefile (with cmake shim) and keep instructions for building GCC or specific Ubuntu versions.

Initially we can just set USE_CMAKE=1 and the actually using the configure script will be the point of no return.

You shouldn't need to force cmake to rerun so please let me know when that happens. In src/ cmake currently just globs *.cpp which means new files aren't picked up automatically. I'll fix that.

kmod:
Let's go ahead and flip this :)  For the install instructions, how about we just swap the ordering of the sections from makefile/extra/"Experimental cmake" to cmake/extra/"deprecated makefile"?  We might want to keep the old install instructions in there until we feel comfortable getting rid of the makefile logic that they correspond to.

kmod:
Hope you don't mind, but I just flipped the switch and made `USE_CMAKE=1` our default!  Thanks again for building this for us :)

dagar:
Thanks @kmod.
One minor thing I wanted to change was to move the build directories to something like build/Debug, build/Release within the source dir. I can imagine some people wouldn't like us polluting their home directory.
It might also make sense to change the default from debug to release.

kmod:
Those sound reasonable, though I've also heard people get upset about polluting the source directory, so I guess we can't win :P  But yeah grouping them together wherever they would live would definitely be nice.

My feeling is we shouldn't change the default to release until our stability is better; I'd rather people see assertion failures than segfaults.  Though maybe we could have a bare `make` build both?

PR #443, opened 2015-04-13
WIP: use custom traceback code to speedup exception raising

undingen:
I unsuccessfully experimented with different libunwind flags in order to speedup exception handling, I then thought I could try to generate the traceback using a custom stack walking routine because we compile with enabled frame pointers.
This is my test code, I put it up here because I would like to know what's your opinion about this hack (if we agree to include it I will clean it up and remove the code duplication).

It works be using the ```__builtin_frame_address()``` and walking the returned list, so this is compiler and arch specific. It works for JITed frames because the ```!current_stmt``` will always end up inside the constant array of the stackmap and we can access it without keeping track of the registers in the stack frame.

For this artificial benchmark:
```python
def foo(x):
     r = 0
     for i in xrange(x):
         try:
             raise NotImplemented
         except:
             r = r + 1
     return r
 print foo(100000)
```

before:
```
us_gettraceback: 2379909
3,39 seconds time elapsed
```
with patch:
```
us_gettraceback: 274704
1,12 seconds time elapsed
```

### DIFF
--- a/src/codegen/unwinding.cpp
+++ b/src/codegen/unwinding.cpp
@@ -387,7 +387,7 @@ public:
         return rtn;
     }
 
-    unw_word_t getFunctionEnd(unw_word_t ip) {
+    static unw_word_t getFunctionEnd(unw_word_t ip) {
         unw_proc_info_t pip;
         int ret = unw_get_proc_info_by_ip(unw_local_addr_space, ip, &pip, NULL);
         RELEASE_ASSERT(ret == 0 && pip.end_ip, "");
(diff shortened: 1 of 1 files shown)


undingen:
Looks like the ```analysis_unittest```  is failing because gtest is compiled with ```-fomit-frame-pointers```...

rntz:
Hm, I'll take a look at this, it may be relevant to the custom unwinder I'm working on.

kmod:
So as I was saying on gitter I'm not sure how much we need to improve getTraceback(), since as Michael said the tracebacks are supposed to be generated as we cross a Python frame.  This means that for a common case of exceptions that get thrown+caught completely within the runtime, we don't have to generate any traceback entries at all, and even otherwise we can probably just fold the work into the unwinding that we have to do for the exception itself.

But I think there are definitely other times that this PR could be useful; it seems like the benefit is that if we know we don't need the registers then we can unwind way faster, and I'm not sure if we could/should teach libunwind that.  But it seems like a bunch of our stack-inspection features doesn't actually need the registers, either because the values are constant (getCurrentStatement, getCF) or because they live on the stack (getFrameExcInfo).  Might be interesting to have a `unwindPythonFrames(fast_and_skip_registers=true)` version?

@dagar it looks like we're getting gtest from the llvm tree; is there a way we could pass `--fno-omit-frame-pointer` to just that part of the build?

dagar:
No I don't think we can pass options to only gtest within llvm. We can either hack it (llvm_patches) or build our own gtest. Somewhat strangely there doesn't seem to be an official gtest git mirror.

kmod:
+1 on hacking it via a llvm_patch

dagar:
Just throw add_definitions("--fno-omit-frame-pointer") in LLVM utils/unittest/CMakeLists.txt

PR #445, opened 2015-04-15
fix recursive searching problem in Makefile

rntz:


kmod:
ah, much better, thanks :)

PR #446, opened 2015-04-16
Implement misc stuff

undingen:
- Add buffering argument support to ```open()```  aka 3-arg ```open()```
- Marshal module support
     - but currently does not support longs and code objects (will abort())
- Support for the extended slicing syntax (multiple slices inside a tuple)
- add ```apply()```
- ```PyDict_Values()```, ```PyDict_Items()``` (did not test them...)
- ```PyIter_Next()``` did not check ```__hasnext__``` and the set iterator depends on this behavior. While there I thought it's best to let it use our runtime ICs.
- add ```Py_intptr_t``` and ```Py_uintptr_t```
- Moved the ```tupleobject.h``` include inside ```Python.h``` to the position cpython uses.


kmod:
awesome :)

PR #447, opened 2015-04-16
Attach BoxedFile::gcHandler.

jvkersch:
Fixes the issue with `file->f_encoding` being collected in #424.


kmod:
Travis-CI (and probably Travis our team member as well) doesn't seem to like the formatting (just run `make format`).

jvkersch:
That'll teach me not to run tests, thinking that the PR is too short to bother. Sorry Travises!

Anyway, I amended my earlier commit, and tests pass on my local machine now. Hopefully that does it. Thanks for pointing this out.

PR #448, opened 2015-04-17
Set correct site-package directory, fix 'str * -1', add sys.dont_write_bytecode

undingen:
Small fixes for getting pyxl to work. Next step is teaching pypa about "# coding".


kmod:
cool, thanks for tracking it down :)

Are there ways we can bring our directory layout more in line with what CPython or PyPy do?  It's not that urgent but it might be worth starting to think about.  I don't have a good sense of whether the differences are gratuitous or not.

undingen:
I agree that the directory layout is not yet correct everywhere, but this change is correct in the sense that it replicates the behavior of pypy. 
When one creates a virtual env the installed packaged will get installed directly into  ```site-packages/```, which I thought makes more sense for us than putting them into ```from_cpython/Lib/site-packages/```.

PR #449, opened 2015-04-19
Investigate travis-ci failures

kmod:
I think doing this as a PR is less noisy than as commits to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ install:
   - ninja libunwind ext_cpython
 
 script:
+  - echo $TRAVIS_BUILD_DIR
+  - echo $TRAVIS_BUILD_DIR
   - ninja -j4 check-pyston
 
 os:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 2.8)
 project(pyston C CXX ASM)
 
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+message(STATUS "SOURCE ${CMAKE_SOURCE_DIR}")
+message(STATUS "BINARY ${CMAKE_BINARY_DIR}")
 
 include(CheckTypeSize)
 include(ExternalProject)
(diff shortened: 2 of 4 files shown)


dagar:
I think I have the same failure locally.

     % ./pyston /home/dagar/git/pyston/test/integration/django_test.py                                                                                                        
    Running 'startproject testsite'
    pyston: /home/dagar/git/pyston/src/codegen/pypa-parser.cpp:178: pyston::InternedString     pyston::readName(pypa::AstExpression &, pyston::InternedStringPool &): Assertion `e
    .type == pypa::AstType::Name' failed.
    zsh: abort (core dumped)  ./pyston /home/dagar/git/pyston/test/integration/django_test.py

kmod:
what happens if you pass the `-x` flag?  (pypa has an issue with django at the moment.)

dagar:
It works with -x once I set PYTHONPATH to pyston/test/integration/django

kmod:
And what happens if you don't?  In case it's not clear this is definitely different behavior than I'm getting :)  Sounds like it was able to import django at first (to run into the -x error) but not subsequently (in order to need PYTHONPATH getting set)?

What directory are you running from?  If you add a `print sys.path` right before the django import, what does it print out?

dagar:
It looks like you've figured it out, but for the record I tried running from the build directory (cmake bin), source directory, test/integration, and test/integration/django before adding to PYTHONPATH.

    dagar@dagar-thinkpad ~/pyston-build % ./pyston -x /home/dagar/git/pyston/test/integration/django_test.py
    Removing the existing 'testsite/' directory
    Running 'startproject testsite'
    Running testsite/manage.py migrate
      File "testsite/manage.py", line 8, in <module>:
        from django.core.management import execute_from_command_line
    ImportError: No module named django.core.management
      File "/home/dagar/git/pyston/test/integration/django_test.py", line 43, in <module>:
        subprocess.check_call([sys.executable, ARGS, "testsite/manage.py", "migrate"])
      File "/home/dagar/pyston-build/from_cpython/Lib/subprocess.py", line 540, in check_call:
        raise CalledProcessError(retcode, cmd)
    subprocess.CalledProcessError: Command '['/home/dagar/pyston-build/pyston', '-xu', 'testsite/manage.py', 'migrate']' returned non-zero exit status 1


    sys.path =  ['/home/dagar/git/pyston/test/integration', '/home/dagar/pyston-build/lib/python21.zip', '/home/dagar/pyston-build/from_cpython/Lib', '/home/dagar/pyston-build/lib_pyston', '/usr/local/from_cpython/Lib', '/usr/local/from_cpython/Lib/lib-dynload', '/usr/local/lib/python2.7/site-packages']

kmod:
thank :) would have saved me some headache ;(

turns out I had a copy of `lib_pyston/django/` lying around which was masking the issue; I had removed it from the source directory but it still existed in the build directory.

dagar:
We probably shouldn't be copying from_cpython and lib_pyston to the build directory. Maybe set sys.path appropriately once we can use the configure script? I'll take a look at what cpython does.

kmod:
Python packaging is a pretty hairy topic :)  If you're interested I'd start off by looking at distutils/sysconfig.py, but I think CPython does some pretty ugly things too (determines if it's in the build directory or if it was installed).  I'm working on making sure that all of this stuff gets tested (the incoming virtualenv test and a subsequent distutils test) at which point it should be easier to play around with.

undingen:
- cpython uses the ```PYTHONHOME``` variable if set
- otherwise it tries to determine if it's got started inside the build directory
- if not is will start with the current executable directory and search for "lib/python2.7/os.py" if not found it will try with the parent directory and repeat this process until there is no parent directory.
- if still not found it will take the ```PREFIX``` path set at configure time 
(Code is in getpath.c)
Hope this helps a little.

PR #452, opened 2015-04-20
Increase test parallelism on pyston_old_parser

kmod:
@dagar says there's some intermittent failures here but I'd like to track them down.


kmod:
Ok I'm going to merge this and we'll just deal with the issues as they come up.

PR #453, opened 2015-04-20
Speedup analysis by using llvms more cache friendly map/set algorithms

undingen:
This change does not change our analysis algorithms but instead speeds them up by using llvms containers. This should not change the generated output.
```
for test/integration/django_test.py
us_compiling_analysis_definedness: 130569 --> 29087
us_compiling_analysis_liveness: 132502 --> 102092
us_compiling_analysis_phis: 229944 --> 188633
us_compiling_analysis_types: 214522 --> 36165
```

### DIFF
--- a/src/analysis/fpc.h
+++ b/src/analysis/fpc.h
@@ -17,6 +17,10 @@
 
 #include <queue>
 
+#include "llvm/ADT/DenseMap.h"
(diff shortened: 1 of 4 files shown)


kmod:
Wow, that's a pretty impressive speedup for just switching container implementations!

PR #455, opened 2015-04-21
Update libpypa

kmod:
pulls in:
- allow trailing commas in set literals
- nested attribute decorators


kmod:
(I guess I had snuck this in another PR)

Issue #439, opened 2015-04-07
Instructions are broken for ubuntu 14.04


ekarlso:
ubuntu@pyston:~/pyston-build$ CC='clang' CXX='clang++' cmake -GNinja ~/pyston
CMake Error at /usr/share/cmake-2.8/Modules/CMakeDetermineCCompiler.cmake:56 (message):
  Could not find compiler set in environment variable CC:

  clang.
Call Stack (most recent call first):
  CMakeLists.txt:2 (project)

CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_C_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_C_COMPILER
CMake Error: Could not find cmake module file: /home/ubuntu/pyston-build/CMakeFiles/2.8.12.2/CMakeCCompiler.cmake
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_CXX_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_CXX_COMPILER
CMake Error: Could not find cmake module file: /home/ubuntu/pyston-build/CMakeFiles/2.8.12.2/CMakeCXXCompiler.cmake
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_ASM_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_ASM_COMPILER
CMake Error: Could not find cmake module file: /home/ubuntu/pyston-build/CMakeFiles/2.8.12.2/CMakeASMCompiler.cmake
-- Configuring incomplete, errors occurred!
See also "/home/ubuntu/pyston-build/CMakeFiles/CMakeOutput.log".

arbipher:
It seems you are installing pyston using CMake
The error message "Could not find compiler set in environment variable CC" tells CC is missing.
Since CC='clang' in command, you might miss clang.

Instructions for 14.04 is at the end of [INSTALLING.md](https://github.com/dropbox/pyston/blob/master/docs/INSTALLING.md)

The first line is

[...]

Issue #316, opened 2015-02-21
static assert ubuntu 14.10


js7222:
Hi,

I'm currently getting the following static assert failures when running check-test. Does anyone know what the root cause is?

/home/james/pyston/src/codegen/ast_interpreter.cpp
In file included from /home/james/pyston/src/codegen/ast_interpreter.cpp:36:
In file included from /home/james/pyston/src/runtime/generator.h:19:
/home/james/pyston/src/runtime/types.h:266:1: error: static_assert failed ""
static_assert(offsetof(pyston::BoxedClass, gc_visit) == offsetof(struct _typeobject, _gcvisit_func), "");
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/james/pyston/src/runtime/types.h:267:1: error: static_assert failed ""
static_assert(sizeof(pyston::BoxedClass) == sizeof(struct _typeobject), "");
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/james/pyston/src/runtime/types.h:269:1: error: static_assert failed ""
static_assert(offsetof(pyston::BoxedHeapClass, as_number) == offsetof(PyHeapTypeObject, as_number), "");
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/james/pyston/src/runtime/types.h:270:1: error: static_assert failed ""
static_assert(offsetof(pyston::BoxedHeapClass, as_mapping) == offsetof(PyHeapTypeObject, as_mapping), "");
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/james/pyston/src/runtime/types.h:271:1: error: static_assert failed ""
static_assert(offsetof(pyston::BoxedHeapClass, as_sequence) == offsetof(PyHeapTypeObject, as_sequence), "");
[...]

Issue #13, opened 2014-04-04
Change build system?


darnuria:
Have you plan to move to an other build system like Cmake or ~~autotools~~?
This will be in your roadmap #11 ?

Thanks! And sorry to disturb you but this questions seems important.

darnuria:
Hello I viewed this issue still opened in my dashboard. It seems that this #460 PR resolved this bug.
So I think this is time to close it two years later!

A good bug is a resolved and closed bug. :)

kmod:
Can't believe we (by which I mean @dagar) made it happen!

PR #460, opened 2015-04-22
cmake change default build directory, configure script, Fedora support

dagar:
Tried to hijack #441 for this, but couldn't once I deleted the original branch.

This changes the build directories to
$(SRC_DIR)/build/Debug
$(SRC_DIR)/build/Release
$(SRC_DIR)/build/Debug-gcc

I think the usual complaint about polluting the source directory is when the object files are right next to source, but I guess we'll see. It's easy to override BUILD_DIR.


dagar:
@kmod I fixed the configure script here. For reasons I don't fully understand lz4 using CheckTypeSize causes a conflict between our imported version and the older system version. I resolved it by renaming the imported newer version to CheckTypeSizeof.

dagar:
Integration tests failing. Not sure how to get the unordered_map size into this compilation.

    pycrypto_test.py    FAILED (Exited with code 1 (expected code 0))
warning: PCTBuildPy: byte-compiling is disabled, skipping.

In file included from /home/dagar/pyston-build/from_cpython/Include/Python.h:53:0,
                 from src/pycrypto_common.h:27,
                 from src/_fastmath.c:29:
/home/dagar/pyston-build/from_cpython/Include/object.h:458:24: error: ‘SIZEOF_UNORDEREDMAP’ undeclared here (not in a function)
     char _dep_getattrs[SIZEOF_UNORDEREDMAP + sizeof(void *)];
                        ^
error: command 'gcc' failed with exit status 1
  File "/home/dagar/git/pyston/test/integration/pycrypto_test.py", line 26, in <module>:
    subprocess.check_call([sys.executable, "setup.py", "build"], stdout=devnull)
  File "/home/dagar/pyston-build/from_cpython/Lib/subprocess.py", line 540, in check_call:
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/home/dagar/pyston-build/pyston', 'setup.py', 'build']' returned non-zero exit status 1

dagar:
Cmake could copy object.h and dictobject.h to the build directory and insert the actual size, but I'm not sure how I feel about that.

kmod:
Ah ok, sounds like we should switch from having our own config.h to using the existing pyconfig.h.  That entire file should be generated, so we'll eventually want to be using this process to generate it anyway, and it makes sense as a place for `SIZEOF_UNORDEREDMAP` to live.  Plus it lives in from_cpython/Include so it already gets copied into the build directory (or if we change how that works, it will at least still get the same treatment as all the other headers we have to make available).

dagar:
Moved SIZEOF_UNORDEREDMAP to pyconfig.h and renamed to pyconfig.h.in so there's no confusion.

dagar:
I removed the original Makefile instructions because they'll no longer work.

dagar:
Resolves issue #316.

dagar:
Added Fedora dependencies, but ninja is actually ninja-build so we'll need to add a check to the Makefile.

dagar:
Added more header checks to pyconfig.h

kmod:
Sorry, can we split this up into different PRs?  the stuff all looks good but I don't think we're ready to get rid of the makefile build system just yet, so we could get some of this in if we could split out a PR that doesn't remove the old system.  for that PR, how about let's leave the makefile instructions in for now, and limit the pyconfig.h.in changes to the SIZEOF_UNORDEREDMAP define (which we can add makefile support for).

I think the things that make us keep looking at the makefile build system are using a custom gcc build (we've been experimenting with patching it and I don't think we figured out how to get cmake to use the patched version -- maybe @toshok can comment), and the valgrind mode (didn't seem to work, maybe @rntz can comment).

we're getting closer to not needing it though :)

dagar:
`make pyston_gcc` already works with ~/pyston_deps/gcc-4.8.2-install/ and I made Valgrind work through the Makefile.

Do you want to just leave this open until we've resolved that last few issues instead of hacking a size check into the Makefile?

dagar:
I merged with your pyconfig.h changes.

Apologies for the continued merges, the rebase was getting pretty hairy.

dagar:
-closes #13
-closes #316
-closes #439

dagar:
Resolved the merge conflicts. Any outstanding issues here?

dagar:
Resolved .travis.yml merge conflicts

libpypa
```diff
@@ -1 +1 @@
-Subproject commit 89f615a75fbc86d0dda71b407cb704077749c2cc
+Subproject commit 7f14a66674f6d65c717199512e7ca705110c8da1
```
kmod: Was this change intentional?  I wish github had better submodule support, but it looks like this is going back a few revisions so I assume it's a merge conflict.

src/CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ add_library(PYSTON_OBJECTS OBJECT ${OPTIONAL_SRCS}
 		runtime/util.cpp
 		)
 
-add_dependencies(PYSTON_OBJECTS libunwind_patched libunwind pypa ${LLVM_LIBS})
+add_dependencies(PYSTON_OBJECTS FROM_CPYTHON libunwind_patched libunwind pypa ${LLVM_LIBS})
```
kmod: Does it work to make it depend on `copy_stdlib` instead?  (I'm assuming this additional dependency comes from switching the include dir to the copy in the build directory.)

kmod:
Looks great with some minor comments :)  sorry again for the delay.

after we get this merged we can delete a few hundred lines from the makefile :)

libpypa
dagar: Not sure how this slipped in. I'll rebase the bad merge so history is intact and take a final look at the full PR after all the merges.

src/CMakeLists.txt
dagar: Yes, that's better.

dagar:
@kmod - I fixed the libpypa submodule and set the pyston dependency to copy_stdlib instead of building the entire from_cpython.

kmod:
congrats!  long live the new build system :)

PR #463, opened 2015-04-24
Rewrite calls into BoxedMethodDescriptors

kmod:
This involves
- special-casing their `__get__`
- adding some ability to rewrite calls to functions that take *args


toshok:
:100: :+1:

PR #466, opened 2015-04-27
slots (WIP)

tjhance:
There are still a couple of tests failing (including the one with the recursive GC call that I talked about on Gitter... I'm still not sure if that's partially the fault of this diff or not), but I'll go ahead and put it up here. Also, this ended up being one of those diffs where I fix one thing and that leads to fixing another thing...

Anyway, here is what is happening in this diff:
- In `typeNew`, I add support for `__slots__`. This adds a member descriptor for each name in the slots array. Also, if `__dict__` is in the slots, then it adds a dict (well, really an attrs, since it's pyston).
- Hopefully this should reduce allocations since most objects with __slots__ won't have to allocate attrs.
- To garbage collect the slots, we need to keep track of the slot locations in the type object. CPython lists `PyMemberDef`s at the end of a type object. We don't need the full `PyMemberDef` struct (as far as I know--maybe the c api requires it though), but for now, I just put the slot offsets in a variable array in the type object.
- In the process of writing the above, logic, I stumbled across a tangential problem with the way we add a dict (or in our case, attrs) to a class when subclassing a class that is variable-size. This is tricky, since you have to leave the base layout alone, but that means you don't have a fixed offset to put the dict afterwards. CPython solves this by allowing `tp_dictoffset` to be negative, so you can offset it from the *end* of the object. From the [https://docs.python.org/2/c-api/typeobj.html#c.PyTypeObject.tp_dictoffset](python docs):

> If the value is less than zero, it specifies the offset from the end of the instance structure. A negative offset is more expensive to use, and should only be used when the instance structure contains a variable-length part. This is used for example to add an instance variable dictionary to subtypes of str or tuple. Note that the tp_basicsize field should account for the dictionary added to the end in that case, even though the dictionary is not included in the basic object layout.

In the process of supporting this, I had to make sure that string and tuple subclassed `VarBox` and set the `ob_size` field. It's a bit messy and I think more can be done here to clean things up, but I didn't want to dig too far on this since, again, it's kind of tangential.

### DIFF
--- a/from_cpython/Include/object.h
+++ b/from_cpython/Include/object.h
@@ -479,7 +479,6 @@ typedef struct _heaptypeobject {
                                       see add_operators() in typeobject.c . */
     PyBufferProcs as_buffer;
     PyObject *ht_name, *ht_slots;
-    Py_ssize_t nslots;
     /* here are optional user slots, followed by the members. */
 } PyHeapTypeObject;
 
--- a/from_cpython/Include/object.h
+++ b/from_cpython/Include/object.h
@@ -479,6 +479,7 @@ typedef struct _heaptypeobject {
                                       see add_operators() in typeobject.c . */
     PyBufferProcs as_buffer;
     PyObject *ht_name, *ht_slots;
+    Py_ssize_t nslots;
     /* here are optional user slots, followed by the members. */
 } PyHeapTypeObject;
 
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -495,8 +495,27 @@ extern "C" PyObject* PystonType_GenericAlloc(BoxedClass* cls, Py_ssize_t nitems)
     std::string per_name_allocsize_name = "allocsize." + std::string(cls->tp_name);                                    \
     Stats::log(Stats::getStatId(per_name_alloc_name));                                                                 \
     Stats::log(Stats::getStatId(per_name_allocsize_name), size);
+#define ALLOC_STATS_VAR(cls)                                                                                           \
+    if (cls->tp_name) {                                                                                                \
+        std::string per_name_alloc_name = "alloc." + std::string(cls->tp_name);                                        \
+        std::string per_name_alloc_name0 = "alloc." + std::string(cls->tp_name) + "(0)";                               \
+        std::string per_name_allocsize_name = "allocsize." + std::string(cls->tp_name);                                \
+        std::string per_name_allocsize_name0 = "allocsize." + std::string(cls->tp_name) + "(0)";                       \
+        static StatCounter alloc_name(per_name_alloc_name);                                                            \
+        static StatCounter alloc_name0(per_name_alloc_name0);                                                          \
+        static StatCounter allocsize_name(per_name_allocsize_name);                                                    \
+        static StatCounter allocsize_name0(per_name_allocsize_name0);                                                  \
+        if (nitems == 0) {                                                                                             \
+            alloc_name0.log();                                                                                         \
+            allocsize_name0.log(_PyObject_VAR_SIZE(cls, nitems));                                                      \
+        } else {                                                                                                       \
+            alloc_name.log();                                                                                          \
+            allocsize_name.log(_PyObject_VAR_SIZE(cls, nitems));                                                       \
+        }                                                                                                              \
+    }
 #else
 #define ALLOC_STATS(cls)
+#define ALLOC_STATS_VAR(cls)
 #endif
 
 
@@ -555,10 +574,12 @@ extern "C" PyObject* PystonType_GenericAlloc(BoxedClass* cls, Py_ssize_t nitems)
     }                                                                                                                  \
                                                                                                                        \
     void* operator new(size_t size, BoxedClass * cls, size_t nitems) __attribute__((visibility("default"))) {          \
+        ALLOC_STATS_VAR(default_cls)                                                                                   \
         assert(cls->tp_itemsize == itemsize);                                                                          \
         return BoxVar::operator new(size, cls, nitems);                                                                \
     }                                                                                                                  \
     void* operator new(size_t size, size_t nitems) __attribute__((visibility("default"))) {                            \
+        ALLOC_STATS_VAR(default_cls)                                                                                   \
         assert(default_cls->tp_alloc == PystonType_GenericAlloc);                                                      \
         assert(default_cls->tp_itemsize == itemsize);                                                                  \
         assert(default_cls->tp_basicsize == size);                                                                     \
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -440,6 +440,7 @@ public:
     // Add a no-op constructor to make sure that we don't zero-initialize cls
     Box() {}
 
+    void* operator new(size_t size, BoxedClass* cls, size_t nitems) __attribute__((visibility("default")));
     void* operator new(size_t size, BoxedClass* cls) __attribute__((visibility("default")));
     void operator delete(void* ptr) __attribute__((visibility("default"))) { abort(); }
 
@@ -448,9 +449,11 @@ public:
 
     llvm::iterator_range<BoxIterator> pyElements();
 
+    size_t getHCAttrsOffset();
     HCAttrs* getHCAttrsPtr();
     void setDict(BoxedDict* d);
     BoxedDict* getDict();
+
 
     void setattr(const std::string& attr, Box* val, SetattrRewriteArgs* rewrite_args);
     void giveAttr(const std::string& attr, Box* val) {
@@ -481,10 +484,15 @@ extern "C" PyObject* PystonType_GenericAlloc(BoxedClass* cls, Py_ssize_t nitems)
 
 #define DEFAULT_CLASS(default_cls)                                                                                     \
     void* operator new(size_t size, BoxedClass * cls) __attribute__((visibility("default"))) {                         \
-        return Box::operator new(size, cls);                                                                           \
+        return Box::operator new(size, cls, 0);                                                                        \
     }                                                                                                                  \
     void* operator new(size_t size) __attribute__((visibility("default"))) {                                           \
-        return Box::operator new(size, default_cls);                                                                   \
+        return Box::operator new(size, default_cls, 0);                                                                \
+    }
+
+#define ALLOCATABLE_WITH_ITEMS                                                                                         \
+    void* operator new(size_t size, BoxedClass * cls, size_t nitems) __attribute__((visibility("default"))) {          \
+        return Box::operator new(size, cls, nitems);                                                                   \
     }
 
 // The restrictions on when you can use the SIMPLE (ie fast) variant are encoded as
@@ -522,6 +530,7 @@ class BoxVar : public Box {
 public:
     Py_ssize_t ob_size;
 
+    BoxVar() {}
     BoxVar(Py_ssize_t ob_size) : ob_size(ob_size) {}
 };
 static_assert(offsetof(BoxVar, ob_size) == offsetof(struct _varobject, ob_size), "");
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -440,7 +440,6 @@ public:
     // Add a no-op constructor to make sure that we don't zero-initialize cls
     Box() {}
 
-    void* operator new(size_t size, BoxedClass* cls, size_t nitems) __attribute__((visibility("default")));
     void* operator new(size_t size, BoxedClass* cls) __attribute__((visibility("default")));
     void operator delete(void* ptr) __attribute__((visibility("default"))) { abort(); }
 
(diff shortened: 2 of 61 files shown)


src/runtime/objmodel.cpp
```diff
@@ -607,8 +624,11 @@ Box* Box::getattr(const std::string& attr, GetattrRewriteArgs* rewrite_args) {
             rewrite_args->out_success = true;
 
         if (rewrite_args) {
-            if (!rewrite_args->obj_hcls_guarded)
+            if (!rewrite_args->obj_hcls_guarded) {
+                if (cls->attrs_offset < 0)
+                    REWRITE_ABORTED("");
```
kmod: `REWRITE_ABORTED` doesn't actually abort the rewrite, but is just supposed to be a way to figure out where the abort happened; I think you still need to set rewrite_args to NULL.

kmod:
I chatted a bit offline with chris and looked into the GC issue, and it looks like this is an existing problem, but it basically shouldn't ever activate.  Looks like we're adding weaklist pointers on variable-sized classes that shouldn't get them -- you probably need to do something with the code in `BoxedClass::BoxedClass` that sets weaklistoffset, now that that's being handled in typeNew.

tjhance:
OK, with that I think I got the kinks out.

tjhance:
OK, I did a bit of cleaning up with declaring the variable-size class stuff.

tjhance:
got this to pass the build

src/runtime/objmodel.cpp
```diff
@@ -3900,6 +3945,52 @@ static void assertInitNone(Box* obj) {
...
+    /* We must reject an empty name.  As a hack, we bump the
+       length to 1 so that the loop will balk on the trailing \0. */
+    if (n == 0)
+        n = 1;
+    for (i = 0; i < n; i++, p++) {
+        if (!(i == 0 ? isalpha(*p) : isalnum(*p)) && *p != '_') {
+            raiseExcHelper(TypeError, "__slots__ must be identifiers");
+        }
+    }
+}
```
kmod: If these two are the same except for exception handling, we could maybe just make one call the other?  It's not great from a perf standpoint but I don't think that'll be a big deal here.

also, it's nice to add a `noexcept` specifier to functions that throw CAPI exceptions (`valid_identifier` version) both as documentation and to force the code to terminate if an exception does try to propagate out of it.

test/tests/python_slots_test.py
```diff
@@ -0,0 +1,68 @@
...
+g.a = 5
+print g.a
+g.k = 12
+print g.k
+
+try:
+    class G(C, D):
+        pass
+except TypeError as e:
+    print e.message
```
kmod: Could we add a test that adds something like 1000 slot names?  just as an extra check that the memory accesses are safe, since for example class objects are quite large and there's a lot of room in there to overwrite unused fields.

src/core/types.h
```diff
@@ -434,6 +434,7 @@ class Box {
     // Add a no-op constructor to make sure that we don't zero-initialize cls
     Box() {}
 
+    void* operator new(size_t size, BoxedClass* cls, size_t nitems) __attribute__((visibility("default")));
```
kmod: could/should this constructor live in BoxVar?

src/runtime/objmodel.cpp
tjhance: Actually it looks like I copied one from cpython so I could look at it as I re-wrote it in pyston exception-style, and then forgot to delete the cpython one.

test/tests/python_slots_test.py
tjhance: yeah more tests is good :) and I should probably also write a test case for a class with a `__metaclass__`

tjhance:
unfortunately I left my laptop at the office today -__- I'll get to your comments tomorrow.

src/core/types.h
tjhance: yeah I think so. I didn't really notice that these were here... who uses them? I had assumed that all subclasses of `Box` needed to specify their own `operator new`

kmod: DEFAULT_CLASS is basically just a wrapper around `Box::operator new()` that adds a default argument, and C++ classes are free to not use it and just inherit the base version (not sure of any examples that do this).  I think it should be safe to move this to BoxVar and then change DEFAULT_CLASS_VAR to call `BoxVar::operator new`?

tjhance: oh I just realized you were commenting on a line I had added in this diff. >_< I'm so silly

tjhance:
@kmod I implemented all your suggestions

kmod:
hmm so reading this again, I think there's a slight issue -- requesting `strlen + 1` items as ob_size is different from CPython, which doesn't count the null terminator as part of nitems and says that ob_size is `strlen`.  ie for the empty string, we will say `ob_size == 1`, but cpython will say `ob_size == 0`.  I *think* we've caught all cases where people don't use `PyString_GET_SIZE` to get the size of a string, but it just shows that people definitely assume sometimes that ob_size = number of characters not including null terminator.

It looks like the way CPython handles this is by doing setting `tp_basicsize = sizeof(PyString_Object) + 1`.  (Well, actually they include a `char ob_sval[1]` array, but that ends up adding 4 bytes due to padding, so they do `tp_basicsize = offsetof(PyString_Object, ob_sval) + 1`.)  Could we do something like this?

Also, not necessary here, but I guess CPython's implementation shows that the data always lives at a constant offset and that we could use an array to reference it, instead of an extra pointer field that points inside the data structure.  Not necessary for this PR, but would be a bonus to do that too :)

src/core/types.h
```diff
@@ -517,12 +521,57 @@ extern "C" PyObject* PystonType_GenericAlloc(BoxedClass* cls, Py_ssize_t nitems)
...
+        BoxVar* rtn = static_cast<BoxVar*>(mem);                                                                       \
+        rtn->cls = default_cls;                                                                                        \
+        rtn->ob_size = nitems;                                                                                         \
+        return rtn;                                                                                                    \
+    }
+
 // CPython C API compatibility class:
 class BoxVar : public Box {
 public:
     Py_ssize_t ob_size;
```
kmod: Can we add a note that this field gets initialized by `operator new` and not in the constructor?

src/core/types.h
```diff
@@ -517,12 +521,57 @@ extern "C" PyObject* PystonType_GenericAlloc(BoxedClass* cls, Py_ssize_t nitems)
...
 
+    BoxVar() {}
     BoxVar(Py_ssize_t ob_size) : ob_size(ob_size) {}
```
kmod: Can we either kill this constructor or change it to just assert that the size we passed in matches the size that was already set by operator new?

from_cpython/Include/object.h
```diff
@@ -479,6 +479,7 @@ typedef struct _heaptypeobject {
                                       see add_operators() in typeobject.c . */
     PyBufferProcs as_buffer;
     PyObject *ht_name, *ht_slots;
+    Py_ssize_t nslots;
```
kmod: CPython uses ob_size for this field; could we do the same so that we don't have to add a new field?

src/runtime/objmodel.cpp
```diff
@@ -427,14 +418,14 @@ BoxedHeapClass* BoxedHeapClass::create(BoxedClass* metaclass, BoxedClass* base,
...
 }
 
 BoxedHeapClass* BoxedHeapClass::create(BoxedClass* metaclass, BoxedClass* base, gcvisit_func gc_visit, int attrs_offset,
                                        int weaklist_offset, int instance_size, bool is_user_defined, BoxedString* name,
-                                       BoxedTuple* bases) {
-    BoxedHeapClass* made = new (metaclass)
-        BoxedHeapClass(base, gc_visit, attrs_offset, weaklist_offset, instance_size, is_user_defined, name);
+                                       BoxedTuple* bases, size_t nslots) {
+    BoxedHeapClass* made = new (metaclass, nslots)
+        BoxedHeapClass(base, gc_visit, attrs_offset, weaklist_offset, instance_size, is_user_defined, name, nslots);
```
kmod: I can't comment up on that line (why github??), but the BoxedHeapClass ctor calls the BoxedClass ctor which sets ob_size to 0, which we probably don't want.

src/runtime/objmodel.cpp
```diff
@@ -4002,28 +4064,165 @@ Box* typeNew(Box* _cls, Box* arg1, Box* arg2, Box** _args) {
...
+
+    // Add slot offsets for slots of the base
+    // NOTE: I don't think CPython does this
+    if (base->tp_flags & Py_TPFLAGS_HEAPTYPE) {
+        BoxedHeapClass::SlotOffset* slot_offsets = made->slotOffsets();
+        BoxedHeapClass* base_heap_cls = static_cast<BoxedHeapClass*>(base);
+        BoxedHeapClass::SlotOffset* base_slot_offsets = base_heap_cls->slotOffsets();
+        memcpy(&slot_offsets[final_slot_names.size()], base_slot_offsets,
+               base_heap_cls->nslots * sizeof(BoxedHeapClass::SlotOffset));
+    }
```
kmod: ok spent some time trying to think about why we're doing this and cpython doesn't.  you should check what I'm saying and then add a comment if it's along the right lines :P

at first I thought CPython was directly constructing the member descriptors into this array.  it turns out that they just use the space to create PyMemberDef objects, which PyType_Ready will turn into normal descriptor objects.  It looks like they want to keep them around for the same reason we do, which is for GC traversing of the instances of the class.  except that they crawl the inheritance hierarchy (I guess you just need to check the main base?).

so are we just copying them all into here so that we can scan objects more quickly?  I think it's a decent tradeoff to make -- it increases the size of classes that inherit from classes with slots, but makes objects faster to visit with the GC.

also one thing I was confused about was I thought we were also adding member descriptors for the base slots, which would be user-visible.  so I think this copying is just for gc purposes?

src/runtime/types.cpp
```diff
@@ -1958,6 +1981,17 @@ extern "C" PyObject* PyObject_Init(PyObject* op, PyTypeObject* tp) noexcept {
...
+    // Initialize the variables declared in __slots__ to NULL.
+    if (tp->tp_flags & Py_TPFLAGS_HEAPTYPE) {
+        BoxedHeapClass* heap_cls = static_cast<BoxedHeapClass*>(tp);
+        if (heap_cls->nslots > 0) {
+            BoxedHeapClass::SlotOffset* slotOffsets = heap_cls->slotOffsets();
+            for (int i = 0; i < heap_cls->nslots; i++) {
+                *(Box**)((char*)op + slotOffsets[i]) = NULL;
+            }
+        }
+    }
```
kmod: I think we already memset all of the data to zero on allocation -- I changed this from a `... = NULL` to `assert(... == NULL)` and it seems to work ok.

tjhance:
> Also, not necessary here, but I guess CPython's implementation shows that the data always lives at a constant offset and that we could use an array to reference it, instead of an extra pointer field that points inside the data structure. Not necessary for this PR, but would be a bonus to do that too :)

Yep, I thought of that too. Right now we use the `llvm::StringRef` so I'm not sure how to actually implement this suggestion.

from_cpython/Include/object.h
tjhance: yep, you're right

src/runtime/objmodel.cpp
tjhance: Yeah this seems to be what I was thinking. I wanted to put them all here so that GC scanning would be faster and we wouldn't have to crawl the hierarchy.

Yeah so I wasn't sure if we would be required to actually have the `PyMemberDef`s here in a user-visible way, so I'm sort of betting that that we won't be required to (hence why I did something different) but I'm not totally sure. If I'm wrong, well, we'll have to restructure this.

kmod: well it's not very user visible; definitely not at the Python level, and it seems unlikely that there would be extensions that look at them.

let's just comment out PyHeapType_GET_MEMBERS in from_cpython/Include/object.h, and make a note that this difference shouldn't be user visible and help with performance, and I think we should be good to go.

kmod:
ok let's save the "convert BoxedString to have a array at a constant offset" part for another PR (I just noticed that your "clean up BoxedTuple" change does this for tuples).  I was thinking that `BoxedString::s` could be a function that constructed the StringRef on demand.

tjhance:
@kmod I implemented all your suggestions.

The only thing left is to fix this merge conflict: chris added some stat logging code to the `operator new`, which I moved into a macro, so his stuff got removed in the merge. I'll just need to add his code back into the macro.

tjhance:
Okay, I added the allocation stat stuff to the new macro. Assuming the me-ci build passes, this is probably ready.

src/core/types.h
```diff
@@ -446,10 +446,12 @@ class Box {
 
     llvm::iterator_range<BoxIterator> pyElements();
 
+    size_t getHCAttrsOffset();
```
kmod: It looks like we have a couple uses of attrs_offset that should be updated to call this instead -- the ones I see are in objmodel.cpp related to rewriting.  Could you take a look?  You can check out the test I added in #511

kmod: oh shoot I forgot you had handled that by just aborting the rewrite when we run into those cases.  I'll take out that test.

kmod:
I merged in the change I just made today, and then merged this into master.  :star2:

there was another change that was made that I'm surprised didn't conflict, but there's now an extra char in BoxedString for `interned_state`.  I think this means that we can now save 7 bytes over the current approach of using `sizeof(BoxedString) + 1` as the tp_basicsize by doing the CPython trick of doing `offsetof(BoxedString, charbuf) + 1`

tjhance:
Why does that save 7 bytes?

kmod:
I think there's now 7 bytes of padding after `char interned_state`, and `sizeof(BoxedString)` will include that padding, but we could instead use it for storing character data.

PR #467, opened 2015-04-27
Misc stuff for pycrypto

undingen:
Add ```pow()```, ```long.__mod__```, ```__index__```, slice and xrange long arg support,
support for 3arg pow, long("0xBB", 16) and long("0xBBL", 0)

Support for ```__eq__``` method overwriting for old style classes is still missing, but when that's in place we fail only 2 out of 1820 pycrypto tests.


kmod:
Cool, looks good!  Could we get a test for the `pow(1, 1L)` case that you mentioned was tricky?

Also, let's get whatever you're doing to test pycrypto into an integration test, when you get a chance.

PR #468, opened 2015-04-27
minor tester & test fixes for python microrevision compatibility

rntz:


kmod:
We currently only check the last line of stderr (our tracebacks aren't always the same), so I'm not sure we can use the stderr-canonicalization for this, at least for now.

We're probably pretty close to the point that we can check the entire stderr, but maybe the easiest thing for now is to do the canonicalization inside the test program itself?  Something like `print repr(e).replace("because ", "")`.

rntz:
@kmod I've fixed the issues you mention; this should be merge-able now.

kmod:
cool, thanks :)

PR #469, opened 2015-04-27
Stats

toshok:
Adds some useful features for stats collection:

1. outputs the stats even if you Ctrl-C
2. adds Stats::clear(), which just resets the counters to 0 (without removing their names)
3. Stats::dump() now takes a bool argument, if `true` (the default) we output the 0's, otherwise we don't.
3. exposes `__pyston__.dumpStats`/`__pyston__.clearStats` so we can call them from python code

also snuck in a joinRuntime call in the SIGINT handling, so we dump the perf maps when exiting that way as well.


kmod:
cool :)

PR #474, opened 2015-04-29
Add a pycrypto integration test

undingen:
Adds:
- pycrypto integration test
- pyston specific CAPI extension for converting between a python longs and GMP
- old style classes support for ```__eq__``` and ```__ne__```
- cancel OSR if number of live syms exceeds 225

With this changes all pycrypto tests succeed


kmod:
Awesome, that's exciting :)

PR #475, opened 2015-04-29
Stat timers

toshok:
Each thread has a stack of timers (which should have automatic storage.)  creating a timer starts it and pushes it onto the stack, and also pauses the currently running one (if there is one) to avoid double counting.  When a timer goes out of scope it's stopped and popped off the stack, and the previous one on the stack is resumed.

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -284,6 +284,8 @@ public:
 Value ASTInterpreter::execute(ASTInterpreter& interpreter, CFGBlock* start_block, AST_stmt* start_at) {
     threading::allowGLReadPreemption();
 
+    STAT_TIMER(t0, "us_timer_astinterpreter_execute");
+
     void* frame_addr = __builtin_frame_address(0);
     RegisterHelper frame_registerer(&interpreter, frame_addr);
 
@@ -540,6 +542,7 @@ Value ASTInterpreter::visit_jump(AST_Jump* node) {
                 arg_array.push_back(it.second);
             }
 
+            STAT_TIMER(t0, "us_timer_astinterpreter_jump_osrexit");
             CompiledFunction* partial_func = compilePartialFuncInternal(&exit);
             auto arg_tuple = getTupleFromArgsArray(&arg_array[0], arg_array.size());
             Box* r = partial_func->call(std::get<0>(arg_tuple), std::get<1>(arg_tuple), std::get<2>(arg_tuple),
@@ -926,6 +929,8 @@ Value ASTInterpreter::visit_print(AST_Print* node) {
     static const std::string write_str("write");
     static const std::string newline_str("\n");
     static const std::string space_str(" ");
+
+    STAT_TIMER(t0, "us_timer_visit_print");
 
     Box* dest = node->dest ? visit_expr(node->dest).o : getSysStdout();
     int nvals = node->values.size();
--- a/src/core/stats.cpp
+++ b/src/core/stats.cpp
@@ -22,9 +22,108 @@
 namespace pyston {
 
 #if !DISABLE_STATS
-std::vector<long>* Stats::counts;
+extern "C" const char* getStatTimerNameById(int id) {
+    return Stats::getStatName(id).c_str();
+}
+
+extern "C" int getStatTimerId() {
+    return StatTimer::getStack()->getId();
+}
+
+extern "C" const char* getStatTimerName() {
+    return getStatTimerNameById(getStatTimerId());
+}
+
+__thread StatTimer* StatTimer::stack;
+StatTimer::StatTimer(int statid, bool push) {
+    uint64_t at_time = getCPUTicks();
+    _duration = 0;
+    _start_time = 0;
+    _statid = statid;
+
+    if (!push) {
+        _prev = NULL;
+        return;
+    }
+
+    _prev = stack;
+    stack = this;
+    if (_prev) {
+        _prev->pause(at_time);
+    }
+    resume(at_time);
+}
+
+StatTimer::StatTimer(int statid, uint64_t at_time) {
+    _duration = 0;
+    _start_time = 0;
+    _statid = statid;
+    _prev = stack;
+    stack = this;
+    if (_prev) {
+        _prev->pause(at_time);
+    }
+    resume(at_time);
+}
+
+StatTimer::~StatTimer() {
+    assert(stack == this);
+
+    uint64_t at_time;
+    if (!isPaused()) {
+        at_time = getCPUTicks();
+        pause(at_time);
+    } else {
+        // fprintf (stderr, "WARNING: timer was paused.\n");
+        at_time = _last_pause_time;
+    }
+
+    stack = _prev;
+    if (stack) {
+        stack->resume(at_time);
+    }
+}
+
+void StatTimer::pause(uint64_t at_time) {
+    assert(!isPaused());
+    assert(at_time > _start_time);
+    auto cur_duration = _duration;
+    _duration = at_time - _start_time;
+    assert(_duration > cur_duration);
+
+    Stats::log(_statid, _duration);
+
+    _duration = 0;
+    _start_time = 0;
+    _last_pause_time = at_time;
+    // fprintf (stderr, "paused  %d at %lu\n", _statid, at_time);
+}
+
+void StatTimer::resume(uint64_t at_time) {
+    assert(isPaused());
+    _start_time = at_time;
+    // fprintf (stderr, "resumed %d at %lu\n", _statid, at_time);
+}
+
+StatTimer* StatTimer::swapStack(StatTimer* s, uint64_t at_time) {
+    StatTimer* prev_stack = stack;
+    if (stack) {
+        stack->pause(at_time);
+    }
+    stack = s;
+    if (stack) {
+        stack->resume(at_time);
+    }
+    return prev_stack;
+}
+
+std::vector<uint64_t>* Stats::counts;
 std::unordered_map<int, std::string>* Stats::names;
 bool Stats::enabled;
+
+timespec Stats::start_ts;
+uint64_t Stats::start_tick;
+
 StatCounter::StatCounter(const std::string& name) : id(Stats::getStatId(name)) {
 }
 
@@ -38,7 +137,7 @@ int Stats::getStatId(const std::string& name) {
     // hacky but easy way of getting around static constructor ordering issues for now:
     static std::unordered_map<int, std::string> names;
     Stats::names = &names;
-    static std::vector<long> counts;
+    static std::vector<uint64_t> counts;
     Stats::counts = &counts;
     static std::unordered_map<std::string, int> made;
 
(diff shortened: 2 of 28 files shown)


src/core/threading.cpp
```diff
@@ -107,6 +119,15 @@ class ThreadStateInternal {
         StackInfo& stack = previous_stacks.back();
         stack_start = stack.stack_start;
         previous_stacks.pop_back();
+#if SLOWPATH_TIMERS
+        StatTimer* st = StatTimer::getStack();
+        if (st && !st->isPaused())
+            st->pause();
+        stack.next_generator->statTimers = st;
```
kmod: is this right?

src/core/stats.h
```diff
@@ -55,6 +59,59 @@ struct StatCounter {
...
+        _paused = false;
+        _timer.restart(NULL);
+    }
+
+    bool isPaused() const { return _paused; }
+
+    static StatTimer* getStack() { return stack; }
+    static void setStack(StatTimer* s, bool andResume = false) {
+        stack = s;
+        if (andResume && s && s->isPaused()) {
```
kmod: why wouldn't the other stack be paused?

there are a couple places we check isPaused, but it seems like we should just assert(paused)

src/codegen/ast_interpreter.cpp
```diff
@@ -285,6 +285,11 @@ class RegisterHelper {
 Value ASTInterpreter::execute(ASTInterpreter& interpreter, CFGBlock* start_block, AST_stmt* start_at) {
     threading::allowGLReadPreemption();
 
+#if SLOWPATH_TIMERS
+    static int _stid = Stats::getStatId("zzz_slowpath_astinterpreter_execute_us");
+    StatTimer _st(_stid);
+#endif
```
kmod: I would make this into a macro :)

toshok: yeah, that was already on my todo list :)

On Wed, Apr 29, 2015 at 2:20 PM, Kevin Modzelewski <notifications@github.com

toshok:
Okay, I think this is finally ready to be looked at seriously and merged.

I removed in the ad-hoc StatTimer in `__cxa_throw` and `__cxa_begin_catch`, because even though I can explain it and it should work, it causes ticks to not add up.  It's definitely more accurate, though, so either I can take a look again, or once the unwinder goes in we can add StatTimers there.

I split things up some of the timers bit more but there's still splitting work to be done wrt the `chosen_cf` timers.  I also cleaned up `jit.cpp` such that we log the main time in ticks as a stat counter itself (previously it was always printing it to the console, something that breaks the test runner.)  The CPU MHz estimate is also logged with stats.

Ticks and StatTimers both store their durations in terms of cpu ticks, and we convert (for any stat that starts with `us_`) at Stats::dump() time.  Anything that starts with `us_timer` is also accumulated and printed at the end.  So checking to verify timers are working properly can be done by checking `ticks_in_main` vs. `ticks_all_timers` - the two should always be equal.  Here's an example snippet from the stats dump running `minibenchmarks/django-template.py`:

```
$ ./pyston_release -x -s minibenchmarks/django-template.py
took 64.5ms for last iteration
Stats:
estimated_cpu_mhz: 2799.99810
...
ticks_in_main: 93958186739
us_timer_PyEq: 89934
us_timer_PyHasher: 124069
us_timer_PyLt: 996
us_timer_astInterpretFunction: 71019
us_timer_astinterpreter_execute: 367038
us_timer_astinterpreter_jump_osrexit: 3577
us_timer_boxedmethoddescriptor__call__: 380664
us_timer_caching_parse_file: 107982
us_timer_chosen_cf_body_builtins: 5053399
us_timer_chosen_cf_body_jitted: 2528026
us_timer_compileFunction: 904530
us_timer_cpyton_parsing: 196588
us_timer_gc_collection: 1796842
us_timer_generator_toplevel: 13068
us_timer_getTopPythonFrame: 107349
us_timer_gettraceback: 13866249
us_timer_importFrom: 660
us_timer_importStar: 3820
us_timer_interpreted_module_toplevel: 295
us_timer_main_toplevel: 43979
us_timer_slowpath_augbinop: 4759
us_timer_slowpath_binop: 75327
us_timer_slowpath_callattr: 1094992
us_timer_slowpath_compare: 192531
us_timer_slowpath_createBoxedIterWrapperIfNeeded: 10995
us_timer_slowpath_delattr: 59
us_timer_slowpath_delitem: 295
us_timer_slowpath_getPystonIter: 2531
us_timer_slowpath_getattr: 3884737
us_timer_slowpath_getclsattr: 35
us_timer_slowpath_getglobal: 28497
us_timer_slowpath_getitem: 335760
us_timer_slowpath_len: 22
us_timer_slowpath_nonzero: 163599
us_timer_slowpath_runtimecall: 1683543
us_timer_slowpath_setitem: 8890
us_timer_slowpath_setsattr: 94094
us_timer_slowpath_unaryop: 13
us_timer_slowpath_unboxedLen: 23
us_timer_tp_getattr: 40
us_timer_tp_getattro: 38492
us_timer_tp_setattr: 0
us_timer_tp_setattro: 4259
us_timer_typeNew: 272916
us_timer_visit_print: 3
zzz_gc_alloc_bytes: 0
ticks_all_timers: 93958186739
```

maybe it would be better to make two passes over the stat counters, accumulating total ticks, then using a StatCounter inside Stats::dump() to cause `ticks_all_timers` and `ticks_in_main` to be right next to each other.  might also be worth it to print some sort of error when they don't match.

src/core/stats.cpp
```diff
@@ -22,9 +22,108 @@
...
+StatTimer::~StatTimer() {
+    assert(stack == this);
+
+    uint64_t at_time;
+    if (!isPaused()) {
+        at_time = getCPUTicks();
+        pause(at_time);
+    } else {
+        // fprintf (stderr, "WARNING: timer was paused.\n");
+        at_time = _last_pause_time;
```
kmod: when would this happen?

src/core/stats.cpp
```diff
@@ -69,10 +193,22 @@ void Stats::dump(bool includeZeros) {
...
+            if (startswith(pairs[i].first, "us_")) {
+                fprintf(stderr, "%s: %lu\n", pairs[i].first.c_str(),
+                        (uint64_t)((*counts)[pairs[i].second] / cycles_per_us));
+
+            } else
+                fprintf(stderr, "%s: %lu\n", pairs[i].first.c_str(), (*counts)[pairs[i].second]);
+
+            if (startswith(pairs[i].first, "us_timer"))
+                accumulated_stat_timer_ticks += (*counts)[pairs[i].second];
+        }
```
kmod: Could we do something ilke `if (pairs[1].first == "ticks_in_main") ticks_in_main = pairs[i].second;` and then later just check accumulated_stat_timer_ticks against it?

src/runtime/objmodel.cpp
```diff
@@ -125,6 +125,7 @@ static Box* (*callattrInternal3)(Box*, const std::string*, LookupScope, CallRewr
        * (*)(Box*, const std::string*, LookupScope, CallRewriteArgs*, ArgPassSpec, Box*, Box*, Box*))callattrInternal;
 
 size_t PyHasher::operator()(Box* b) const {
+    STAT_TIMER(t0, "us_timer_PyHasher");
```
kmod: should these be SLOWPATH_TIMER's instead?

src/runtime/generator.cpp
```diff
@@ -86,28 +86,34 @@ Context* getReturnContextForGeneratorFrame(void* frame_addr) {
...
+        }
 
-    // we returned from the body of the generator. next/send/throw will notify the caller
-    g->entryExited = true;
-    threading::popGenerator();
+        // we returned from the body of the generator. next/send/throw will notify the caller
+        g->entryExited = true;
+        threading::popGenerator();
+
+        g->timer_time = getCPUTicks(); // store off the timer that our caller (in g->returnContext) will resume at
```
kmod: We'll probably have to guard this (and some other stuff in this file) with `#if SLOWPATH_TIMERS`.

src/core/stats.h
```diff
@@ -22,23 +22,57 @@
...
+
+#define STAT_TIMER_NAME(id) _st##id
+
+#if !DISABLE_STATS && SLOWPATH_TIMERS
+#define SLOWPATH_TIMER(id, name) STAT_TIMER(id, name)
+#define SLOWPATH_TIMER2(id, name, at_time) STAT_TIMER2(id, name, at_time)
+#else
+#define SLOWPATH_TIMER(id, name)
+#define SLOWPATH_TIMER2(id, name, at_time)
+#endif
```
kmod: just a thought, but should we combine SLOWPATH_TIMER and STAT_TIMER?  some of the timers we probably want all the time (as opposed to the slowpath ones which seem to add some overhead), but it seems like it's tricky to remember to call the right type of timer.

kmod:
lgtm with some minor comments.  we should make sure it works when we do `#define SLOWPATH_TIMERS 0`.

we have some bikeshedding to do on the names of the stats :P  but we can always iterate once it's in :)

src/core/stats.cpp
toshok: happens at the end of both the main timer (jit.cpp:373) and generatorEntry (generator.cpp:115)

basically there are a couple cases where we want to mark a timer as paused at a specific time, instead of when the destructor is called at approximately the next line.

Not sure the gymnastics are necessary in jit.cpp anymore, since those were there initially to estimate CPU MHz.  The generator one is to ensure that the timers for different stacks are paused/resumed at the same tick.

src/core/stats.h
toshok: agreed.  I think combining them would be good.  I'm running into a problem where I have nearly too many timers to show on a single page of console output.  At some point we might want to have timer levels of some sort, and we can at that point add another parameter to the `STAT_TIMER`.  we could either disable them altogether at compile time based on the level, or keep them in and prune the output based on it.

src/core/stats.cpp
toshok: yup, definitely.

src/core/stats.cpp
toshok: ah, turns out we do need the gymnastics in jit.cpp.  without them the ticks don't match up and we get a warning.  sweet.

src/runtime/generator.cpp
toshok: I changed the generator ifdefs from `#if SLOWPATH_TIMERS` to `#if !DISABLE_STATS` since in this particular case we want it to always be there - the timer registered in generatorEntry is the toplevel timer for the generator (much like the main timer in jit.cpp)

src/core/stats.h
kmod: hmm thinking about it some more, I think we still want all of our previous timers to run even if we turn the expensive timers off.  maybe we could un-convert some of the ones you converted over?

kmod: oh, and we definitely don't want to have to turn off all stats to get the new stat timers :)  we should probably have the stat timers turned off by default or measure the impact on the benchmark suite.

toshok:
oops.  let's merge this (after travis-ci rebuilds) instead of closing it.

PR #476, opened 2015-04-29
Add support for specifying the encoding inside source files.

undingen:
I expect that there are still bugs hiding but at least the pyston tests pass and pyxl helloworld works.

This patch depends on a not yet merged pypa change https://github.com/vinzenz/libpypa/pull/31 so is not yet ready for merging.


undingen:
- updated to new pypa version which includes the necessary change
- rebased to head
- passes pyxl test suite

undingen:
I updated the pull request again too also include the latest libpypa upper case string prefix change.

kmod:
awesome :)  would it be easy to come up with a test that uses some sort of custom encoding?  it's a bit hard for me to reason about what exactly the added tests do (ex I'm not sure what encoding github decided to display them as, or what they will print out as).

would it be hard to encode the pyxl test you did into a script?

PR #479, opened 2015-04-30
Fix misc smaller issues

undingen:
- fix crash when calling ```copy.deepcopy()```
- add ```long.__floordiv__```
- use ```str.join``` from cpython which supports unicode and doesn't accept non string arguments.
- support ```set``` subclasses
- ```listIAdd()``` can return a subclass from a list

### DIFF
--- a/test/tests/copy_test.py
+++ b/test/tests/copy_test.py
@@ -1,0 +1,12 @@
+import copy
+class C(object):
+    pass
+
+c = C()
+c.a = 1
+c.b = (["str", 1], 2L)
+print c.__dict__.items()
+cc = copy.deepcopy(c)
+del c.a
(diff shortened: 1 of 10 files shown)


test/tests/copy_test.py
```diff
@@ -0,0 +1,12 @@
...
+class C(object):
+    pass
+
+c = C()
+c.a = 1
+c.b = (["str", 1], 2L)
+print c.__dict__.items()
+cc = copy.deepcopy(c)
+del c.a
+print cc.__dict__.items()
```
kmod: tiny, but this really should be `sorted(dict.items())`.

undingen: forget about the ordering, thank for catching it

PR #483, opened 2015-05-02
ignore untracked files in the gflags submodule

tjhance:
so that the untracked pyc files don't cause the repo to be marked as dirty


kmod:
Oh nice, I didn't know you could configure it like that.

PR #484, opened 2015-05-02
implement `real` and `imag` for `int`, `long`, and `float`

tjhance:


tjhance:
(relevant cpython test case is `test_abstract_numbers.py`)

kmod:
Python sure does have a lot of different features...

PR #487, opened 2015-05-02
Clean up the BoxedTuple class

tjhance:
(This depends on a few other PRs so just look at the last commit. As far as I know github doesn't have a great way to do PRs that depend on other PRs that live in a fork. :\ )

Since I made BoxedTuple a subclass of BoxVar, there's no need for an `nelts` field, since we can just use `BoxVar::ob_size`. Also, since the elements are stored at a fixed offset, we don't need a `Box** elts`, we can just make it an array `Box* elts[]`.

### DIFF
--- a/src/runtime/tuple.cpp
+++ b/src/runtime/tuple.cpp
@@ -128,7 +128,7 @@ int BoxedTuple::Resize(BoxedTuple** pv, size_t newsize) noexcept {
 
     if (newsize < t->size()) {
         // XXX resize the box (by reallocating) smaller if it makes sense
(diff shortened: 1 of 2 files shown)


kmod:
lgtm (the last commit) -- maybe the best way to get it in is for you just cherry-pick it onto master?

tjhance:
well it does depend on the other commits which aren't in yet (i.e. the build wouldn't pass if I cherry-picked it in now)

tjhance:
Now that slots is in, this should be good to go, I think.

PR #489, opened 2015-05-03
rename member -> member_descriptor

tjhance:
This was named wrong and it was getting confusing. On top of that, its `tp_name` was wrong.


kmod:
I hope you're right... I can never keep these straight.

kmod:
(and maybe this misnaming is part of why)

PR #490, opened 2015-05-04
list __eq__ and __ne__ work for rhs's that are list subclasses

toshok:
splitting out the smaller self-contained bits of the django-fixes PR/branch.  `==` and `!=` both handle list subclasses.


kmod:
cool :)  any reason to not add this for the other comparisons?

toshok:
none at all.  I'll verify those others work as well.  This was just the
problem with django's lru cache (actually come to think of it I think it
was just list __eq__ there.)

PR #491, opened 2015-05-04
zip() takes varargs. django passes 3 containers to it.

toshok:


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -554,19 +554,36 @@ Box* filter2(Box* f, Box* container) {
...
-    llvm::iterator_range<BoxIterator> range1 = container1->pyElements();
-    llvm::iterator_range<BoxIterator> range2 = container2->pyElements();
+    std::vector<llvm::iterator_range<BoxIterator>> ranges;
+    for (auto container : *containers) {
+        ranges.push_back(container->pyElements());
+    }
 
-    BoxIterator it1 = range1.begin();
-    BoxIterator it2 = range2.begin();
+    std::vector<BoxIterator> iterators;
```
kmod: I think for this vector (as well as `ranges`) we need to use the StlCompatAllocator -- a BoxIterator can wrap an actual Box* value that we might not otherwise have a reference to.  (looks like map() would potentially also need to be updated)

PR #492, opened 2015-05-04
Misc bugfixes and features for protobuf

undingen:
- Add ```round(x,d)```
- handle ```int << int``` overflow, 
- fix ```int(9223372036854775808, 0)```
- add support for assigning a generator to a list slice
...

### DIFF
--- a/src/capi/float.cpp
+++ b/src/capi/float.cpp
@@ -14,6 +14,8 @@
 
 // This file is mostly copied from CPython
 
+#include <cfloat>
+
 #include "Python.h"
 
 #include "core/types.h"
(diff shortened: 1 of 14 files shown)


src/runtime/list.cpp
```diff
@@ -439,7 +439,20 @@ extern "C" Box* listSetitemSlice(BoxedList* self, BoxedSlice* slice, Box* v) {
         v_size = tv->size();
         v_elts = &tv->elts[0];
     } else {
-        RELEASE_ASSERT(0, "unsupported type for list slice assignment: '%s'", getTypeName(v));
+        BoxedList* l = new BoxedList();
+        try {
+            for (auto* b : v->pyElements())
+                listAppendInternal(l, b);
+        } catch (ExcInfo e) {
+            raiseExcHelper(TypeError, "can only assign an iterable");
```
kmod: Hmm I'm assuming that this is just to catch errors from pyElements(), but will this also swallow any exceptions ex that are thrown inside a generator?

src/capi/float.cpp
```diff
@@ -25,8 +27,45 @@ extern "C" {
...
+#else
+    format = unknown_format;
+#endif
+    return format;
+}
+
 static float_format_type double_format, float_format;
-static float_format_type detected_double_format, detected_float_format;
+static float_format_type detected_double_format = _detect_double_format(),
+                         detected_float_format = _detect_float_format();
```
kmod: oh doh @tjhance it looks like we should have added the float getformat stuff to this file (capi/float.cpp) rather than the one that we did in your pr (runtime/float.cpp).  it looks like we had added the variables here already and even have other code that looks at them, so I think it makes sense to have the implementation live here; mind if we move it over?

undingen:
I replaced the ```try { pyElements() } catch``` inside ```listSetitemSlice``` with ```PySequence_Fast``` and removed the explicit ```tuple``` and ```list``` handling.
The ```float_format_type``` thing is still in I will update it when it's moved.

PR #493, opened 2015-05-05
Add im_class field to BoxedInstanceMethod, implement instancemethodRepr

tjhance:
Add the `im_class` field to `BoxedInstanceMethod` and implement `repr`.

There were a lot of places to insert it, hopefully I didn't screw up. It's kind of tricky: for example you can't assume it's a `BoxedClass`. One occasion where it wouldn't be a `BoxedClass` is if you were using old-style classes:

```
>>> class C:
...     def f(self): pass
... 
>>> C.f.im_class
<class __main__.C at 0x7fa97d41e258>
>>> type(C.f.im_class)
<type 'classobj'>
```

I also had to add support for the new field for the `InstanceMethodType` in codegen.

### DIFF
--- a/src/codegen/compvars.cpp
+++ b/src/codegen/compvars.cpp
@@ -75,9 +75,10 @@ std::string ValuedCompilerType<llvm::Value*>::debugName() {
 }
 
 struct RawInstanceMethod {
-    CompilerVariable* obj, *func;
+    CompilerVariable* obj, *func, *im_class;
 
-    RawInstanceMethod(CompilerVariable* obj, CompilerVariable* func) : obj(obj), func(func) {}
+    RawInstanceMethod(CompilerVariable* obj, CompilerVariable* func, CompilerVariable* im_class)
+        : obj(obj), func(func), im_class(im_class) {}
 };
 
 class InstanceMethodType : public ValuedCompilerType<RawInstanceMethod*> {
@@ -108,9 +109,9 @@ public:
         return rtn;
     }
 
-    static CompilerVariable* makeIM(CompilerVariable* obj, CompilerVariable* func) {
+    static CompilerVariable* makeIM(CompilerVariable* obj, CompilerVariable* func, CompilerVariable* im_class) {
         CompilerVariable* rtn = new ValuedCompilerVariable<RawInstanceMethod*>(
-            InstanceMethodType::get(obj->getType(), func->getType()), new RawInstanceMethod(obj, func), true);
+            InstanceMethodType::get(obj->getType(), func->getType()), new RawInstanceMethod(obj, func, im_class), true);
         obj->incvref();
         func->incvref();
         return rtn;
@@ -160,9 +161,10 @@ public:
         assert(im->func);
         ConcreteCompilerVariable* obj = im->obj->makeConverted(emitter, UNKNOWN);
         ConcreteCompilerVariable* func = im->func->makeConverted(emitter, UNKNOWN);
+        ConcreteCompilerVariable* im_class = im->im_class->makeConverted(emitter, UNKNOWN);
 
-        llvm::Value* boxed
-            = emitter.getBuilder()->CreateCall2(g.funcs.boxInstanceMethod, obj->getValue(), func->getValue());
+        llvm::Value* boxed = emitter.getBuilder()->CreateCall3(g.funcs.boxInstanceMethod, obj->getValue(),
+                                                               func->getValue(), im_class->getValue());
 
         obj->decvref(emitter);
         func->decvref(emitter);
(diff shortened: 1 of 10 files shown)


src/codegen/compvars.cpp
```diff
@@ -195,7 +216,8 @@ class InstanceMethodType : public ValuedCompilerType<RawInstanceMethod*> {
 
     int numFrameArgs() override { return obj_type->numFrameArgs() + function_type->numFrameArgs(); }
 };
-std::unordered_map<std::pair<CompilerType*, CompilerType*>, InstanceMethodType*> InstanceMethodType::made;
+std::unordered_map<std::pair<CompilerType*, std::pair<CompilerType*, CompilerType*>>, InstanceMethodType*>
```
kmod: can we use std::tuple here?

tjhance: hash isn't defined for std::tuple

test/tests/function_instancemethod.py
```diff
@@ -117,3 +117,13 @@ def l(inst):
...
+
+### Test instancemethod repr
+print 'test instancemethod repr'
+class C(object):
+    def f(self):
+        pass
+    def __repr__(self):
+        return '(alpacas are cool)'
+print repr(C.f)
+print repr(C().f)
```
kmod: can we add a `im_class` Python attribute?  (we do that for im_func and im_self)

src/codegen/compvars.cpp
```diff
@@ -74,43 +74,62 @@ std::string ValuedCompilerType<llvm::Value*>::debugName() {
...
+    static_assert(offsetof(Box, cls) == 0, "");
+    llvm::Value* cls_ptr = emitter.getBuilder()->CreateConstInBoundsGEP2_32(var->getValue(), 0, 0);
+    llvm::Value* cls = emitter.getBuilder()->CreateLoad(cls_ptr);
+    assert(cls->getType() == g.llvm_class_type_ptr);
+    // cast from BoxedClass* to Box*
+    llvm::Value* cls_as_box = emitter.getBuilder()->CreateGEP(
+        cls, { llvm::ConstantInt::get(g.i32, 0), llvm::ConstantInt::get(g.i32, 0), llvm::ConstantInt::get(g.i32, 0) });
+    assert(cls_as_box->getType() == g.llvm_value_type_ptr);
+    return new ConcreteCompilerVariable(UNKNOWN, cls_as_box, false);
+}
```
kmod: I think this is probably better as a method of the CompilerType class, since for a bunch of compiler types we will know exactly what the class is and they can override the behavior.  can maybe punt on that for now since we might not need/want this function.

src/runtime/types.cpp
```diff
@@ -919,13 +919,18 @@ Box* instancemethodGet(BoxedInstanceMethod* self, Box* obj, Box* type) {
...
+        return self;
+    }
 
-    return new BoxedInstanceMethod(obj, self->func);
+    return new BoxedInstanceMethod(obj, self->func, self->im_class);
 }
 
 Box* instancemethodNew(BoxedClass* cls, Box* func, Box* self, Box** args) {
     Box* classObj = args[0];
+    RELEASE_ASSERT(PyType_Check(classObj), "need to check this and throw an exception");
```
kmod: is this right?  this (and the cast to BoxedClass a little bit later) seems unnecessary

src/codegen/compvars.cpp
kmod: nor is it by default for std::pair :P  check out core/common.h if you're interested

src/codegen/compvars.cpp
```diff
@@ -74,43 +74,62 @@ std::string ValuedCompilerType<llvm::Value*>::debugName() {
...
-    static std::unordered_map<std::pair<CompilerType*, CompilerType*>, InstanceMethodType*> made;
+    static std::unordered_map<std::pair<CompilerType*, std::pair<CompilerType*, CompilerType*>>, InstanceMethodType*>
+        made;
 
-    CompilerType* obj_type, *function_type;
-    InstanceMethodType(CompilerType* obj_type, CompilerType* function_type)
-        : obj_type(obj_type), function_type(function_type) {}
+    CompilerType* obj_type, *function_type, *class_type;
+    InstanceMethodType(CompilerType* obj_type, CompilerType* function_type, CompilerType* class_type)
+        : obj_type(obj_type), function_type(function_type), class_type(class_type) {}
```
kmod: It looks like class_type is unused -- we include the other two types here since they are needed for type analysis, but it seems like we don't need to add class_type?  ie the type behavior is already determined by the pair (obj_type, function_type).

src/codegen/compvars.cpp
```diff
@@ -1476,7 +1498,8 @@ class NormalObjectType : public ConcreteCompilerType {
             if (rtattr->cls == function_cls) {
                 CompilerVariable* clattr = new ConcreteCompilerVariable(
                     typeFromClass(function_cls), embedRelocatablePtr(rtattr, g.llvm_value_type_ptr), false);
-                return InstanceMethodType::makeIM(var, clattr);
+
+                return InstanceMethodType::makeIM(var, clattr, getClassCompilerVariable(emitter, var));
```
kmod: We could replace `getClassCompilerVariable()` with `new ConcreteCompilerVariable(UNKNOWN, embedRelocatablePtr(cls, g.llvm_value_type_ptr), false)`

test/tests/function_instancemethod.py
kmod: and we should definitely test what happens with old-style classes :)

tjhance:
@kmod I addressed all your comments

kmod:
Awesome, thanks :)  I didn't check the logic too carefully so I'm trusting it's the same as CPython :P

PR #495, opened 2015-05-05
fix the remaining subclass checks for list comparisons <, <=, >, >=

toshok:
basically the same as 5209849, but for the other comparisons.  also add `subclass op subclass` comparison tests so that we ensure pyston doesn't rewrite the op to the reverse and succeed due to one of them being `list_cls`.


kmod:
cool, thanks :)

PR #496, opened 2015-05-05
Add support for weakref.proxy

undingen:
I had to change ```new_weakref``` to use ```tp_alloc``` instead of ```PyObject_GC_New```
in order to create a weak ref because ```PyObject_GC_New``` allocated memory is scanned
conservatively.
```tp_alloc``` points to ```PystonType_GenericAlloc``` and is of PYTHON type.

@toshok Maybe you could also take a look at it and check if this makes sense.


undingen:
I added a test for setattr and implemented PyObject_SetAttr

toshok:
yay awesome

PR #461, opened 2015-04-22
[WIP] travis-ci enable core dumps

dagar:

PR #497, opened 2015-05-06
Have the tester print out core dumps

kmod:
based on dagar's #461, but do the dumping in the tester.


kmod:
ended up adding this in #499; if you pass the `-G` flag it will attach gdb on segfault.

PR #498, opened 2015-05-07
put back in the Timer-based stats I removed in the stats-timer branch…

toshok:
…, and allow stat timers to be disabled separately from the rest of the stats (disable them by default.)


kmod:
cool, thanks :)

PR #500, opened 2015-05-07
Implement built-in input() function.

rudi-c:
First commit!

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -1012,8 +1012,31 @@ Box* rawInput(Box* prompt) {
 }
 
 Box* input(Box* prompt) {
-    fatalOrError(PyExc_NotImplementedError, "unimplemented");
-    throwCAPIException();
+    char* str;
+
+    PyObject* line = raw_input(prompt);
+    if (line == NULL)
+        throwCAPIException();
+
+    if (!PyArg_Parse(line, "s;embedded '\\0' in input line", &str))
+        throwCAPIException();
+
+    // CPython trims the string first, but our eval function takes care of that.
+    // while (*str == ' ' || *str == '\t')
+    //    str++;
+
+    Box* gbls = globals();
+    Box* lcls = locals();
+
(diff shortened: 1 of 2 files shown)


rudi-c:
I like 2 also.

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -954,8 +954,29 @@ Box* rawInput(Box* prompt) {
...
+    //    str++;
+
+    // CPython has these safety checks that the builtin functions exist
+    // in the current global scope. We shouldn't have this problem so these
+    // checks are omitted.
+    // if (PyDict_GetItemString(globals, "__builtins__") == NULL) {
+    //     if (PyDict_SetItemString(globals, "__builtins__",
+    //                              PyEval_GetBuiltins()) != 0)
+    //         return NULL;
+    // }
```
kmod: Is this for cases like `eval(input(), {})`?

test/tests/input_fail.py
```diff
@@ -0,0 +1,11 @@
...
+
+# test failure case to make sure the exception types match
+import StringIO
+import sys
+
+orig_stdin = sys.stdin
+sio = StringIO.StringIO("abcd")
+sys.stdin = sio
+print repr(input())
+sys.stdin = orig_stdin
```
kmod: Minor stylistic note, but we should try to avoid `# should_error` tests if we can (and usually we can now).  those have bitten us in the past since if we add a typo to the test, all of a sudden we're only testing whether we detect the typo the same way that CPython does.

to verify if an exception is thrown usually we can do something like
```python
try:
    print repr(input())
except SyntaxError:
    print "caught expected syntax error"
```

src/runtime/builtin_modules/builtins.cpp
kmod: by which I mean `eval('input()', {})`

toshok:
awesome, lgtm.  could you squash the commits down to 1?  I'll get this merged after.

toshok:
:+1:

PR #503, opened 2015-05-07
StatTimer comparisons

toshok:
if `PYSTON_TIMERS_IN_TICKS` environment variable is set, don't convert to usec on output.

tools/timer_diff.py outputs differences (above a threshold) from before/after stats dumps and color codes green the positive changes and red the negative ones.  for instance this output shows a good change (everything here is green):

```
$ python tools/timer_diff.py before after
 us_timer_caching_parse_file, before 13064175701, after 272745376, delta -12791430325 (approx -4.57 sec) 
 us_timer_chosen_cf_body_builtins, before 5193281273, after 3538990235, delta -1654291038 (approx -590.82 msec) 
 us_timer_gettraceback, before 39122506231, after 29284971184, delta -9837535047 (approx -3.51 sec) 
 us_timer_raiseraw, before 16759458630, after 11021429623, delta -5738029007 (approx -2.05 sec) 
 us_timer_slowpath_callattr, before 3185888676, after 1009675902, delta -2176212774 (approx -777.22 msec) 
```

### DIFF
--- a/tools/timer_diff.py
+++ b/tools/timer_diff.py
@@ -1,0 +1,94 @@
+import sys
+
+estimated_mhz = 0
+
+def tally(fn):
+    counts = []
+
+    for l in open(fn):
+        if ':' in l:
+            (key, value) = l.split(':')
+            if key == "estimated_cpu_mhz":
+                mhz = float(value)
+            if key.startswith('us_timer_'):
+                counts.append((key, int(value)))
+
+    return (counts, mhz)
+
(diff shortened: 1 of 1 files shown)


kmod:
hmm why do we want to compare the ticks and not the microseconds?

toshok:
reviving this PR.  I've updated it to the deal with microseconds (instead of ticks) in the output.

PR #507, opened 2015-05-08
miscellaneous stuff that gets us a bit farther in test_math

tjhance:


src/runtime/float.cpp
```diff
@@ -33,12 +35,37 @@ extern "C" PyObject* PyFloat_FromString(PyObject* v, char** pend) noexcept {
...
-        return static_cast<BoxedInt*>(o)->n;
-    Py_FatalError("unimplemented");
-    return 0.0;
+
+    // special case: int (avoids all the boxing below, and int_cls->tp_as_number
+    // isn't implemented at the time of this writing)
+    if (PyInt_Check(o))
+        return (double)static_cast<BoxedInt*>(o)->n;
+    // special case: long
+    if (PyLong_Check(o))
```
kmod: do you know if these should be CheckExact?

tjhance: Ah... is `tp_as_number` overrideable? I don't really get how inheritance of
the tp_ fields works in general.

kmod:
Also, if you feel like it, we could switch the CPython test suite to run out of the from_cpython/Lib/test directory instead of symlinking them into the test/cpython directory; this would let us put the data files in the from_cpython directory where they probably belong.

src/runtime/float.cpp
kmod: I think yes they are overrideable, so the int and long special cases should be CheckExact (or maybe something like `PyInt_Check(o) && o->cls->tp_as_number->nb_float == int_cls->tp_as_number->nb_float`).  But! the float case should stay as PyFloat_Check (not CheckExact) since that's what CPython does.

Here's a test case; the math methods call PyFloat_AsDouble on their arguments so I'm using that to test it:
```python
class F(float):
    def __float__(self):
        print "F.__float__"
        return 1.0

class I(int):
    def __float__(self):
        print "I.__float__"
        return 1.0

import math
print math.sqrt(F(0.0)) # doesn't call F.__float__, returns 0.0
print math.sqrt(I(0))   # does call I.__float__, returns 1.0
```

kmod: other than this lgtm

PR #508, opened 2015-05-08
Add support for string interning

undingen:
This implements the ```intern()``` function and interning support for our string type.
Sadly this is also increases the size of our string type.
We could use this in much more places if performance and memory benchmarking shows it's an improvement.


undingen:
this is based on #506.

PR #509, opened 2015-05-08
Add more imp.* functions, set correct module.__file__, import * with non module globals

undingen:


kmod:
I added the `quick_check` makefile target, which makes sure to build the unittests early on :P

src/runtime/import.cpp
```diff
@@ -767,10 +771,63 @@ Box* impReleaseLock() {
...
+#ifdef WIN32
+            "msvcrt", "_locale",
+#endif
+            "_subprocess", "_codecs", "_weakref", "_hotshot", "_random", "_bisect", "_heapq", "_lsprof", "itertools",
+            "_collections", "_symtable", "mmap", "_csv", "_sre", "parser", "_winreg", "_struct", "datetime",
+            "_functools", "_json", "xxsubtype", "zipimport", "zlib",
+
+            /* CJK codecs */
+            "_multibytecodec", "_codecs_cn", "_codecs_hk", "_codecs_iso2022", "_codecs_jp", "_codecs_kr", "_codecs_tw",
+            "marshal", "imp", "__main__", "__builtin__", "sys", "exceptions", "_warnings", "_io", "__pyston__" };
```
kmod: Would it work to check if the name is in sys.builtin_module_names?

undingen:
Changed to a comparison with the ```sys.builtin_module_names``` list - much nicer :-)
I rarely break the unit tests so I was betting that I wouldn't break them this time too... :-D

undingen:
rebased to head to resolve the conflict

PR #510, opened 2015-05-08
Singleton hcls storage strategy

kmod:


src/runtime/dict.cpp
```diff
@@ -249,6 +248,16 @@ extern "C" int PyDict_SetItemString(PyObject* mp, const char* key, PyObject* ite
 
 extern "C" PyObject* PyDict_GetItem(PyObject* dict, PyObject* key) noexcept {
     ASSERT(isSubclass(dict->cls, dict_cls) || dict->cls == attrwrapper_cls, "%s", getTypeName(dict));
+    if (isSubclass(dict->cls, dict_cls)) {
+        BoxedDict* d = static_cast<BoxedDict*>(dict);
+        auto it = d->d.find(key);
```
toshok: `return d->getOrNull(key);` ?

kmod: ah nice, that function has been helpful :)  I'll update this.

PR #512, opened 2015-05-10
Get `cpython/test_property.py` to pass

tjhance:


kmod:
lgtm, though personally I'd keep code that's copied from CPython as close to the original version as possible unless there's some reason not to (ie not convert propertyDocCopy to use our internal methods).

PR #514, opened 2015-05-10
support arg unpacking

tjhance:
Support arg unpacking

This doesn't work for single-element tuples, due to a bug in pypa. (I submitted a PR to dropbox/libpypa to fix this.)

### DIFF
--- a/src/core/cfg.cpp
+++ b/src/core/cfg.cpp
@@ -2523,6 +2523,8 @@ CFG* computeCFG(SourceInfo* source, std::vector<AST_stmt*> body) {
                 AST_Name* arg_name_expr
                     = new AST_Name(arg_name, AST_TYPE::Load, arg_expr->lineno, arg_expr->col_offset);
                 visitor.pushAssign(arg_expr, arg_name_expr);
+            } else {
+                assert(arg_expr->type == AST_TYPE::Name);
             }
             counter++;
         }
--- a/src/core/cfg.cpp
+++ b/src/core/cfg.cpp
@@ -2502,6 +2502,27 @@ CFG* computeCFG(SourceInfo* source, std::vector<AST_stmt*> body) {
         }
     }
 
+    if (source->ast->type == AST_TYPE::FunctionDef) {
+        // Unpack tuple arguments
+        // Tuple arguments get assigned names ".0", ".1" etc. So this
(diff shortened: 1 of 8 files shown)


tjhance:
I also got the `global_and_local.py` test to pass, which checks that we throw an exception when you try to declare a param name as global.

(A bit unrelated, but I put it in the same PR to avoid conflicts.)

tjhance:
I updated the libpypa submodule and added a test case for the single-element tuple argument.

undingen:
I think BeautifulSoup.py depends on this change because it uses:
```python
convert = lambda(k, val): (k, re.sub("&(#\d+|#x[0-9a-fA-F]+|\w+);",
                                           self._convertEntities,
                                           val))

```
Whats the status on this pull request?

tjhance:
Ready as far as I know, it just needs review.

tjhance:
Added arg unpacking support for lambdas.

test/tests/global_and_local.py
```diff
@@ -1,10 +1,9 @@
-# expected: fail
 # - this particular check isn't implemented yet
```
kmod: we should probably remove this comment too?

src/core/cfg.cpp
```diff
@@ -2502,6 +2502,32 @@ CFG* computeCFG(SourceInfo* source, std::vector<AST_stmt*> body) {
...
+            args = ast_cast<AST_Lambda>(source->ast)->args;
+        }
+        int counter = 0;
+        for (AST_expr* arg_expr : args->args) {
+            if (arg_expr->type == AST_TYPE::Tuple) {
+                InternedString arg_name = source->getInternedStrings().get("." + std::to_string(counter));
+                AST_Name* arg_name_expr
+                    = new AST_Name(arg_name, AST_TYPE::Load, arg_expr->lineno, arg_expr->col_offset);
+                visitor.pushAssign(arg_expr, arg_name_expr);
+            }
```
kmod: are Tuples the only things that can get this treatment?  a quick test seems to indicate it, but an assert that that the type is a tuple or a name could make that clear.

kmod:
lgtm

tjhance:
@kmod implemented your comments

PR #515, opened 2015-05-10
turn on ConstructorInitializerAllOnOneLineOrOnePerLine

tjhance:
The really long constructor initializer lists are pretty hard to read. This turns on the ConstructorInitializerAllOnOneLineOrOnePerLine option, which puts them one-per-line.

I realize that format style changes can be contentious :P so feel free to ignore if you disagree.

### DIFF
--- a/src/.clang-format
+++ b/src/.clang-format
@@ -9,6 +9,7 @@ AllowShortIfStatementsOnASingleLine: false
 BreakBeforeBraces: Attach
 BreakConstructorInitializersBeforeComma: false
 ColumnLimit: 120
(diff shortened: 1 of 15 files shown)


toshok:
said this on gitter, but repeating here:

I almost never scroll/page to find code, so the increase in line count won't bother me, and it definitely helps readability.  :+1:

kmod:
ok done :)

while we're on the topic of formatting, we have a pretty big column limit (120).  I've always wanted to try that but after running with it for a while, I'm starting to find it annoying.  I tried reducing it to 100 but the diff is 5k lines :(

PR #517, opened 2015-05-11
use llvm::StringRef instead of std::string typeNew

toshok:
found this while adding more timers and figured it was small enough to just fix instead of creating a trello card for it.

turns out we're converting from const char* to std::string (including the malloc/free) for num_slots * num_typeNew_calls.

### DIFF
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -1482,7 +1482,7 @@ static slotdef slotdefs[]
         SQSLOT("__contains__", sq_contains, slot_sq_contains, wrap_objobjproc, "x.__contains__(y) <==> y in x"),
         SQSLOT("__iadd__", sq_inplace_concat, NULL, wrap_binaryfunc, "x.__iadd__(y) <==> x+=y"),
         SQSLOT("__imul__", sq_inplace_repeat, NULL, wrap_indexargfunc, "x.__imul__(y) <==> x*=y"),
-        { NULL, 0, NULL, NULL, NULL, 0 } };
+        { "", 0, NULL, NULL, "", 0 } };
 
 static void init_slotdefs() noexcept {
     static bool initialized = false;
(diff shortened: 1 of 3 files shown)


kmod:
Sounds good; what do you think about renaming wrapper_def to WrapperDef?  ie wrapper_def was supposed to be a copy of the CPython version (but I guess I got the name wrong, it's called wrapperbase in cpython), and if we change the types we might want to either change the name or have some comments explaining the difference.

Another way this could go is have Box::getattr and giveAttr (which I think are the current allocation sites this pr fixes?) take llvm::StringRef's instead.

toshok:
```
Saving results as 'fewer-string-allocations'
              pyston django_migrate.py                  :    2.2s baseline: 2.2 (-1.2%)
              pyston virtualenv_bench.py                :    5.9s baseline: 5.9 (-0.4%)
              pyston django-template.py                 :   11.3s baseline: 11.4 (-1.1%)
              pyston interp2.py                         :    5.4s baseline: 5.5 (-1.7%)
              pyston raytrace.py                        :    6.4s baseline: 6.3 (+0.1%)
              pyston nbody.py                           :    8.3s baseline: 8.2 (+1.5%)
              pyston fannkuch.py                        :    7.3s baseline: 7.6 (-3.6%)
              pyston chaos.py                           :   21.4s baseline: 20.5 (+4.3%)
              pyston fasta.py                           :    5.1s baseline: 5.3 (-4.5%)
              pyston pidigits.py                        :    5.6s baseline: 5.6 (+0.2%)
              pyston richards.py                        :    2.5s baseline: 2.5 (-1.0%)
              pyston deltablue.py                       :    1.9s baseline: 1.8 (+1.7%)
              pyston (geomean-1a6f)                     :    5.6s baseline: 5.6 (-0.5%)
```

not a huge savings, but helpful for the more real world benchmarks.

kmod:
Not sure I believe the performance numbers but it seems like a good change :)

PR #520, opened 2015-05-12
Add pyxl hello world integration test

undingen:


kmod:
cool, thanks :)  with more virtualenv-based tests we'll probably need to look into virtualenv performance again soon.

PR #525, opened 2015-05-13
Fewer StopIteration exceptions

toshok:
convert both generators and PyCallIter (used for the sre match iteration code, amongst other places) to use `__hasnext__`-style iteration.  This reduces the number of exceptions thrown by django, and trims about 3s off the django integration test (using pyston_dbg).

the gcc warning fix and `fact()` change were to get make check to succeed for me.

### DIFF
--- a/src/capi/types.h
+++ b/src/capi/types.h
@@ -139,7 +139,7 @@ public:
         if (flags == PyWrapperFlag_KEYWORDS) {
             wrapperfunc_kwds wk = (wrapperfunc_kwds)wrapper;
             rtn = (*wk)(self->obj, args, self->descr->wrapped, kwds);
-        } else if (flags == 0) {
+        } else if (flags == PyWrapperFlag_PYSTON || flags == 0) {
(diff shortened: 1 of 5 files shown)


src/capi/types.h
```diff
@@ -139,7 +139,7 @@ class BoxedWrapperObject : public Box {
         if (flags == PyWrapperFlag_KEYWORDS) {
             wrapperfunc_kwds wk = (wrapperfunc_kwds)wrapper;
             rtn = (*wk)(self->obj, args, self->descr->wrapped, kwds);
-        } else if (flags == 0) {
+        } else if (flags == PyWrapperFlag_PYSTON || flags == 0) {
```
kmod: is there a more general way we could handle this?  (ie what if we have `PyWrapperFlag_PYSTON | PyWrapperFlag_KEYWORDS`

toshok: hm, maybe the if could be written as:

```
if (flags & PyWrapperFlag_KEYWORDS) {
}
else {
}
```

will try that.

PR #527, opened 2015-05-13
fix linter to allow comments after includes

rntz:
sometimes it's helpful to add a comment after an `#include` to say what it's for. This makes the linter understand that.


rntz:
@kmod this should be mergeable

PR #528, opened 2015-05-14
increase size and number of generic ICs slots (used for getBoxedIterWrapperIfNeeded)

toshok:
also add some megamorphic microbenchmarks.  The code changes in this commit fix
iteration-megamorphic1.py.  The second rewrite was larger than 128 bytes and didn't fit
in the slot as allocated.  This fixes some of the easiest megamorphic IC's we're seeing
in the django template test.

iteration-megamorphic2.py is harder for us since we get a new class every time. If
we can figure out a way to guard on just the mro (and guarantee that the mro is shared
across all subclasses that don't specify their own), then we should still have 2 slots
used in iteration-megamorphic2.py

### DIFF
--- a/microbenchmarks/iteration-megamorphic1.py
+++ b/microbenchmarks/iteration-megamorphic1.py
@@ -1,0 +1,26 @@
+class Iterable(object):
+    def __iter__(self):
+        return self
+
+    def __hasnext__(self):
+        False
+
+    def next(self):
+        raise StopIteration()
+
+class IterableSub(Iterable):
+    pass
+
+def do_iter(r):
+    for i in r:
+        pass
+
+
+i1 = Iterable()
+i2 = IterableSub()
+
+f = 0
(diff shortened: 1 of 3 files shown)


toshok:
rebased this to see if it still gives any improvement:

```
                           1ee49a674757637588:  7ab52216c4240b0ca3:
       django_template.py             4.6s (2)             4.6s (4)  +0.0%
            pyxl_bench.py             3.8s (2)             3.8s (4)  -1.0%
sqlalchemy_imperative2.py             5.0s (2)             5.0s (4)  -1.1%
        django_migrate.py             1.8s (2)             1.8s (4)  -0.4%
      virtualenv_bench.py             7.8s (2)             7.7s (4)  -0.3%
                  geomean                 4.1s                 4.1s  -0.5%
```

not a tremendous change, but shows that there's still some wiggle room for iterators.

PR #530, opened 2015-05-15
When loading a shared library let the GC know about the static variables

undingen:


toshok:
are there C extension libraries that don't register GC roots stored in globals (like, do they incref and assume that's enough?)  It should be relatively quick to scan the regions but it's sad that we'll need to.

kmod:
oh nice!  this definitely eases a lot of concerns about all of the CAPI code we're building without looking at.

PR #534, opened 2015-05-18
implement float.hex and float.fromhex

tjhance:
Relevant test is `cpython/test_float.py`

### DIFF
--- a/src/runtime/float.cpp
+++ b/src/runtime/float.cpp
@@ -18,6 +18,7 @@
 #include <gmp.h>
 
 #include "core/types.h"
+#include "runtime/capi.h"
 #include "runtime/inline/boxing.h"
 #include "runtime/long.h"
 #include "runtime/objmodel.h"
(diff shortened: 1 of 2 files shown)


kmod:
hmm I'm not a fan of checking in code that has no test coverage.  if test_float.py isn't close to working, could we add even just a single call to the two functions?

src/runtime/float.cpp
```diff
@@ -1424,6 +1425,385 @@ extern "C" PyObject* _Py_double_round(double x, int ndigits) noexcept {
...
+            x = 15;
+            break;
+        default:
+            x = -1;
+            break;
+    }
+    return x;
+}
+
+// From CPython floatobject.c
```
kmod: hmm my recommendation is for cases like this, just copy the CPython code unmodified and don't adapt any of it to our APIs.  both float_fromhex and float_hex have some minor issues with exception handling (throw both C API and Pyston exceptions), and in general I think the less we diverge from CPython the better :)

in this particular case, you could add a separate floatFromHex() function that just calls CPython's from_fromhex and then only have a single place that we need to convert exceptions.  or probably even better, is copy from CPython all the way up to PyMethodDef, and let BoxedMethodDescriptor handle all of the details of converting between CPython and Pyston sides.  check out str.cpp::string_methods for an example :)

PR #535, opened 2015-05-18
Add support for compile('single') and func.__globals__/func.func_globals

undingen:


src/runtime/types.cpp
```diff
@@ -848,6 +848,15 @@ static Box* functionDefaults(Box* self, void*) {
     return BoxedTuple::create(func->ndefaults, &func->defaults->elts[0]);
 }
 
+static Box* functionGlobals(Box* self, void*) {
+    assert(self->cls == function_cls);
+    BoxedFunction* func = static_cast<BoxedFunction*>(self);
+    if (func->globals)
```
kmod: Having just spent some time today fixing an issue around our handling of globals I'm a bit worried here :P  can we add something like `assert((func->globals == NULL) == (!func->f->source || func->f->source->scoping->areGlobalsFromModule())` (or whatever the right check is)?  we've changed our contracts here a couple times so I think it's nice to have it in an assert.

undingen:
I added the assert and added a test for the non module globals path.

PR #536, opened 2015-05-18
Handle the extensive edge cases related to sequence slicing

rudi-c:


src/runtime/list.cpp
```diff
@@ -1042,12 +1049,30 @@ void setupList() {
...
+    CLFunction* getitem = createRTFunction(2, 0, false, false);
     addRTFunction(getitem, (void*)listGetitemInt, UNKNOWN, std::vector<ConcreteCompilerType*>{ LIST, BOXED_INT });
     addRTFunction(getitem, (void*)listGetitemSlice, LIST, std::vector<ConcreteCompilerType*>{ LIST, SLICE });
     addRTFunction(getitem, (void*)listGetitem, UNKNOWN, std::vector<ConcreteCompilerType*>{ UNKNOWN, UNKNOWN });
     list_cls->giveAttr("__getitem__", new BoxedFunction(getitem));
 
+    list_cls->giveAttr("__getslice__", new BoxedFunction(boxRTFunction((void*)listGetslice, LIST, 3)));
+
+    CLFunction* setitem = createRTFunction(3, 0, false, false);
+    addRTFunction(setitem, (void*)listSetitemInt, NONE, std::vector<ConcreteCompilerType*>{ LIST, BOXED_INT, UNKNOWN });
```
toshok: Not something we *really* need to fix, and doesn't have to be part of this PR, but maybe we should do a pass and remove the `std::vector<ConcreteCompilerType*>`s from `addRTFunction` callsites.  it definitely makes things more compact (there are instances in src/runtime without the type), and it isn't required by the compiler.

undingen:
If you don't mind you could also add support for list self assignment (currently aborts).
Something like this should be enough:
```c++
if (self == value)
        value = _listSlice(self, 0, self->size, 1, self->size);
```

src/runtime/objmodel.cpp
```diff
@@ -3804,7 +3802,85 @@ extern "C" Box* unaryop(Box* operand, int op_type) {
...
+
+// This function decides whether to call the slice operator (e.g. __getslice__)
+// or the item operator (__getitem__).
+Box* callItemOrSliceAttr(Box* target, const std::string* item_str, const std::string* slice_str, Box* slice, Box* value,
+                         CallRewriteArgs* rewrite_args) {
+    // If we don't have a slice operator, fall back to item operator.
+    // No rewriting for the typeLookup here (i.e. no guards if there is no slice operator). These guards
+    // will be added when we call callattrInternal on a slice operator later. If the guards fail, we fall
+    // back into the slow path which has this fallback to the item operator.
+    Box* slice_attr = typeLookup(target->cls, *slice_str, NULL);
```
toshok: I think this typeLookup call needs to use rewrite args so we can emit guards that the type has that attribute (and it has the value) we think it does

rudi-c:
@undingen Sure, but only slice self-assignments right? Item self-assignment currently segfaults in the repl since it doesn't support printing [...] but this seems unrelated.

undingen:
Just the slice self assignment. I saw something like ```l[:] = l``` quite often lately.

src/runtime/objmodel.cpp
```diff
@@ -3914,7 +3908,96 @@ extern "C" Box* unaryop(Box* operand, int op_type) {
...
+        return callattrInternal1(target, item_str, CLASS_ONLY, rewrite_args, ArgPassSpec(1), item);
+    }
+}
+
+// This function decides whether to call the slice operator (e.g. __getslice__)
+// or the item operator (__getitem__).
+Box* callItemOrSliceAttr(Box* target, const std::string& item_str, const std::string& slice_str, Box* slice, Box* value,
+                         CallRewriteArgs* rewrite_args) {
+    // If we don't have a slice operator, fall back to item operator.
+    // No rewriting for the typeLookup here (i.e. no guards if there is no slice operator). These guards
```
rudi-c: I need to update this comment...

src/runtime/objmodel.cpp
```diff
@@ -3914,7 +3908,93 @@ extern "C" Box* unaryop(Box* operand, int op_type) {
...
+    }
+
+    if (!slice_attr) {
+        return callItemAttr(target, item_str, slice, value, rewrite_args);
+    }
+
+    // Need a slice object to use the slice operators.
+    if (slice->cls != slice_cls) {
+        if (rewrite_args) {
+            rewrite_args->arg1->addAttrGuard(offsetof(Box, cls), (uint64_t)slice_cls, /*negate=*/true);
```
toshok: I'm not sure we need this guard here.  callattrInternal will already guard on the specific types of the arguments.

src/runtime/objmodel.cpp
```diff
@@ -3914,7 +3908,93 @@ extern "C" Box* unaryop(Box* operand, int op_type) {
...
+        }
+
+        return callItemAttr(target, item_str, slice, value, rewrite_args);
+    } else {
+        if (rewrite_args) {
+            rewrite_args->arg1->addAttrGuard(offsetof(Box, cls), (uint64_t)slice_cls);
```
toshok: same here

src/runtime/util.cpp
```diff
@@ -26,4 +27,68 @@ void parseSlice(BoxedSlice* slice, int size, i64* out_start, i64* out_stop, i64*
     if (ret == -1)
         throwCAPIException();
 }
+
+void sliceIndex(Box* b, int64_t* out) {
```
toshok: Can we just call `_PyEval_SliceIndex` in places that use this?  or maybe have this method call `_PyEval_SliceIndex` followed by `checkAndThrowCAPIException()` instead of mostly duplicating the body?

toshok:
lgtm otherwise.  It's kind of a maze of special cases, but such is cpython.

Doesn't have to be part of this PR, but while you're currently the expert on the slice/item stuff, could you look at `PySequence_GetItem`/`PySequence_GetSlice`/`PySequence_DelItem` in `capi.cpp`, specifically these comments:

```
// Not sure if this is really the same:
```

would be nice to verify :)

src/runtime/objmodel.cpp
kmod: I would vote for adding guards where we need them, and separately dealing with the resulting duplicate guards (ex peephole optimizer, or remembering what guards we emitted).  ex I think we usually don't need to guard on the arg classes in callattrInternal, but if we fixed that there it would break callers that relied on it doing the guards.

toshok: nod.  so doing the specific guards in here and then setting rewrite_args->args_guarded = true.  In that case shouldn't the guard here be changed to an explicit `cls == slice->cls` check, not a `cls != slice_cls`.

undingen: We should already detect duplicate attr guards:
```c++
void RewriterVar::addAttrGuard(int offset, uint64_t val, bool negate) {
    if (!attr_guards.insert(std::make_tuple(offset, val, negate)).second)
        return; // duplicate guard detected
    rewriter->addAction([=]() { rewriter->_addAttrGuard(this, offset, val, negate); }, { this }, ActionType::GUARD);
}
```

rudi-c:
Updated with performance improvements.

```
pyston (calibration)                      :   1.01s baseline: 1.00 (+0.40%) 
pyston django_template.py                 :   3.89s baseline: 3.87 (+0.30%) 
pyston pyxl_bench.py                      :   3.56s baseline: 3.59 (-0.88%) 
pyston sqlalchemy_imperative2.py          :   4.73s baseline: 4.74 (-0.34%) 
pyston django_migrate.py                  :   1.66s baseline: 1.64 (+1.36%) 
pyston virtualenv_bench.py                :   4.77s baseline: 4.74 (+0.63%) 
pyston interp2.py                         :   3.26s baseline: 3.22 (+1.05%) 
pyston raytrace.py                        :   5.29s baseline: 5.43 (-2.58%) 
pyston nbody.py                           :   6.60s baseline: 6.61 (-0.13%) 
pyston fannkuch.py                        :   6.72s baseline: 6.53 (+3.00%) 
pyston chaos.py                           :  16.32s baseline: 16.85 (-3.13%)
pyston fasta.py                           :   3.85s baseline: 3.98 (-3.29%) 
pyston pidigits.py                        :   4.49s baseline: 4.56 (-1.52%) 
pyston richards.py                        :   1.59s baseline: 1.54 (+3.62%) 
pyston deltablue.py                       :   1.12s baseline: 1.13 (-1.01%) 
pyston (geomean-10eb)                     :   3.92s baseline: 3.92 (-0.23%) 
```

toshok:
not really sure what's changed (just rebase+push?), but after checking things over, it still lgtm.

I took the liberty of running the macrobenchmarks and it looks good:

```
                           649e33f464679d2a02:  309d28a86fa5faa71a:
       django_template.py             3.9s (4)             3.8s (2)  -0.5%
            pyxl_bench.py             3.8s (4)             3.8s (2)  +0.2%
sqlalchemy_imperative2.py             4.5s (4)             4.6s (2)  +0.5%
        django_migrate.py             1.7s (4)             1.7s (2)  +1.7%
      virtualenv_bench.py             4.9s (4)             5.0s (2)  +1.1%
                  geomean                 3.5s                 3.6s  +0.6%
```

rudi-c:
I just added an fast path for built-in types. Built-in types don't really care about getitem vs getslice and having all those extra guards to select between the two caused unecessary slowdown.

toshok:
ah great, thanks for the explanation.  this was in fact my remembrance
(that there was a slowdown).  I wasn't able to reproduce it and couldn't
tell what had changed :)

kmod:
ok lgtm :)  I looked a bit into the performance; it's hard to tell how much the guards are an issue, but I did notice that without the fastpath, list and string slicing is much slower since it hits the un-rewriteable case.

rudi-c:
I'm surprised there was no conflicts (behavior or performance) with https://github.com/dropbox/pyston/commit/5bd967f2c6ba900429307c4ee4874df1d0949eed

Is there anything left for me to do in this PR?

toshok:
:shipit: @rudi-c needs more work!

PR #538, opened 2015-05-19
WIP: add "shapes" for BoxedClass

toshok:
in order to make typeLookup not emit assembly for a given IC slot specialized to
a specific class's mro we need something that is shared between classes.

That something is a 'shape'.  Each class computes its own "local shape", which is
just the sha256 of its attribute names (in ascending order of slot offset).  This
local shape, combined with a parent's total shape, gives a subclass's total shape.

This is similar to the way hidden classes work, but with shapes we want a single
identifier (the sha256 hash) that describes multiple attributes, as opposed to a
separate hidden class for each attribute.

Once we've computed this total shape for a class, we can cache in the IC the index
of the class which has this attribute.  We also should know at this point what the
offset of the actual attribute within this class is.

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -406,6 +406,8 @@ void BoxedClass::finishInitialization() {
     this->tp_dict = this->getAttrWrapper();
 
     commonClassSetup(this);
+
+    computeTotalShape();
 }
 
 LocalShape BoxedClass::computeLocalShape() {
@@ -426,7 +428,7 @@ LocalShape BoxedClass::computeLocalShape() {
         // compute a hash of all attribute names
         SHA256OStream hash_stream;
         for (auto& s : attrs) {
-            hash_stream << s.first;
+            hash_stream << s.first.size() << s.first;
         }
         uint64_t lshape[4];
         hash_stream.getHash(lshape);
@@ -448,7 +450,7 @@ Shape* BoxedClass::computeTotalShape() {
     for (int i = mro->ob_size - 1; i >= 0; i--) {
         BoxedClass* b = static_cast<BoxedClass*>(mro->elts[i]);
 
-        LocalShape b_shape = computeLocalShape();
+        LocalShape b_shape = b->computeLocalShape();
         b->total_shape = shape->getOrMakeChild(b_shape);
         shape = b->total_shape;
     }
@@ -2022,6 +2024,9 @@ void setattrGeneric(Box* obj, llvm::StringRef attr, Box* val, SetattrRewriteArgs
             rewrite_args = NULL;
             REWRITE_ABORTED("");
         }
+
+        self->local_shape = LocalShape::uninitialized();
+        PyType_Modified(self);
     }
 }
 
@@ -4127,6 +4132,9 @@ extern "C" void delattrGeneric(Box* obj, llvm::StringRef attr, DelattrRewriteArg
             rewrite_args = NULL;
             REWRITE_ABORTED("");
         }
+
+        self->local_shape = LocalShape::uninitialized();
+        PyType_Modified(self);
     }
 
     // Extra "use" of rewrite_args to make the compiler happy:
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -407,15 +407,20 @@ void BoxedClass::finishInitialization() {
 
     commonClassSetup(this);
 
-    computeTotalShape();
+    Shape::computeTotalShape(cls);
 }
 
-LocalShape BoxedClass::computeLocalShape() {
-    if (local_shape.isUninitialized()) {
+LocalShape Shape::computeLocalShape(Box* b) {
+    if (!isSubclass(b->cls, type_cls))
+        return LocalShape(b);
+
+    auto cls = static_cast<BoxedClass*>(b);
+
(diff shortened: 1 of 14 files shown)


tjhance:
> as opposed to a separate hidden class for each attribute.

What do you mean by this? A hidden class on an object covers all the attributes in the attrs list, right?

toshok:
right, but the hidden class is independent of a given object/class.

You could have a hierarchy like:

```
class Foo(object):
    pass

class Bar1(Foo):
    attr1 = 5
    attr2 = 10

class Baz1(Foo):
    attr1 = 5

class Baz2(Baz1):
    attr2 = 10
```

and Baz2 would have the same hidden class as Bar1.  We want them to be different.

src/runtime/objmodel.cpp
```diff
@@ -405,6 +412,54 @@ void BoxedClass::finishInitialization() {
...
+        }
+        // sort the vector to hopefully give a deterministic ordering
+        std::sort(attrs.begin(), attrs.end(), [](std::pair<llvm::StringRef, int> a, std::pair<llvm::StringRef, int> b) {
+            return a.second < b.second;
+        });
+
+        // compute a hash of all attribute names
+        SHA256OStream hash_stream;
+        for (auto& s : attrs) {
+            hash_stream << s.first;
```
kmod: I think this can't distinguish between a class with attributes `["a", "bc"]` and `["ab", "c"]`.  There might be a better way to create the aggregated hash, but it seems like a simple fix for now is do something like `hash_stream << s.first.size() << s.first()`, which would make unambiguous the mapping from stream -> attribute set.

src/runtime/objmodel.cpp
```diff
@@ -405,6 +412,54 @@ void BoxedClass::finishInitialization() {
...
+
+    Shape* shape = Shape::root_shape;
+    auto mro = static_cast<BoxedTuple*>(this->tp_mro);
+
+    assert(this == mro->elts[0]);
+
+    for (int i = mro->ob_size - 1; i >= 0; i--) {
+        BoxedClass* b = static_cast<BoxedClass*>(mro->elts[i]);
+
+        LocalShape b_shape = computeLocalShape();
```
kmod: I'm not sure the elements of the mro are guaranteed to be BoxedClasses (I think you can simultaneously inherit from old-style and new-style classes) -- if it is the case, we should add an assert to that effect (`assert(isSubclass(mro->elts[i]->cls, type_cls)`).  If not, maybe these shape calculations don't need to be specific to BoxedClasses?  Though hopefully we would be able to do that without increasing the object size.

kmod: Also, I think this should be `b->computeLocalShape()`

kmod:
sorry if I'm missing it or commenting on an out-of-date branch, but it looks we need something that invalidates the local+total shape if an attribute gets added or deleted?

on the bikeshedding side, what do you think about calling it something like "typeLookupShape" + "localAttributeSet"?  there are a lot of things that could be "shapes", but if I understand it, this one is just about the shape wrt typeLookup.

src/runtime/objmodel.cpp
```diff
@@ -905,43 +960,57 @@ Box* typeLookup(BoxedClass* cls, const std::string& attr, GetattrRewriteArgs* re
...
-                assert(rewrite_args->obj == obj_saved);
-            } else {
-                rewrite_args->obj = rewrite_args->rewriter->loadConst((intptr_t)base, Location::any());
-            }
-            val = base->getattr(attr, rewrite_args);
-            assert(rewrite_args->out_success);
-            if (val)
+        // Guard on the total shape of the class.
+        // from here we can assume that the index into the mro + the offset into the object remain the same.
+        Shape* total_shape = cls->computeTotalShape();
```
kmod: this is probably low on the todo list, but maybe we should call `computeTotalShape` at some point during class initialization?  otherwise, it seems like every new class will fail the guards since the stored totalShape will be null.

also, seems like a decent time to remark that we should probably sort the attributes during typeNew -- they're just sorted by however std::unordered_map decides to iterate over them.

src/runtime/objmodel.cpp
toshok: good call, will fix.

src/runtime/objmodel.cpp
toshok: re: calling `computeTotalShape` at initialization time, I was running into this very problem.  The computation of the shape would be deferred until `typeLookup` was called, but that only happened in the slowpath. So even if a type matched a shape already checked for in the IC, we'd hit the slowpath an additional time to compute it.

src/runtime/objmodel.cpp
toshok: ah, yeah.  surprising that worked at all re: `b->computeLocalShape()`

on the subject of old-style classes, I need to read more about how they're handled.  There isn't really anything *too* specific about the shapes.  I think the only thing BoxedClass specific is the special code we emit for doing the actual getattr on the BoxedClass.  but we could probably make something like that work independent of the type of the thing in the mro.

toshok:
yup, you aren't mistaken.  I was trying to hold off on adding invalidation on the hope that there was some point after type initialization but before the type was used that I could slip a call to `computeTotalShape` in.

toshok:
re: bikeshedding, `typeLookupShape` sounds good.  how about `localAttributeSetHash` ?

PR #539, opened 2015-05-19
Add PyWrapperDescr_Type, allow cls->tp_descr_get, allow overwriting attributes inside PyType_Ready

undingen:


undingen:
Sneaked in the virtualenv update (removes a few hacks)

PR #542, opened 2015-05-20
unwinder

rntz:
@kmod: here's the unwinder's present state. It's also coupled with some changes in libunwind, which you can see at https://github.com/rntz/libunwind/compare/rntz. I don't understand git submodules yet, so I don't know how to get those changes included in this PR.

Problems:
- The commits are a mess; it needs rebasing.
- There are a few miscellaneous non-unwinder-but-useful changes that should probably be a separate PR.
- The code could be organized better; it's not clear where `registerDynamicEhFrame` should go, for example.

Missing features:
- incremental tracebacks
- binary search in libunwind
- GC awareness for `pyston::exception_ferry`
- decrementing IC count when unwinding through an IC


src/codegen/ast_interpreter.cpp
```diff
@@ -265,6 +270,11 @@ void ASTInterpreter::initArguments(int nargs, BoxedClosure* _closure, BoxedGener
     }
 }
 
+// Map from stack frame pointers for frames corresponding to ASTInterpreter::execute() to the ASTInterpreter handling
+// them. Used to look up information about that frame. This is used for getting tracebacks. I think it's also used for
+// CPython introspection (sys._getframe & co).
+//
+// TODO: get kmod to check this is an accurate description
```
kmod: sgtm.  looks like it's currently used for those things as well as part of gc scanning.

src/codegen/irgen/irgenerator.cpp
```diff
@@ -1814,6 +1814,7 @@ class IRGeneratorImpl : public IRGenerator {
         static const std::string newline_str("\n");
         static const std::string space_str(" ");
 
+        // TODO: why are we inline-generating all this code instead of just emitting a call to some runtime function?
```
kmod: probably no good reason; as we've supported more of the print-statement-semantics we've just incrementally added them in.  it's a good question though if we should just switch it to calling a runtime function.

src/codegen/unwinding.cpp
```diff
@@ -502,6 +519,36 @@ static const LineInfo* lineInfoForFrame(PythonFrameIteratorImpl& frame_it) {
...
+//    (a) A JIT-compiled Python function.
+//    (b) ASTInterpreter::execute() in codegen/ast_interpreter.cpp.
+//    (c) generatorEntry() in runtime/generator.cpp.
+//    (d) Something else.
+//
+//    By cases:
+//
+//    (2a, 2b) If the previous frame we visited was an OSR frame (which we know from its CompiledFunction*), then we
+//    skip this frame (it's the frame we replaced on-stack) and keep unwinding. (FIXME: Why are we guaranteed that we
+//    on-stack-replaced at most one frame?) Otherwise, we found a frame for our traceback! Proceed to step 3.
```
kmod: well, at least in our current implementation, OSRs only replace the top-most frame.  we can have multiple OSR'd frames in a row, such as 
1: interpreter frame
2: interpreter->lowest level jit OSR
3: jit -> optimized jit OSR

frames 2 and 3 will have the "is osr" bit set on them, which should mean that frames 1 and 2 get excluded from the traceback.  so in theory we should get one traceback entry per non-osr frame.

src/core/options.cpp
```diff
@@ -75,6 +75,7 @@ bool ENABLE_TYPE_FEEDBACK = 1 && _GLOBAL_ENABLE;
 bool ENABLE_RUNTIME_ICS = 1 && _GLOBAL_ENABLE;
 bool ENABLE_JIT_OBJECT_CACHE = 1 && _GLOBAL_ENABLE;
 
+// TODO(rntz): do we ever not want to have frame introspection? should we remove this option?
```
kmod: agreed that it's pretty rare that we flip this off.  I think we test it every now and then to see what the perf impact is of adding all the extra LLVM args.

src/runtime/capi.cpp
```diff
@@ -667,6 +667,8 @@ void checkAndThrowCAPIException() {
         PyErr_Clear();
 
         // This is similar to PyErr_NormalizeException:
+        // TODO: why are we doing something "similar" to PyErr_NormalizeException instead of calling
+        // PyErr_NormalizeException?
```
kmod: I think the "why" is just because we started off doing something that wasn't quite the same as CPython and we haven't taken the time to make it exactly the same :)  I think there's a good todo in here for making them similar (I think there are still some issues around this)

rntz:
ok, I've cleaned up the commits so they should make more sense now. The content is basically the same, though.

kmod:
can we add 'ignore=all' to the .gitmodules definition for libunwind?  otherwise the tree is marked as dirty after a build.

src/runtime/cxx_unwind.cpp
```diff
@@ -0,0 +1,728 @@
...
+
+extern "C" void __cxa_end_catch() {
+    if (VERBOSITY("cxx_unwind"))
+        printf("***** __cxa_end_catch() *****\n");
+    // See comment in __cxa_begin_catch for why we don't clear the exception ferry here.
+}
+
+extern "C" void __cxa_throw(void* exc_obj, std::type_info* tinfo, void (*dtor)(void*)) {
+    assert(!pyston::in_cleanup_code);
+    assert(exc_obj);
```
kmod: Is there some way to release_assert that tinfo corresponds to ExcInfo?  I think the C++ stdlib throws other kinds of exceptions (out of memory, std::terminate if an exception propagates through a noexcept function, std::logic_error if you call std::string((char*)NULL)`).  We don't need to handle them, and I think the right behavior is just to crash in one way or another, but I assume that if we don't check the typeinfo we will just blindly interpret the value as an excinfo and keep going?

rntz: Unfortunately, I don't think so? The natural way to do it would be

```c++
RELEASE_ASSERT(tinfo == typeid(ExcInfo), "blah");
```

But `typeid` is disabled if we disable RTTI, which we do.

kmod: I *think* we can still do it -- exceptions support will add RTTI for anything that gets thrown (and/or caught?), so the type info is there but I don't think you can access it with `typeid`.  I think if you can find the name it mangles to, you can add a symbol with that name and get at the value :P

kmod: _ZTSN6pyston7ExcInfoE

kmod:
awesome!  just in the nick of time ;)

PR #543, opened 2015-05-20
implement float.hex and float.fromhex

tjhance:
bring in `floatobject.c`

### DIFF
--- a/from_cpython/Include/floatobject.h
+++ b/from_cpython/Include/floatobject.h
@@ -12,16 +12,10 @@ PyFloatObject represents a (double precision) floating point number.
 extern "C" {
 #endif
 
-// Pyston change: this is not the format we're using
-// - actually I think it is but there's no reason to have multiple definitions.
-#if 0
 typedef struct {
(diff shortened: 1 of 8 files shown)


src/runtime/types.h
```diff
@@ -407,6 +407,7 @@ class BoxedFloat : public Box {
 
     DEFAULT_CLASS_SIMPLE(float_cls);
 };
+static_assert(offsetof(BoxedFloat, d) == offsetof(PyFloatObject, ob_fval), "");
```
kmod: can we also get an assert on the size of the objects?

from_cpython/Include/floatobject.h
```diff
@@ -14,14 +14,10 @@ extern "C" {
 
 // Pyston change: this is not the format we're using
 // - actually I think it is but there's no reason to have multiple definitions.
```
kmod: I think this comment should get deleted?

otherwise, lgtm

tjhance:
addressed your comments, @kmod

PR #544, opened 2015-05-20
Misc stuff

undingen:
- Allow changing ```class.__bases__```:
I copied the implementation from cpython but I'm not sure if the ```same_slots_added``` method is correct for pyston.  I'm not sure if I have to also check for additional things.
It looks like we are stricter than cpython (```tp_basicsize``` will more often have a different size) and support less sublassing.

- Add ```attrwrapper.clear()```:
My implementation looks very ugly, is there a cleaner way to clear the attributes?

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -2082,8 +2082,11 @@ static Box* typeBases(Box* b, void*) {
     return type->tp_bases;
 }
 
-static void typeSetBases(Box* b, Box* v, void*) {
-    Py_FatalError("unimplemented");
+static void typeSetBases(Box* b, Box* v, void* c) {
+    RELEASE_ASSERT(isSubclass(b->cls, type_cls), "");
+    BoxedClass* type = static_cast<BoxedClass*>(b);
+    if (type_set_bases(type, v, c) == -1)
+        throwCAPIException();
 }
 
 // cls should be obj->cls.
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -1997,6 +1997,63 @@ static PyObject* object_reduce_ex(PyObject* self, PyObject* args) noexcept {
     return _common_reduce(self, proto);
 }
 
+/*
+   from PEP 3101, this code implements:
+
+   class object:
+       def __format__(self, format_spec):
+       if isinstance(format_spec, str):
+           return format(str(self), format_spec)
+       elif isinstance(format_spec, unicode):
+           return format(unicode(self), format_spec)
+*/
+static PyObject* object_format(PyObject* self, PyObject* args) noexcept {
+    PyObject* format_spec;
+    PyObject* self_as_str = NULL;
+    PyObject* result = NULL;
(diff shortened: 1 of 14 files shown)


from_cpython/Python/mystrtoul.c
```diff
@@ -92,7 +92,7 @@ static int digitlimit[] = {
 **              exceptions - we don't check for them.
 */
 unsigned long
-PyOS_strtoul(register char *str, char **ptr, int base)
+PyOS_strtoul(register const char *str, char **ptr, int base)
```
kmod: hmm I feel like these two arguments should have the same const-ness -- changing the argument to be const so that we can cast the constness away seems like it defeats the purpose :P  my preference would be to have both arguments be const, or I guess we could have them both be non-const and make the caller be ok with casting away the constness?

src/runtime/types.cpp
```diff
@@ -1542,6 +1542,22 @@ class AttrWrapper : public Box {
...
+
+        HCAttrs* attrs = self->b->getHCAttrsPtr();
+        RELEASE_ASSERT(attrs->hcls->type == HiddenClass::NORMAL || attrs->hcls->type == HiddenClass::SINGLETON, "");
+
+        while (true) {
+            const auto& attrMap = attrs->hcls->getStrAttrOffsets();
+            if (attrMap.size() == 0)
+                break;
+            self->b->delattr(attrMap.begin()->first(), NULL);
+        }
```
kmod: I think this should be more efficient than the while loop, if less clear:
```C++
new ((void*)attrs) HCAttrs(root_hcls->getAttrwrapperChild());
self->b->appendNewHCAttr(self, NULL);
```

from_cpython/Python/mystrtoul.c
undingen: While the signature looks strange with the mixed constness I tried to match the signature of ```strtoul``` http://www.cplusplus.com/reference/cstdlib/strtoul/
I will revert this change and cast away the constness at the caller side.

src/runtime/types.cpp
undingen: I tried this one put I had to make ```appendNewHCAttr``` public and and then it crashed so I left it as is :-D.

undingen:
Thanks, I will change the checks

undingen:
I addressed the comment's except the ```static Box* clear(Box* _self)``` thing which I left as is.
I also copied from cpython part of a test for the ```class.__bases__ =``` change in to a separate file because the original test tests much more unrelated stuff which we currently don't support.
I didn't yet squash the last commit to make it easier to review the change.
@tjhance could you please have a look at the last commit and check if I made the changes that you suggested correct.

undingen:
Updated the PR again: changed ```same_slots_added``` to the version I linked to in gitter.

from_cpython/Python/mystrtoul.c
kmod: oh what, that's weird; in that case I think doing it the way you originally had it makes sense :/

PR #546, opened 2015-05-21
remove the llvm::StringRef from BoxedString

tjhance:
Create it on demand whenever we need it. We can do this since we (i) save the size in `ob_size` and (ii) the char data is at a fixed offset.


tjhance:
the most interesting changes (among a sea of boring find-and-replaces) are
- the `class BoxedString` definition
- the implementation of `_PyString_Resize`

kmod:
nice :)

PR #547, opened 2015-05-21
Add pyexpat support

undingen:
I disabled the custom traceback stuff for now because we don't support it yet.
It's using the system libexpat but we could also use cpythons own copy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1189,11 +1189,17 @@ TEST_EXT_MODULE_NAMES := basic_test descr_test slots_test
 TEST_EXT_MODULE_SRCS := $(TEST_EXT_MODULE_NAMES:%=test/test_extension/%.c)
 TEST_EXT_MODULE_OBJS := $(TEST_EXT_MODULE_NAMES:%=test/test_extension/%.pyston.so)
 
-SHAREDMODS_NAMES := _multiprocessing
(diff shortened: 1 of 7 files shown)


kmod:
test? :)

also, how do you feel about adding this to the "sharedmods" (from_cpython/setup.py) instead of the builtins?  I guess the benefit would be that it wouldn't have to get initialized for programs that don't use it, and it's also how cpython does it.

undingen:
Ah I missed the shared module support that's cool (thought the other alternative would be to static link against the custom expat. That's why I thought using the shared from the system is better). 
But I see a problem: currently virtualenv doesn't copy the ```lib_pyston``` folder - I will add it.
Also added the test from cpython which I forgot to add.

kmod:
I thought that virtualenv didn't copy all of the libraries over, and instead put the original directory on sys.path? It seems to work ok to import _multiprocessing from within a virtualenv.

Also, putting them into lib_pyston is a bit of a hack; CPython creates a separate `lib-dynload` directory for them, but I didn't know / feel like figuring out the places we would have to update to add a new directory :P

undingen:
I'm kind of surprised that this works but can conform that it also works here :-D.
Looks like I added the ```lib_pyston``` path already to the ```site.py``` file.

PR #549, opened 2015-05-23
Float pow

tjhance:
fix float.pow, which is wrong

The actual float.pow covers a bunch of special cases that the c++ stdlib `pow` doesn't cover (e.g. `1.0 ** large_number` is always 1.0), and there are exceptions to throw and stuff. So just use `float_pow` from `floatobject.c`.

I refactored `float_pow` in floatobject.c in order to add an unboxed version `float_pow_unboxed`, which I call directly in the implementation of `float_pow_pow` (our specialization of pow to unboxed floats).

### DIFF
--- a/from_cpython/Objects/floatobject.c
+++ b/from_cpython/Objects/floatobject.c
@@ -820,12 +820,14 @@ float_floor_div(PyObject *v, PyObject *w)
    x is not an infinity or nan. */
 #define DOUBLE_IS_ODD_INTEGER(x) (fmod(fabs(x), 2.0) == 1.0)
 
-// pyston change: make not static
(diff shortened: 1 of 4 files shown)


tjhance:
Note: I still need to do the equivalent work for int pow

kmod:
tests?  otherwise lgtm.

tjhance:
Well I figured we already had pow tests somewhere, and we pass additional tests in cpython/float_test.py.

But sure, I'll add some tests when I get a chance.

tjhance:
back from vacation!

added a unit test

kmod:
awesome, thanks :)  hope you had a good time!

PR #552, opened 2015-05-24
Add a bunch of capi and fix a bug inside PyNumber_Int (was throwing a c++ exception)

undingen:


kmod:
hmm I wonder if there's some easy way for us to mass-test our capi code.  this pr seems pretty straightforward since most of it seems copied from CPython, but I'm always uneasy when we have code going in with no test coverage :/

PR #556, opened 2015-05-26
add libunwind_patch for binary search

toshok:
This patch adds binary searching of dynamically registered objects (same as the libgcc hack from before).

before the patch pyston_release does django-template in ~14 seconds
after the patch it takes ~10 seconds.


kmod:
nice :)

PR #559, opened 2015-05-27
switch to CPython's thread._local implementation

kmod:
Gives us some nice features, but had some more dependencies than I anticipated:
- setting object.tp_init properly (cpython wants to check `cls->tp_init == object_cls->tp_init`)
- supporting weakrefs on extension types

### DIFF
--- a/from_cpython/CMakeLists.txt
+++ b/from_cpython/CMakeLists.txt
@@ -64,6 +64,7 @@ file(GLOB_RECURSE STDMODULE_SRCS Modules
     stringio.c
     stropmodule.c
     textio.c
+    threadmodule.c
     timemodule.c
     unicodedata.c
     util.c
--- a/from_cpython/Include/methodobject.h
+++ b/from_cpython/Include/methodobject.h
@@ -91,15 +91,12 @@ typedef struct PyMethodChain {
 PyAPI_FUNC(PyObject *) Py_FindMethodInChain(PyMethodChain *, PyObject *,
                                             const char *) PYSTON_NOEXCEPT;
 
-// Pyston change: not our format
-#if 0
 typedef struct {
     PyObject_HEAD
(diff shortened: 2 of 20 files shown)


kmod:
Hmm it looks like there's a slight perf regression here which I'll look into (I think from the object_init / object_new changes), but if you could take a look I think it's worth merging as-is.  (I fixed the ci failure)

```
              pyston django_migrate.py                  :    2.3s baseline: 2.3 (-0.8%)
              pyston django-template.py                 :   15.4s baseline: 15.4 (-0.0%)
              pyston interp2.py                         :    6.3s baseline: 6.1 (+3.2%)
              pyston raytrace.py                        :    6.3s baseline: 6.2 (+0.9%)
              pyston nbody.py                           :    8.2s baseline: 8.3 (-0.6%)
              pyston fannkuch.py                        :    7.5s baseline: 7.3 (+2.2%)
              pyston chaos.py                           :   20.3s baseline: 20.5 (-0.6%)
              pyston fasta.py                           :    5.5s baseline: 5.3 (+2.8%)
              pyston pidigits.py                        :    5.7s baseline: 5.6 (+1.0%)
              pyston richards.py                        :    2.7s baseline: 2.7 (-0.5%)
              pyston deltablue.py                       :    1.8s baseline: 1.8 (+2.5%)
              pyston (geomean-3424)                     :    5.8s baseline: 5.8 (+0.9%)
```

toshok:
yeah, presumably the object_init/object_new change means we'll inline less / use a less optimal ic since we're using a slot?

lgtm though :+1:

kmod:
Well direct calls to `__init__` and `__new__` would definitely be slower, but I think the majority of them happen from `typeCallInternal()`, which has some additional optimizations it tries to apply so I'm not 100% sure how the perf gets affected.  Some of the benchmarks got (reliably) faster...

PR #560, opened 2015-05-27
Add PyErr_GetExcInfo, make __builtins__ more similar to cpython and id() output more useful

undingen:
- changed the ```id()``` builtin to directly output the pointer value. I find this useful for debugging.
- Added ```PyErr_GetExcInfo``` and ```PyErr_SetExcInfo```
 - they only got added in CPython >= 3.3 but Cython uses them in non cpython mode and they are useful.
 - they are used to set/retrieve ```sys.exc_info```
- fixed ```__builtins__["attr"]``` usage inside imported modules. (The become dicts not modules - that's why this syntax works too)
- fixed ```PyObject_GetAttr``` with unicode attribute name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,6 +344,7 @@ STDMODULE_SRCS := \
 	signalmodule.c \
 	selectmodule.c \
 	fcntlmodule.c \
+	threadmodule.c \
 	timemodule.c \
 	arraymodule.c \
 	zlibmodule.c \
--- a/from_cpython/Include/pyerrors.h
+++ b/from_cpython/Include/pyerrors.h
@@ -83,6 +83,10 @@ PyAPI_FUNC(void) PyErr_Clear(void) PYSTON_NOEXCEPT;
 PyAPI_FUNC(void) PyErr_Fetch(PyObject **, PyObject **, PyObject **) PYSTON_NOEXCEPT;
 PyAPI_FUNC(void) PyErr_Restore(PyObject *, PyObject *, PyObject *) PYSTON_NOEXCEPT;
 
+// Pyton change: This functions are normally only available in CPython >= 3.3 and PyPy but Cython can use them.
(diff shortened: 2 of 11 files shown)


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -232,8 +232,7 @@ extern "C" Box* sum(Box* container, Box* initial) {
 }
 
 extern "C" Box* id(Box* arg) {
-    i64 addr = (i64)(arg) ^ 0xdeadbeef00000003;
-    return boxInt(addr);
+    return PyLong_FromVoidPtr(arg);
```
kmod: hmm so the reasoning behind xoring is to not confuse the conservative collector.  I would guess/hope that people don't keep id's around and especially not for objects that are dead, but it's nice to return a value that doesn't point into our heap.  Would it help if we xor'd it with a nicer value, such as `0x1000000000000000`?

You could also try out our `dump()` Python builtin:
```
>> dump(__builtins__)
Raw address: 0x12700282a8
Python object
Class: module
The '__builtin__' module
```

undingen: Oh makes sense, I will revert it and just use ```dump``` the next time.

undingen:
- removed the ```id```change
- fixed gnu make build (not that I'm using it.... no.... one shouldn't use it after the new libunwind changes... :-D)
- set ```__doc__``` string inside exec statements: needed for ```oauth2client```

undingen:
Also added a basic import test for decorator and oauth2client.
We may want to test a specific version but I'm not sure which...

kmod:
I would just see whatever version is latest on pypi (ie whatever the test currently downloads) and then put that into the test.  ie I don't think we care about any particular revision, but it's nice to not have our tests change when a new version of the libraries get upload.d

PR #561, opened 2015-05-27
Call finalizers during garbage collection with ordering

rudi-c:
This PR implements finalizers, using PyPy's finalization ordering algorithm.

Removes the need for some workaround hacks like simple_destructors.

There's more optimization work to be done but this is functional and is getting quite big, so I figured this would be a good time for a code review.


src/gc/collector.cpp
```diff
@@ -273,8 +371,8 @@ void markPhase() {
     VALGRIND_DISABLE_ERROR_REPORTING;
 #endif
 
-    TraceStack stack(roots);
-    GCVisitor visitor(&stack);
+    shared_ptr<MarkStack> stack = make_shared<MarkStack>(roots);
```
toshok: missing `std::`'s here.  we should probably not have the `using namespace std` at the beginning of the file.

src/gc/collector.cpp
```diff
@@ -342,26 +396,121 @@ void markPhase() {
 #endif
 }
 
+void finalizationOrderingFirstPass(Box* obj) {
+    shared_ptr<FirstPhaseStack> stack = make_shared<FirstPhaseStack>();
```
toshok: same here with `std::`

src/gc/heap.cpp
```diff
@@ -137,6 +139,68 @@ void registerGCManagedBytes(size_t bytes) {
...
+void finalizeIfUnordered(Box* b) {
+    if (!hasOrderedFinalizer(b)) {
+        GCAllocation* al = GCAllocation::fromUserData(b);
+        if (!hasFinalized(al)) {
+            finalize(b);
+        }
+    }
+}
+
+bool isWeakReference(Box* b) {
```
toshok: this should probably be called `isWeaklyReferenced`.  `isWeakReference` sounds like we're testing the box to see if it's an instance of a WeakRef type.

src/gc/collector.cpp
```diff
@@ -342,26 +416,121 @@ void markPhase() {
...
+
+    for (Box* marked : finalizer_marked) {
+        GCAllocation* al = GCAllocation::fromUserData(marked);
+
+        FinalizationState state = orderingState(al);
+        assert(state == FinalizationState::REACHABLE_FROM_FINALIZER || state == FinalizationState::ALIVE);
+
+        if (state == FinalizationState::REACHABLE_FROM_FINALIZER) {
+            assert(roots.count(marked) == 0);
+            registerPermanentRoot(marked);
```
toshok: I wouldn't use `registerPermanentRoot`/`deregisterPermanentRoot` for the objects in the `pending_finalization_list`.  instead just add a pass iterating the list during the markPhase.

src/gc/collector.cpp
```diff
@@ -273,8 +391,8 @@ void markPhase() {
     VALGRIND_DISABLE_ERROR_REPORTING;
 #endif
 
-    TraceStack stack(roots);
-    GCVisitor visitor(&stack);
+    std::shared_ptr<MarkStack> stack = std::make_shared<MarkStack>(roots);
```
toshok: I don't think there's a good reason to use shared_ptr's here - the lifetime of the visitor should be bounded by that of the stack in all cases (and should be easily verified.)

src/gc/collector.cpp
```diff
@@ -342,26 +416,121 @@ void markPhase() {
...
+        visitByGCKind(p, visitor);
+    }
+}
+
+// Implementation of PyPy's finalization ordering algorithm:
+// http://pypy.readthedocs.org/en/latest/discussion/finalizer-order.html
+void finalizationOrderingPhase() {
+    // TODO: Replace with iterator or something that takes a lambda for memory efficiency?.
+    std::vector<Box*> objects_with_finalizers;
+    global_heap.getOrderedFinalizers(objects_with_finalizers);
```
toshok: `getOrderedFinalizers` does another full walk of the heap (basically another sweep phase).  Is there a way we can add the `objects_with_finalizers` creation to the sweep phase?

src/gc/collector.cpp
```diff
@@ -82,20 +107,18 @@ class TraceStack {
...
-        for (void* p : rhs) {
-            assert(!isMarked(GCAllocation::fromUserData(p)));
-            push(p);
-        }
-    }
+    virtual ~TraceStack() {}
+
+    // Returns false if we should skip visiting the children.
+    // true otherwise.
+    virtual bool visit_action(GCAllocation* al) = 0;
```
toshok: the `virtual` makes me nervous, considering how stupidly hot the trace stack stuff is.  We can defer it until after the `objects_with_finalizers` pass is merged into the sweep phase, but this might benefit from refactoring the TraceStack and using templates + functors to guarantee everything gets inlined.

src/gc/collector.cpp
rudi-c: Ah yes that'd be a lot more efficient.

src/gc/collector.cpp
rudi-c: unique_ptr?

src/gc/collector.cpp
rudi-c: I don't think so, because unreachable objects with finalizers can mark other unreachable objects that it references. I think we'd need to consider other objects like tracking finalizable objects as they get created.

src/gc/heap.cpp
```diff
@@ -137,6 +139,66 @@ void registerGCManagedBytes(size_t bytes) {
     }
 }
 
+//////
+/// Finalizers
+
+bool hasFinalizer(Box* b) {
```
toshok: i think all of these can be `static`

src/gc/collector.cpp
toshok: sorry, I edited my comment shortly after posting and don't know which you were responding to.  It originally suggested building up `objects_with_finalizers` during the sweep phase.  Was your response about that comment or the one about doing it when we allocate objects?

rudi-c: I replied to the original comment.

If we build `objects_with_finalizers` during allocation, `__del__` methods added after the fact will be a problem. I'm not sure if that'd be hard, or if it'd even be an edge case worth handling?

toshok: alas, this prints "hello world":
```
class WithoutDelAtStart(object):
    pass

def delobj(self):
    print "hello world"

WithoutDelAtStart.__del__ = delobj

f = WithoutDelAtStart()
f = None
```

src/gc/collector.cpp
```diff
@@ -116,16 +139,86 @@ class TraceStack {
...
+        if (isMarked(al)) {
+            return false;
+        } else {
+            setMark(al);
+            return true;
+        }
+    }
+};
+
+class FirstPhaseStack : public TraceStack {
```
toshok: bikeshedding here, but FirstPhase/SecondPhase aren't very descriptive (same with finalizationOrderingFirstPass/finalizationOrderingSecondPass).  I don't really have any good suggestions, though.

src/gc/collector.cpp
toshok: unique_ptr would work, I think clang -O3 can inline all of it.

src/gc/collector.cpp
```diff
@@ -116,16 +139,86 @@ class TraceStack {
...
+        for (void* p : root_handles) {
+            assert(!isMarked(GCAllocation::fromUserData(p)));
+            push(p);
+        }
+    }
+
+    virtual bool visit_action(GCAllocation* al) {
+// Use this to print the directed edges of the GC graph traversal.
+// i.e. print every a > b where a is a pointer and b is something a references
+#if 0
```
toshok: could this be enabled, but in a separate subclass/instantiation of MarkStack?  I'd very much like `pyston::gc::whatReferences(void* p)` that I can use in gdb.

toshok:
this is getting close imo.  we do need some idea of what the perf hit will be across the board.  Also, I didn't see any tests that test finalization of C extension types.  popen, file, and socket are probably good places to start.

rudi-c:
So there's a couple more things that could be done, which I'm thinking of leaving for another PR:
1) Figure out a way to use templates for TraceStack which would remove a layer of indirection.
2) A helper version of MarkStack that has a `whatReferences` method.
3) More tests for built-ins like popen, file, and socket.
4) More performance work. There are still some tests that are currently suffering a massive hit.

```
pyston (calibration)                      :    1.0s baseline: 1.0 (-2.0%)  
pyston django_template.py                 :    6.3s baseline: 5.5 (+14.4%)               
pyston django_migrate.py                  :    1.9s baseline: 1.8 (+4.7%)            
pyston interp2.py                         :    4.2s baseline: 4.1 (+3.4%)  
pyston raytrace.py                        :    6.8s baseline: 5.8 (+17.1%) 
pyston nbody.py                           :   10.1s baseline: 6.8 (+47.9%) 
pyston fannkuch.py                        :    7.5s baseline: 6.3 (+18.6%) 
pyston chaos.py                           :   21.2s baseline: 18.4 (+15.5%)
pyston fasta.py                           :    5.8s baseline: 4.3 (+34.1%) 
pyston pidigits.py                        :    6.4s baseline: 5.7 (+12.4%) 
pyston richards.py                        :    1.5s baseline: 1.5 (+2.2%)  
pyston deltablue.py                       :    1.6s baseline: 1.5 (+1.4%)
```

rudi-c:
Inlining one function helped get the % down quite a bit for django-template.
```
pyston (calibration)                      :    1.0s baseline: 1.0 (-0.0%)
pyston django_template.py                 :    5.8s baseline: 5.5 (+5.8%)
pyston django_migrate.py                  :    1.9s baseline: 1.8 (+2.3%)
pyston interp2.py                         :    4.5s baseline: 4.0 (+11.8%)
pyston raytrace.py                        :    6.4s baseline: 5.5 (+16.2%)
pyston nbody.py                           :    9.6s baseline: 6.8 (+40.4%)
pyston fannkuch.py                        :    7.4s baseline: 6.4 (+15.6%)
pyston chaos.py                           :   20.3s baseline: 17.7 (+14.6%)
pyston fasta.py                           :    5.9s baseline: 4.2 (+40.9%)
pyston pidigits.py                        :    6.3s baseline: 5.5 (+14.4%)
pyston richards.py                        :    1.5s baseline: 1.5 (+1.3%)
pyston deltablue.py                       :    1.5s baseline: 1.4 (+2.7%)
```

rudi-c:
Removing the extra sweep phase (for most collections). Now most tests aren't seeing too big of a slowdown.
```
pyston (calibration)                      :    1.0s baseline: 1.0 (-0.0%)
pyston django_template.py                 :    5.3s baseline: 5.1 (+5.3%)
pyston django_migrate.py                  :    1.8s baseline: 1.7 (+4.9%)
pyston interp2.py                         :    4.5s baseline: 4.0 (+13.8%)
pyston raytrace.py                        :    6.0s baseline: 5.3 (+13.0%)
pyston nbody.py                           :    8.0s baseline: 7.1 (+12.3%)
pyston fannkuch.py                        :    6.6s baseline: 6.4 (+2.2%)
pyston chaos.py                           :   18.9s baseline: 17.1 (+10.8%)
pyston fasta.py                           :    4.5s baseline: 4.1 (+10.1%)
pyston pidigits.py                        :    5.7s baseline: 5.5 (+4.6%)
pyston richards.py                        :    1.5s baseline: 1.4 (+3.1%)
pyston deltablue.py                       :    1.4s baseline: 1.4 (+4.3%)
```

src/gc/collector.cpp
```diff
@@ -297,73 +565,24 @@ void markPhase() {
...
-    GC_TRACE_LOG("Looking at root handles\n");
-    for (auto h : *getRootHandles()) {
-        visitor.visit(h->value);
-    }
-
-    GC_TRACE_LOG("Looking at potential root ranges\n");
-    for (auto& e : potential_root_ranges) {
-        visitor.visitPotentialRange((void* const*)e.first, (void* const*)e.second);
+    if (also_free_classes) {
+        // If we allow freeing classes, we must make sure that even for unreachable objects,
```
toshok: I still think there's a way to do this by ordering things during the sweep phase.  While sweeping we can check if the object we're about to free is a class and add it to a list, then when we're done freeing instances, we do a walk of that list freeing the classes.

If the hit of doing the additional linear walk of just the classes is still really slow, we can do it every few collections, but it would be better to base the decision to collect classes on collection frequency instead of number of classes allocated, since the former goes up with heap occupancy/pressure and the latter doesn't -- i.e. I allocate ALLOC_CLASSES_PER_CLASS_COLLECTION-1 classes, only keep a reference to 1 of them, then fill up the heap over and over.

rudi-c:
After rebasing.
```
pyston django_template.py                 :    4.7s baseline: 4.6 (+3.0%)  
pyston django_migrate.py                  :    1.8s baseline: 1.7 (+0.9%)  
pyston interp2.py                         :    4.1s baseline: 4.4 (-6.7%)  
pyston raytrace.py                        :    5.7s baseline: 5.7 (+0.4%)  
pyston nbody.py                           :    7.1s baseline: 6.8 (+3.3%)  
pyston fannkuch.py                        :    6.4s baseline: 6.3 (+1.4%)  
pyston chaos.py                           :   18.7s baseline: 16.8 (+11.3%)
pyston fasta.py                           :    4.4s baseline: 4.3 (+2.6%)  
pyston pidigits.py                        :    5.8s baseline: 5.5 (+5.1%)  
pyston richards.py                        :    1.5s baseline: 1.5 (+0.1%)  
pyston deltablue.py                       :    1.4s baseline: 1.3 (+4.1%)  
```

kmod:
I wouldn't worry too much about the microbenchmarks -- I would pull pyston-perf, and then only look at the first five benchmarks (or use investigate.py).  The older benchmarks can be interesting if they expose certain issues with a change, but for the most part they're just so small that unrelated changes can have huge effects.  Check out [the recent history for interp2.py](http://speed.pyston.org/timeline/#/?exe=1,3&base=none&ben=interp2&env=2&revs=50&equid=on) to get a sense of what I mean -- the sawtooth pattern seems like more of a testing limitation than an actual representation of our performance.

rudi-c:
Will split into separate PRs.

PR #562, opened 2015-05-28
Convert runtime functions to take llvm::StringRef

kmod:
```
   pyston django_migrate.py                  :    2.3s baseline: 2.3 (-1.7%)
   pyston django-template.py                 :   15.1s baseline: 15.4 (-1.6%)
   pyston interp2.py                         :    5.3s baseline: 6.3 (-15.1%)
   pyston raytrace.py                        :    6.1s baseline: 6.2 (-0.7%)
   pyston nbody.py                           :    8.4s baseline: 8.1 (+4.1%)
   pyston fannkuch.py                        :    7.5s baseline: 7.5 (+0.2%)
   pyston chaos.py                           :   20.2s baseline: 20.0 (+0.7%)
   pyston fasta.py                           :    5.4s baseline: 5.4 (+0.3%)
   pyston pidigits.py                        :    5.7s baseline: 5.7 (+0.0%)
   pyston richards.py                        :    2.5s baseline: 2.7 (-6.2%)
   pyston deltablue.py                       :    1.8s baseline: 1.8 (-0.0%)
   pyston (geomean-3424)                     :    5.7s baseline: 5.8 (-2.0%)
```

The nbody regression seems to just be benchmark sensitivity -- as far as I can tell it comes from a single function `unpackIntoArray` that gets called the same number of times and has the same assembly.  I'm going to work on getting a virtualenv test in and maybe we should just remove the old benchmarks.

### DIFF
--- a/from_cpython/Objects/floatobject.c
+++ b/from_cpython/Objects/floatobject.c
@@ -412,6 +412,8 @@ float_str(PyFloatObject *v)
  * coercion to double.  So this part is painful too.
  */
 
+// Pyston change: don't need this for now
+#if 0
 static PyObject*
 float_richcompare(PyObject *v, PyObject *w, int op)
 {
@@ -625,6 +627,7 @@ float_richcompare(PyObject *v, PyObject *w, int op)
     Py_INCREF(Py_NotImplemented);
     return Py_NotImplemented;
 }
+#endif
 
 static long
 float_hash(PyFloatObject *v)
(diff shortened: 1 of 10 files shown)


undingen:
Nice change :-), often wanted to do it myself but was always to lazy :-D.
Some of the temporary std::string could now become normal string literals or do you want to keep them?

toshok:
wow, nice change, especially for interp2.  I have also been thinking about
starting on this and was going to send mail.  Glad I found this instead :)

kmod:
Looks like a ~2% improvement to the new virtualenv benchmark as well, which is kind of surprising to me.

PR #564, opened 2015-05-29
Incremental traceback

toshok:
This works, and results in a pretty decent speedup for django-template.py, from ~10s to ~8s.

opening a PR to let travis-ci do the slow tests.  need to clean up the changes and figure out a better way to deal with the state I'm carrying around during unwinding.

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -632,19 +632,8 @@ Value ASTInterpreter::visit_invoke(AST_Invoke* node) {
         next_block = node->normal_dest;
     } catch (ExcInfo e) {
 
-        if (threading::ThreadStateInternal::getUnwindState() == UNWIND_STATE_NORMAL) {
-            // when generating the traceback incrementally we only
-            // include an interpreter frame if we unwind through
-            // ASTInterpreter::execute_inner.  this will keep a toplevel
-            // invoke from showing up, since we catch the exception
-            // here.
-
-            auto source = getCF()->clfunc->source.get();
-            BoxedTraceback::Here(LineInfo(node->lineno, node->col_offset, source->fn, source->getName()),
-                                 reinterpret_cast<BoxedTraceback**>(&e.traceback));
-        }
-
-        threading::ThreadStateInternal::setUnwindState(UNWIND_STATE_NORMAL);
+        auto source = getCF()->clfunc->source.get();
+        exceptionCaughtInInterpreter(LineInfo(node->lineno, node->col_offset, source->fn, source->getName()), &e);
 
         next_block = node->exc_dest;
         last_exception = e;
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -41,7 +41,6 @@
 #include "runtime/long.h"
 #include "runtime/objmodel.h"
 #include "runtime/set.h"
-#include "runtime/traceback.h"
 #include "runtime/types.h"
 
 #ifndef NDEBUG
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -318,7 +318,8 @@ void ASTInterpreter::initArguments(int nargs, BoxedClosure* _closure, BoxedGener
 static std::unordered_map<void*, ASTInterpreter*> s_interpreterMap;
 static_assert(THREADING_USE_GIL, "have to make the interpreter map thread safe!");
 
-RegisterHelper::RegisterHelper() : frame_addr(NULL), interpreter(NULL) { }
+RegisterHelper::RegisterHelper() : frame_addr(NULL), interpreter(NULL) {
+}
 
 RegisterHelper::~RegisterHelper() {
     assert(interpreter);
@@ -335,12 +336,12 @@ void RegisterHelper::doRegister(void* frame_addr, ASTInterpreter* interpreter) {
     interpreter->frame_addr = frame_addr;
     s_interpreterMap[frame_addr] = interpreter;
 }
-    
-    void RegisterHelper::deregister(void* frame_addr) {
-        assert(frame_addr);
-        assert(s_interpreterMap.count(frame_addr));
-        s_interpreterMap.erase(frame_addr);
-    }
+
+void RegisterHelper::deregister(void* frame_addr) {
+    assert(frame_addr);
+    assert(s_interpreterMap.count(frame_addr));
+    s_interpreterMap.erase(frame_addr);
+}
 
 Value ASTInterpreter::executeInner(ASTInterpreter& interpreter, CFGBlock* start_block, AST_stmt* start_at,
                                    RegisterHelper* reg) {
@@ -631,7 +632,7 @@ Value ASTInterpreter::visit_invoke(AST_Invoke* node) {
         next_block = node->normal_dest;
     } catch (ExcInfo e) {
 
-        if (cur_thread_state.unwind_state == UNWIND_STATE_NORMAL) {
+        if (threading::ThreadStateInternal::getUnwindState() == UNWIND_STATE_NORMAL) {
             // when generating the traceback incrementally we only
             // include an interpreter frame if we unwind through
             // ASTInterpreter::execute_inner.  this will keep a toplevel
@@ -643,7 +644,7 @@ Value ASTInterpreter::visit_invoke(AST_Invoke* node) {
                                  reinterpret_cast<BoxedTraceback**>(&e.traceback));
         }
 
-        cur_thread_state.unwind_state = UNWIND_STATE_NORMAL;
+        threading::ThreadStateInternal::setUnwindState(UNWIND_STATE_NORMAL);
 
         next_block = node->exc_dest;
         last_exception = e;
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -41,6 +41,7 @@
 #include "runtime/long.h"
 #include "runtime/objmodel.h"
 #include "runtime/set.h"
+#include "runtime/traceback.h"
 #include "runtime/types.h"
 
 #ifndef NDEBUG
@@ -54,29 +55,6 @@ namespace pyston {
 namespace {
 
 static BoxedClass* astinterpreter_cls;
-
-class ASTInterpreter;
-
-// Map from stack frame pointers for frames corresponding to ASTInterpreter::execute() to the ASTInterpreter handling
-// them. Used to look up information about that frame. This is used for getting tracebacks, for CPython introspection
-// (sys._getframe & co), and for GC scanning.
-static std::unordered_map<void*, ASTInterpreter*> s_interpreterMap;
-static_assert(THREADING_USE_GIL, "have to make the interpreter map thread safe!");
-
-class RegisterHelper {
-private:
-    void* frame_addr;
-    ASTInterpreter* interpreter;
-
-public:
-    RegisterHelper(ASTInterpreter* interpreter, void* frame_addr);
-    ~RegisterHelper();
-
-    static void deregister(void* frame_addr) {
-        assert(s_interpreterMap.count(frame_addr));
-        s_interpreterMap.erase(frame_addr);
-    }
-};
 
 union Value {
     bool b;
@@ -93,6 +71,20 @@ union Value {
     }
 };
 
+class ASTInterpreter;
+
+class RegisterHelper {
+private:
+    void* frame_addr;
+    ASTInterpreter* interpreter;
+
+public:
+    RegisterHelper();
+    ~RegisterHelper();
+    void doRegister(void* frame_addr, ASTInterpreter* interpreter);
+    static void deregister(void* frame_addr);
+};
+
 class ASTInterpreter : public Box {
 public:
     typedef ContiguousMap<InternedString, Box*> SymMap;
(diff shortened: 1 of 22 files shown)


toshok:
```
Saving results as 'incremental-tracebacks'
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.2%)
              pyston django_migrate.py                  :    2.0s baseline: 2.2 (-10.7%)
              pyston virtualenv_bench.py                :    5.9s baseline: 7.0 (-15.8%)
              pyston django-template.py                 :    8.0s baseline: 11.5 (-30.6%)
              pyston interp2.py                         :    5.7s baseline: 5.5 (+3.6%)
              pyston raytrace.py                        :    6.3s baseline: 6.1 (+2.5%)
              pyston nbody.py                           :    8.4s baseline: 8.2 (+1.8%)
              pyston fannkuch.py                        :    7.6s baseline: 7.3 (+4.3%)
              pyston chaos.py                           :   20.4s baseline: 20.7 (-1.5%)
              pyston fasta.py                           :    5.4s baseline: 5.4 (-0.8%)
              pyston pidigits.py                        :    5.6s baseline: 5.5 (+1.4%)
              pyston richards.py                        :    2.5s baseline: 2.5 (+0.4%)
              pyston deltablue.py                       :    1.9s baseline: 1.9 (+0.8%)
              pyston (geomean-1a6f)                     :    5.4s baseline: 5.6 (-4.3%)
```

kmod:
Do we need the `incremental_traceback` flag?  I think tracebacks in Python are fundamentally incremental -- here's the struct definition from CPython:

```C
typedef struct _traceback {
	PyObject_HEAD
	struct _traceback *tb_next;
	struct _frame *tb_frame;
	int tb_lasti;
	int tb_lineno;
} PyTracebackObject;
```
ie a traceback object just represents a single frame, and points to the "next" traceback object down (newer on) the call chain.  And I think CPython pushes new traceback objects onto that list no matter what the source of the exception was?


So you can take an existing exception and throw it somewhere else, and any new frames it traverses will get added to it.  Here's a contrived example:

```python
import traceback
import sys
def f():
    a, b, c = sys.exc_info()
    raise a, b, c

try:
    1/0
except:
    pass

for i in xrange(10):
    try:
        f()
    except:
        pass

traceback.print_exc()
```

toshok:
Oh perfect, that's exactly the sort of test I was wondering about.  I wasn't sure about the incremental-ness of cpython exceptions.

adding test, and will remove the flag.

toshok:
so we need to keep 3 pieces of state, in 2 places, to really make this work:

`was_osr`: same as it was before
`was_interpreter`: this is fragile and relies on the fact that `ASTInterpreter::execute` has a `catch`.  we will always process those frames twice.  This flag keeps us from trying to add a traceback line twice (the second time trips asserts).
`BoxedTraceback::reraising`: I originally had this in `ExcInfo`, but it can't live there because of the way `ASTInterpreter::visit_invoke` works - we have a `catch` there as well, which needs to  do effectively `PyTraceBack_Here`, but only if we aren't currently in the reraising state.  reraising is set to false in two places - in `::visit_invoke`, and in the callback function we pass through maybeAddLineToTraceback.

we could possibly also treat ::visit_invoke the way we treat ::execute, by checking the ip to see if it's in the range for ::visit_invoke, and set was_interpreter and skipping the second time processing things in that function.  That might result in slightly cleaner code.

toshok:
interestingly, the traceback object isn't attached to the exception at all in python 2.7, which kinda boggles the mind.

toshok:
@kmod okay, I think this is finally ready.  the type/value/traceback are now traced during gc, and so gc's are permissible during unwinding (and happen during django-template.py runs).

kmod:
I think we can keep things better separated (I'll try to find some concrete suggestions); for example, it seems odd that the interpreter will have to ask the unwinder, via threadstate, if it had just reraised an exception or not.

src/codegen/ast_interpreter.cpp
```diff
@@ -609,6 +630,21 @@ Value ASTInterpreter::visit_invoke(AST_Invoke* node) {
         v = visit_stmt(node->stmt);
         next_block = node->normal_dest;
     } catch (ExcInfo e) {
+
+        if (threading::ThreadStateInternal::getUnwindState() == UNWIND_STATE_NORMAL) {
```
kmod: Would it work to have this check be something like `!isReraise(node->stmt)`?

src/codegen/unwinding.cpp
```diff
@@ -410,23 +417,83 @@ static unw_word_t getFunctionEnd(unw_word_t ip) {
...
+            int code = unw_get_reg(cursor, i, &r);
+            ASSERT(code == 0 || code == -UNW_EBADREG, "%d %d", code, i);
+            if (code == 0) {
+                info.regs[i] = r;
+                info.regs_valid |= (1 << i);
+            }
+        }
+    }
+
+    if (threading::ThreadStateInternal::getUnwindState() == UNWIND_STATE_NORMAL) {
```
kmod: It feels weird that we combine the logic of how to process the frame with the logic of when to process the frame.  I think it might make more sense to have the caller of unwindProcessFrame know whether to actually process the frame or not, since those rules will be different based on the caller.  For example, `sys._getframe` uses this same machinery to traverse the Python stack, but that traversal will now include the rules to skip the topmost frame if we are handling a reraise.  As far as I can tell, we can never call `sys._getframe` when we are in `UNWIND_STATE_RERAISE` so the extra check *should* be ok, but it seems wrong that we check for reraises when doing `sys._getframe`.

toshok: I've moved the code for getting/setting the unwind state out of `unwindProcessFrame`, and while it still works (and is better from a separation of concerns standpoint), it's a lot of c&p code doing the same checks.  I came up with the following higher order template func which seems to help.  It's only used in unwindPythonStack and maybeTracebackHere

https://gist.github.com/toshok/abf78446ce2adb59782d

that said, the checks are still performed since we go through unwindPythonStack.  It should be impossible to be in the RERAISE state, though, since the first thing unwindPythonStack does is set unwind_state to NORMAL.

toshok: I also think that combining OSR and RERAISE into a single state - SKIP_NEXT - might help things a bit.  Basically any time unwindPythonStack is called we start in NORMAL.  the only way we can start SKIP_NEXT if is someone does `raise0` or passes a traceback to `raise3.`  SKIP_NEXT would be a valid state to be in after the first python frame has been seen, though.

src/codegen/ast_interpreter.cpp
toshok: It does appear to work, using this (which I'm not sure about):

```
static bool isReraise(AST_stmt* stmt) {                                                                                        
    if (stmt->type != AST_TYPE::Raise) return false;                                                                           
    AST_Raise* raise = static_cast<AST_Raise*>(stmt);                                                                          
    return raise->arg0 == NULL  // the 0-arg form                                                                              
        || raise->arg2 != NULL; // the 3-arg form                                                                              
}                                                                                                                              
```

are invoke statements limited to raises?

toshok: to answer my own question, no they aren't.  but using this form seems to work.  I dislike the idea slightly of having multiple ways to check if a raise was a re-raise depending on what context we're in, but this particular instance doesn't really pain me too much, as the only way for there to be a re-raise is for the `node->stmt` to have caused it.

toshok: alas, this doesn't work.  passing `None` as the third arg (as in  `raise exc, tp, None`) should not be treated as a re-raise, but we can't determine that by looking at the AST_stmt.

src/codegen/unwinding.cpp
toshok: I guess my point with both of the above comments is that `sys._getframe` has to deal with osr frames, and thus has to deal with the code to skip over a frame.  The only thing `sys._getframe` shouldn't have to deal with is the re-raise case.

By combining OSR and RERAISE into one state (since they both have the same meaning), we can remove the distasteful aspect of having `sys._getframe` care about re-raising.  It only cares about the possibility that the unwinding has to skip a frame (something it already had to deal with.)

This lets us reframe things:  The requirement is that `sys._getframe` shouldn't ever be used in a state where the *initial* frame is skipped.  That is already guaranteed by `unwindPythonStack` setting the state to `NORMAL` before walking the stack.  All of the entry points to unwinding that deal with introspection from python go through `unwindPythonStack`.

there are 4 places where we can begin unwinding the python stack, 3 of which originate in c++ unwinding.  we initialize the state in each of those places:

1. `raise0` - always sets the state to RERAISE/SKIPNEXT
2. `raise3` - sets the state to NORMAL or RERAISE/SKIPNEXT depending on arg3
3. `raiseExc` - always set the state to NORMAL (I actually need to add this)
3. `unwindPythonStack` - always sets the state to NORMAL


---

we can move the code to deal with maintaining the unwind state into the callers.  That simplifies unwindProcessFrame (now it will just create a PythonFrameIteratorImpl if it can, and call the function passed), and I think it's a good idea.  It just ends up duplicating the same state handling code in every use of `unwindPythonStack` except one: `getTopPythonFrame`.   The gist was just a way to move that code out that also won't c&p the same code everywhere.

toshok:
any further comments?  I'm thinking we just need to go with this.  If there is further decoupling work for what marius might need, we can address that as it comes up.  I have some ideas (and even a little code), but don't really want to commit to an interface until we know more about what's needed (we may not ever need it)

toshok:
current perf against  master:

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.1%) 
              pyston django_migrate.py                  :    1.8s baseline: 2.1 (-14.2%) 
              pyston virtualenv_bench.py                :    5.4s baseline: 6.2 (-13.5%) 
              pyston django-template.py                 :    5.9s baseline: 7.5 (-21.2%) 
              pyston interp2.py                         :    4.3s baseline: 4.6 (-5.6%) 
              pyston raytrace.py                        :    5.8s baseline: 6.1 (-4.9%) 
              pyston nbody.py                           :    8.5s baseline: 8.7 (-2.0%) 
              pyston fannkuch.py                        :    6.9s baseline: 7.0 (-1.3%) 
              pyston chaos.py                           :   19.6s baseline: 19.2 (+2.2%) 
              pyston fasta.py                           :    4.9s baseline: 4.9 (+0.6%) 
              pyston pidigits.py                        :    5.9s baseline: 5.7 (+2.5%) 
              pyston richards.py                        :    2.4s baseline: 2.4 (-1.2%) 
              pyston deltablue.py                       :    1.6s baseline: 1.7 (-2.6%) 
              pyston (geomean-1a6f)                     :    4.9s baseline: 5.1 (-5.4%bration)
```

tjhance:
Would you mind summarizing what this does, just to satisfy my curiosity?

toshok:
the current process of raising exceptions calls getTraceback(), which unwinds the entire python stack, adding line entries for every frame.  Not only is this incorrect (the traceback is only supposed to have entries from the point of raising to the point of catching), it's quite slow, since we have to unwind twice, once to generate the traceback object, then again to do the c++ unwinding.

this patch builds up the traceback as we do the c++ unwind (there's a call in cxx_unwind.cpp - `maybeTracebackHere(&cursor);` that allows the python unwinding code to check if the unwind cursor corresponds to a python frame and allocate and add the traceback for that frame.) so we generate the correct traceback in a little more than half the time.

tjhance:
gotcha

toshok:
the diff is getting pretty hard to manage.  will likely remove the changes I made to `ThreadStateInternal` and squash it down, but the current api I have for python unwinding is:

```
class UnwindSession;
UnwindSession* beginUnwind();
UnwindSession* getUnwind();
void endUnwind(UnwindSession* unwind_session);
void* getExceptionStorage(UnwindSession* unwind_session);
void unwindingThroughFrame(UnwindSession* unwind_session, unw_cursor_t* cursor);

void exceptionCaughtInInterpreter(LineInfo line_info, ExcInfo* exc_info);
```

everything except `exceptionCaughtInInterpreter` is called by the c++ unwinding machinery.

src/runtime/traceback.h
```diff
@@ -27,19 +27,25 @@ class GCVisitor;
 extern "C" BoxedClass* traceback_cls;
 class BoxedTraceback : public Box {
 public:
-    std::vector<const LineInfo*> lines;
+    typedef llvm::SmallVector<LineInfo, 1> LinesVector;
+    Box* tb_next;
+    LinesVector lines;
```
kmod: When will there be more than one line for a traceback object?

toshok: at this point only when getTraceback() is called.  an option is to just
loop there creating multiple traceback objects.

On Wed, Jun 17, 2015 at 6:34 PM, Kevin Modzelewski <notifications@github.com

src/runtime/traceback.h
```diff
@@ -27,19 +27,25 @@ class GCVisitor;
...
 
-    void addLine(const LineInfo* line);
+    void addLine(const LineInfo line);
 
     static Box* getLines(Box* b);
 
     static void gcHandler(gc::GCVisitor* v, Box* b);
+
+    // somewhat equivalent to PyTraceBack_Here
+    static void Here(LineInfo lineInfo, BoxedTraceback** tb);
```
kmod: minor nit but should we call this "here"?

kmod:
Cool, a pretty nice speedup :)

PR #566, opened 2015-05-29
Python-level sampling profiler

kmod:
Super simple and not very reliable, but can give some decent high-level insight into what's going on.  Specifically, it assigns all the time to the *next* python statement, which makes it a bit hard to interpret (fixing this probably involves figuring out how to handle signals for real).  I wrote a little script to collate the results with different levels of granularity, and interestingly the file-level granularity seems pretty useful.

For example, here's some of the output for django-template:

```
Hiding 117 entries that occurred 392 (12.8%) times

Occurs 81 (2.6%) times:
minibenchmarks/../test/integration/django/django/template/defaulttags.py

Occurs 88 (2.9%) times:
minibenchmarks/../test/integration/django/django/template/smartif.py

Occurs 122 (4.0%) times:
/home/kmod/pyston-build-release/from_cpython/Lib/sre_parse.py

Occurs 167 (5.5%) times:
/home/kmod/pyston-build-release/from_cpython/Lib/sre_compile.py

Occurs 2211 (72.2%) times:
minibenchmarks/../test/integration/django/django/template/base.py

Total tracebacks: 3061
```

Diving deeper shows some of the wonkiness -- there are a bunch of "line -1" entries in there, and I suspect that the `if nodelist` entry is really due to the statement before it `nodelist = getattr(self, attr, None)`
```
Hiding 406 entries that occurred 1311 (42.8%) times

Occurs 75 (2.5%) times:
    return TranslateNode(parser.compile_filter(value), noop, asvar,
  File "minibenchmarks/../test/integration/django/django/templatetags/i18n.py", line 82, in __init__:
    self.filter_expression.var = Variable("'%s'" %
  File "minibenchmarks/../test/integration/django/django/template/base.py", line -1, in __init__:

Occurs 136 (4.4%) times:
    return FilterExpression(token, self)
  File "minibenchmarks/../test/integration/django/django/template/base.py", line 566, in __init__:
    var_obj = Variable(constant).resolve({})
  File "minibenchmarks/../test/integration/django/django/template/base.py", line -1, in __init__:

Occurs 580 (18.9%) times:
    return FilterExpression(token, self)
  File "minibenchmarks/../test/integration/django/django/template/base.py", line 573, in __init__:
    var_obj = Variable(var)
  File "minibenchmarks/../test/integration/django/django/template/base.py", line -1, in __init__:

Occurs 959 (31.3%) times:
  File "minibenchmarks/../test/integration/django/django/template/base.py", line 854, in get_nodes_by_type:
    nodes.extend(node.get_nodes_by_type(nodetype))
  File "minibenchmarks/../test/integration/django/django/template/base.py", line 830, in get_nodes_by_type:
    if nodelist:

Total tracebacks: 3061
```

But here's something pretty interesting -- our startup time (ie setting the number of iterations to 0) on that benchmark is quite bad, and it appears to be regex-related:
```
Hiding 108 entries that occurred 193 (37.1%) times

Occurs 8 (1.5%) times:
minibenchmarks/../test/integration/django/django/db/models/signals.py

Occurs 12 (2.3%) times:
/home/kmod/pyston-build-release/from_cpython/Lib/abc.py

Occurs 12 (2.3%) times:
minibenchmarks/../test/integration/django/django/utils/functional.py

Occurs 123 (23.7%) times:
/home/kmod/pyston-build-release/from_cpython/Lib/sre_parse.py

Occurs 172 (33.1%) times:
/home/kmod/pyston-build-release/from_cpython/Lib/sre_compile.py

Total tracebacks: 520
```

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -695,6 +695,10 @@ Value ASTInterpreter::visit_yield(AST_Yield* node) {
 }
 
 Value ASTInterpreter::visit_stmt(AST_stmt* node) {
+#if ENABLE_SAMPLING_PROFILER
+    threading::allowGLReadPreemption();
+#endif
+
     if (0) {
         printf("%20s % 2d ", source_info->getName().c_str(), current_block->idx);
         print_ast(node);
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -300,8 +300,6 @@ public:
 };
 
 Value ASTInterpreter::execute(ASTInterpreter& interpreter, CFGBlock* start_block, AST_stmt* start_at) {
-    threading::allowGLReadPreemption();
-
     STAT_TIMER(t0, "us_timer_astinterpreter_execute");
 
     void* frame_addr = __builtin_frame_address(0);
@@ -314,6 +312,11 @@ Value ASTInterpreter::execute(ASTInterpreter& interpreter, CFGBlock* start_block
         start_block = interpreter.source_info->cfg->getStartingBlock();
         start_at = start_block->body[0];
     }
+
+    // Important that this happens after RegisterHelper:
+    interpreter.current_inst = start_at;
+    threading::allowGLReadPreemption();
+    interpreter.current_inst = NULL;
 
     interpreter.current_block = start_block;
     bool started = false;
--- a/src/codegen/codegen.h
+++ b/src/codegen/codegen.h
@@ -87,6 +87,8 @@ extern GlobalState g;
 // in runtime_hooks.cpp:
 void initGlobalFuncs(GlobalState& g);
 
+extern int sigprof_pending;
+
 DS_DECLARE_RWLOCK(codegen_rwlock);
 }
 
--- a/src/codegen/entry.cpp
+++ b/src/codegen/entry.cpp
@@ -55,6 +55,8 @@
 
 namespace pyston {
 
+int sigprof_pending = 0;
+
 GlobalState g;
 
 extern "C" {
@@ -356,6 +358,10 @@ static void handle_sigusr1(int signum) {
     _printStacktrace();
 }
 
+static void handle_sigprof(int signum) {
+    sigprof_pending++;
+}
+
 static void handle_sigint(int signum) {
     assert(signum == SIGINT);
     // TODO: this should set a flag saying a KeyboardInterrupt is pending.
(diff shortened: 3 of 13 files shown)


toshok:
Is the signal safety issue the fact that we can't right now unwind safely from/through signal frames?  I can take a look at that once the incremental traceback stuff is done.

Also, none of the changes look particularly performance impacting - any chance we can eschew the #ifdef and just have a command line option before it's merged?

kmod:
Yeah this could probably be made to work well enough that it could be a command line option rather than an #ifdef, but the approach is so temporary (will change completely if we can add real signals support) that I didn't want to spend the time validating that it would be ok :/  There are some things that I'm not as sure would be ok, such as adding additional safepoints after every statement (as opposed to the default of backedges and function entries).

there were two issues I ran into with the signal safety stuff; there might have been more but I gave up once I ran into the second one:
- if we receive the signal while in malloc, we basically can't do anything since all of our code allocates memory.  I was able to work around this by overriding malloc, and if a signal happens during malloc we do the stacktrace at the end of our custom malloc :P
- we only have the ability to do frame introspection at certain instruction pointers for jitted code, so if a signal comes in at a different one, we will throw an assert when trying to do the traceback since we won't find any stackmap info.

Maybe they're both solvable, but I didn't want to keep pursuing that route.  The second issue is probably important to the destructor stuff that you and @rudi-c are working on, since it means that really can't call destructors from arbitrary allocation sites :/ (ie we might not have frame introspection info available)

toshok:
re: validating things - I didn't mean that we would unconditionally enable the #ifdef'ed pieces, just that the part in the #ifdef would be behind the command line flag check:

i.e. instead of:
```c++
+#if ENABLE_SAMPLING_PROFILER
+            auto stmt = block->body[i];
+            if (stmt->type != AST_TYPE::Assign) // could be a landingpad
+                doSafePoint(block->body[i]);
+#endif
```

we'd have:

```c++
        if (ENABLE_SAMPLING_PROFILER) {
            auto stmt = block->body[i];
            if (stmt->type != AST_TYPE::Assign) // could be a landingpad
                doSafePoint(block->body[i]);
        }
```

the additional branch cost shouldn't be a problem (there's only 1 additional branch per #ifdef), and it won't impact the runtime of jitted code at all.

kmod:
I agree, I think that part wouldn't be too bad, but I would still worry about the additions to allowGLReadPreemption and the interpreter :/  I'm sure we could figure out how to make those fast when turned off, but I didn't feel like it was worth it for now.

PR #568, opened 2015-05-29
Add extra tests for babel and cheetah

undingen:


undingen:
Closing because I split this up into #597 and #671.

Issue #450, opened 2015-04-20
Get more CPython tests working


kmod:
The cpython test suite is a good source of tasks; we're currently failing 92 of 142 tests, and a lot of them for fairly simple reasons.  ex `make run_test_pow` runs the "test_pow.py" CPython test, and fails with `NameError: global name 'pow' is not defined`.

I think the best source of info on the state of the testsuite is [here](https://github.com/dropbox/pyston/blob/master/test/CPYTHON_TEST_NOTES.md).

arbipher:
It seems a good start to me.

kmod:
I just took another pass over the [notes doc](https://raw.githubusercontent.com/dropbox/pyston/master/test/cpython/NOTES.org) and it should be current again.  I also added the "probably easy" label here since there are a bunch of easy tasks in the list, though it seems like there are also some quite hard ones too.

I'm not sure happy with the current way of managing the documentation on the failing tests (ie the text file in the repo), but it feels like there'd be drawbacks to mass-importing it into github issues so for now I don't feel like going through that process.

deronnax:
the notes you link to are 404. Maybe you're refering to https://github.com/dropbox/pyston/blob/master/test/CPYTHON_TEST_NOTES.md ? Maybe you should update the link you give

undingen:
thanks :+1:  I edited the comments to refer to the correct file :-)

deronnax:
you linked to the very raw file which is pretty unpleasant to read. I think you would prefer to link the not-raw version, like the link I gave :)

PR #569, opened 2015-05-31
add long.__float__

aisk:
https://github.com/dropbox/pyston/issues/450


kmod:
Hi, can we get a quick test for this?  I know it's a bit weird to ask for since the motivation of this was to get the CPython test to work, but until we have that test enabled, the code you added is untested.  Just something simple like calling `print float(1L), (2L).__float__()`.

undingen:
And you have to change the ```float``` to a ```double``` otherwise ```(16777217L).__float__()``` will return ```16777216.0``` :-D (haven't tried it but the code looks like it)

src/runtime/long.cpp
```diff
@@ -688,6 +688,18 @@ Box* longInt(Box* v) {
         return new BoxedInt(n);
 }
 
+Box* longFloat(BoxedLong* v) {
+    if (!isSubclass(v->cls, long_cls))
+        raiseExcHelper(TypeError, "descriptor '__float__' requires a 'long' object but received a '%s'",
+                       getTypeName(v));
+    double inf = std::numeric_limits<double>::infinity();
+    double f = mpz_get_d(v->n);
+    if (f == inf || f == -inf) {
```
kmod: Can we switch this to `std::isinf(f)`?

kmod:
So looking at the CPython code for this, it looks like their `__float__` implementation is just a wrapper around `PyLong_AsDouble()`.  I didn't think about it when you started the PR, but it seems like we should do that too in order to reduce duplication.  Case in point: PyLong_AsDouble should do the same overflow checking that you added here :)

Also, minor thing, but my preference is to keep trivial merges out of PRs.  Would you mind rebasing at some point?

aisk:
Got, I will do this :)

kmod:
Superseded by #683

PR #573, opened 2015-06-01
Switch hash to using tp_hash

kmod:
We weren't even doing any rewriting for hash, so there's not much
downside.  This also cuts down on boxing quite a bit since
we can usually avoid boxing the hash value.

### DIFF
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -685,6 +685,9 @@ static PyObject* slot_tp_str(PyObject* self) noexcept {
 }
 
 static long slot_tp_hash(PyObject* self) noexcept {
+    static StatCounter slowpath_hash("slowpath_hash");
+    slowpath_hash.log();
+
     PyObject* func;
     static PyObject* hash_str, *eq_str, *cmp_str;
     long h;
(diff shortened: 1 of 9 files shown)


kmod:
This seems overall helpful, but it showing some weird regressions:
```
              pyston django_migrate.py                  :    2.2s baseline: 2.2 (-1.6%)
              pyston virtualenv_bench.py                :    5.9s baseline: 5.8 (+0.8%)
              pyston django-template.py                 :   15.4s baseline: 15.7 (-1.9%)
              pyston interp2.py                         :    5.7s baseline: 5.7 (+0.5%)
              pyston raytrace.py                        :    6.4s baseline: 6.4 (+0.5%)
              pyston nbody.py                           :    8.4s baseline: 8.6 (-3.2%)
              pyston fannkuch.py                        :    7.5s baseline: 7.3 (+2.4%)
              pyston chaos.py                           :   20.9s baseline: 20.8 (+0.8%)
              pyston fasta.py                           :    5.6s baseline: 5.4 (+4.4%)
              pyston pidigits.py                        :    5.7s baseline: 5.7 (-0.1%)
              pyston richards.py                        :    2.5s baseline: 2.5 (-0.6%)
              pyston deltablue.py                       :    1.8s baseline: 1.8 (-0.3%)
              pyston (geomean-1a6f)                     :    5.8s baseline: 5.8 (+0.1%)
```

Looking into fasta.py, it seems to have regressed due to intLtInt being more expensive...?  I'm going to look into these perf oddities.

kmod:
I have no idea why we see that ~4% regression on fasta.  intLtInt somehow became 3x as expensive, despite having the exact same assembly.  jitdev has limited hardware counter support so I tried looking into it on my desktop, but I don't get the same behavior there.

I'm going to go with my gut and say that this is the right way forward even if the benchmarks don't quite back it up for now.

PR #578, opened 2015-06-03
for getattrFunc use getattrInternal instead of getattr

toshok:
getattr throws an exception if the attribute is not present.  getattrFunc already throws the same exception (if there isn't a default value passed in).

For some reason Terminal.app has decided to not let me select text, but the result of this change for django-template.py is:

```
  pyston django-template.py        : 8.0s baseline: 10.9 (-26.5%)
```


toshok:
more stats:

before change, `__cxa_throw` is called 142428 times
after change, `__cxa_throw` is called 46857 times

kmod:
cool, that's a pretty nice speedup :)

undingen:
wow amazing speedup!

PR #580, opened 2015-06-03
output stderr/stdout diffs for failed tests

toshok:


kmod:
Hmm what about handling this same way as printing the stderr on failure -- if we're in KEEP_GOING, then we print the diagnostics inline if VERBOSE?

toshok:
factored out the diffing, and used it for both stderr and stdout.  example from an intentional failure:

```
$ make quick_check
                                1.py    FAILED (bad stderr)
--- /tmp/expectederr_4Zs258	2015-06-08 22:51:00.080876999 +0000
+++ /tmp/receivederr_IkmCfP	2015-06-08 22:51:00.080876999 +0000
@@ -1 +0,0 @@
-I AM CPYTHON
```

kmod:
cool :)

PR #581, opened 2015-06-03
Add the _elementtree and update libpypa

undingen:
- _elementtree has currently better perf than the python implementation


undingen:
added a small fix for: ```d = {2 : 2}; print d[2L]```

PR #583, opened 2015-06-04
allocate runtime ics using mmap (with PROT_EXEC)

toshok:
otherwise valgrind segfaults when we call into the ic.

also mmap the eh_frame to (hopefully) be nearby the ic, so the offset fits.

### DIFF
--- a/src/runtime/ics.cpp
+++ b/src/runtime/ics.cpp
@@ -14,7 +14,9 @@
 
 #include "runtime/ics.h"
 
+#ifndef NVALGRIND
 #include <sys/mman.h>
+#endif
 
 #include "asm_writing/icinfo.h"
 #include "asm_writing/rewriter.h"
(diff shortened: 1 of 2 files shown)


kmod:
Hmm I wonder how much it will hurt to have to do page-size allocations for both the instruction memory and the eh frame... we might want to measure it, or maybe easier just put this change behind an #ifdef so that we can turn it on when we want to use valgrind?  I think we even have an `NVALGRIND` flag that enables certain valgrind support.

toshok:
TL;DR: Put the code in !NVALGRIND blocks, leaving original code in.

I looked a little into performance changes due to it, and it didn't look to have any significant impact on execution time.  It obviously has a larger effect on vmem mappings, but I didn't  measure that.  We could potentially save some of that by allocating both eh_frame + code in the same mmap'ed region, but we'll probably need some other specialized allocator for code memory in order to not waste too much space.

I do think at some point (when we're talking security risks, etc) it would good to revisit this sort of thing, along with the llvm code responsible for jitting (and likely our rewriter/assembler as well) to make sure that pages are W^X.

I can add issues/tasks for the above.

PR #585, opened 2015-06-04
Generator: don't generate the traceback for the StopIteration exc when the generator is exiting back to a hasnext call.

undingen:
This is a small perf improvement for django-template.py: took  7.4ms for last iteration from 8.2ms

### DIFF
--- a/src/runtime/generator.cpp
+++ b/src/runtime/generator.cpp
@@ -161,7 +161,10 @@ static void generatorSendInternal(BoxedGenerator* self, Box* v) {
 
     if (self->entryExited) {
         freeGeneratorStack(self);
-        self->exception = excInfoForRaise(StopIteration, None, None);
+        // Reset the current exception.
+        // We could directly create the StopIteration exception but we delay creating it because often the caller is not
+        // interested in the exception (=generatorHasnext). If we really need it we will create it inside generatorSend.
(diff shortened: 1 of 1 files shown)


undingen:
@toshok mmmh just realized that with the incremental traceback change this should be an non issue?

toshok:
it's not as big an improvement with the incremental traceback change, but it's still a win:

before:
```
took  6.8ms for last iteration

real	0m6.058s
user	0m5.880s
sys	0m0.171s
```

after:
```
took  6.6ms for last iteration

real	0m5.934s
user	0m5.750s
sys	0m0.171s
```

so :+1: here :)

kmod:
Good idea :)  In trying to understand the generator control flow, I discovered we don't handle this correctly (we assert):

```python
def f():
    yield 1/0

g = f()

try:
    g.next()
except Exception as e:
    print type(e), e

try:
    g.next()
except Exception as e:
    print type(e), e
```

undingen:
yeah understanding the control flow of the generator stuff is really hard :-(.
I can fix you example if you like - should be in the same lines of code I just touched.

kmod:
Turns out there wasn't much to do -- #589

undingen:
:-)

Issue #254, opened 2015-01-02
Finish old-style-class support


kmod:
We need to delegate the new-style special attributes (set on the class "instance") to the old-style ones (set on instances of the "classobj" class).  Each individual one isn't too hard but we just need to go through and add them and make sure they are tested.  Here's what one looks like, from codegen/classobj.cpp:

```c++
Box* instanceGetitem(Box* _inst, Box* key) {
    RELEASE_ASSERT(_inst->cls == instance_cls, "");
    BoxedInstance* inst = static_cast<BoxedInstance*>(_inst);

    Box* getitem_func = _instanceGetattribute(inst, boxStrConstant("__getitem__"), true);
    return runtimeCall(getitem_func, ArgPassSpec(1), key, NULL, NULL, NULL, NULL);
}

void setupClassobj() {
    [...]
    instance_cls->giveAttr("__getitem__", new BoxedFunction(boxRTFunction((void*)instanceGetitem, UNKNOWN, 2)));
    [...]
}
```

This could probably also be made generic, since it's all boilerplate except for the name of the attribute and the number of arguments.

Browsing https://docs.python.org/2/reference/datamodel.html it looks like we don't support the following dunder methods:
- [ ] lt/le/eq/ne/gt/ge
- [ ] cmp/rcmp
- [x] hash
- [x] setattr, delattr
- [ ] get, set, delete
- [x] call
- [x] delitem, contains
- [ ] missing, reversed
- [x] iter
- [ ] getslice, setslice, delslice
- [x] add/sub/mul/floordiv/mod/...
- [x] radd/rsub/rmul/...
- [x] iadd/isub/imul/...
- [ ] neg, pos, abs, invert, complex, int, long, float
- [ ] coerce
- [ ] enter, exit

corona10:
i fix '__eq__ overload not working #253' on pull request #380 (i think i need to sperate pull request but i don't know how to do.. sorry)

dorafmon:
These two functions in ``classobject.cpp`` share the exact same structure:
```C++
Box* instanceEq(Box* _inst, Box* other) {
    RELEASE_ASSERT(_inst->cls == instance_cls, "");
    BoxedInstance* inst = static_cast<BoxedInstance*>(_inst);

    Box* func = _instanceGetattribute(inst, boxStrConstant("__eq__"), false);
    if (!func)
        return NotImplemented;
    return runtimeCall(func, ArgPassSpec(1), other, NULL, NULL, NULL, NULL);
}

[...]

PR #586, opened 2015-06-04
Remove redundant code and pave way for #254

dorafmon:
So in #254  all binary operations can be implemented by the newly added function ``_instanceBinary``. Do you think this is the right way to do it? If so I will go ahead and add more code that implement other functions other than ``__eq__`` and  ``__ne__``.

### DIFF
--- a/src/runtime/classobj.cpp
+++ b/src/runtime/classobj.cpp
@@ -712,24 +712,22 @@ static PyObject* instance_index(PyObject* self) noexcept {
     return res;
 }
 
-Box* instanceEq(Box* _inst, Box* other) {
+Box* _instanceBinary(Box* _inst, Box* other, const char* attr) {
     RELEASE_ASSERT(_inst->cls == instance_cls, "");
     BoxedInstance* inst = static_cast<BoxedInstance*>(_inst);
 
-    Box* func = _instanceGetattribute(inst, boxStrConstant("__eq__"), false);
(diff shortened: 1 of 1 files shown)


kmod:
Looks good :)  If you're interesting in working on this I think there are a couple other things we could do: first, we should only do the `boxStrConstant` call once, at initialization, rather than once per call through this.  You can do something like `static Box* eq_box = PyString_InternFromString("__eq__");`.

A bigger change, but worth doing if you're interested, is switching from Python-level attributes to a tp_richcompare function.  I just did this in https://github.com/dropbox/pyston/commit/9cb2b9f1eda1cc13143ba1adef49f4362aed92f2 for int+long+str, and I *think* it's a performance improvement in addition to definitely a compatibility improvement.  I think concretely, this would mean copying over instance_richcompare() + half_richcompare() + init_name_op() from CPython's classobject.c.  You might have to convert some of the function calls from their functions to ours, and make sure to translate between exception models if necessary.

dorafmon:
Thanks for the merge, sure I will have a look and see what I can do. BTW I think the code now is too undocumented if you see what I mean. It's often quite difficult for me to find a function or a method. This makes understanding the code much harder. Is it possible that we should slowly add some document, or even better with some documentation generator (Sphinx for example)?

dorafmon:
I suggest to open an dedicated issue to discuss this.

PR #588, opened 2015-06-05
Fix compile error when enabling memstats + call tp_richcompare even if rewrite_args == NULL

undingen:


kmod:
oops thanks for the fixes!

PR #589, opened 2015-06-05
Slightly change generator StopIteration handling

kmod:
Unconditionally raise a StopIteration when trying to iterate
a finished generator, regardless of what originally stopped
it from iterating.


src/runtime/generator.cpp
```diff
@@ -124,6 +124,7 @@ static void generatorSendInternal(BoxedGenerator* self, Box* v) {
     // check if the generator already exited
     if (self->entryExited) {
         freeGeneratorStack(self);
+        raiseExcHelper(StopIteration, (const char*)nullptr);
         return;
```
toshok: could probably do without the return here now that we're raising

kmod: good point :)

PR #592, opened 2015-06-05
free generator even if the don't exit

undingen:
This is not yet ready but I would like to get early feedback from travis-ci.

### DIFF
--- a/src/codegen/irgen/irgenerator.cpp
+++ b/src/codegen/irgen/irgenerator.cpp
@@ -117,8 +117,7 @@ static llvm::Value* getBoxedLocalsGep(llvm::IRBuilder<true>& builder, llvm::Valu
 
 static llvm::Value* getExcinfoGep(llvm::IRBuilder<true>& builder, llvm::Value* v) {
     static_assert(offsetof(FrameInfo, exc) == 0, "");
-    static_assert(offsetof(ExcInfo, type) == 0, "");
-    return builder.CreateConstInBoundsGEP2_32(builder.CreateConstInBoundsGEP2_32(v, 0, 0), 0, 0);
+    return builder.CreateConstInBoundsGEP2_32(v, 0, 0);
 }
 
 static llvm::Value* getFrameObjGep(llvm::IRBuilder<true>& builder, llvm::Value* v) {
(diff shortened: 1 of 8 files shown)


tjhance:
totally random question: if we free a generator that hasn’t returned yet, what happens if the execution was in the middle of a `try` which has a `finally` block? Does the code in the `finally` block get run? What is supposed to happen, and do we handle this here?

undingen:
Thanks for the interesting idea.
```python
 def G():
     try:
         yield 0
         yield 1
         print "end"
     finally:
         print "finally"
 
 def foo():
     g = G()
     print g.next()
     print g.next()
 foo()
```
outputs:
```
0
1
finally
```
So the finally block does get executed... Not sure how to handle that yet.

toshok:
do cpython generators have a tp_dealloc that handles unwinding the internal
state?  we could e.g. do generatorThrow inside its finalizer

toshok:
maybe add an 'except:' clause to the test and print out the exception if
one is thrown?

undingen:
The gcmodule does it: it checks for finalizers and handles generators specially.
And there does not get an exception thrown.

toshok:
hm, yuck.  that leaves another possibility for resurrection outside of
finalizers

undingen:
I cleaned it up a little bit and fixed the virtualenv crash.
This should be now ready for review.
It does not implement the ```finally``` cleanup behavior but we don't call it without this patch either so it shouldn't make things worse and it's probably something which should get fixed when we support ```__del__``` / non trivial destructors.

kmod:
I looked into it, and it looks like CPython does throw a GeneratorExit exception which is what triggers finally blocks.  So I was hoping we could just rely on that to unwind the stack frames, but the generator code is free to ignore the exception, in which case we'd still end up with the dead-but-rooted generator->interpreter->generator cycles :/

src/codegen/irgen/irgenerator.cpp
```diff
@@ -180,8 +179,16 @@ llvm::Value* IRGenState::getFrameInfoVar() {
...
+            // frame_info.exc.value = NULL
+            // frame_info.exc.traceback = NULL
+            llvm::Constant* null_value = getNullPtr(g.llvm_value_type_ptr);
+            llvm::Value* exc_info = getExcinfoGep(builder, al);
+            builder.CreateStore(
+                null_value, builder.CreateConstInBoundsGEP2_32(exc_info, 0, offsetof(ExcInfo, type) / sizeof(Box*)));
+            builder.CreateStore(
+                null_value, builder.CreateConstInBoundsGEP2_32(exc_info, 0, offsetof(ExcInfo, value) / sizeof(Box*)));
+            builder.CreateStore(null_value, builder.CreateConstInBoundsGEP2_32(exc_info, 0, offsetof(ExcInfo, traceback)
+                                                                                                / sizeof(Box*)));
```
kmod: Did we need this for when we deopt and pass the frameinfo back to the interpreter?  If so, maybe we could keep this optimization if we zeroed out `value` and `traceback` on deopt?  But I'm not really sure how much that the optimization even helps in the first place -- it's a very small win, but it happens on every function entry.

undingen: Yes it's for the deopt. I created https://github.com/dropbox/pyston/pull/595 to bring back the old behavior.

PR #593, opened 2015-06-06
flesh out super() handling a bit more

toshok:
super() can be partially applied, and it acts as a descriptor, filling in the
missing pieces.

test/tests/super_getattribute.py is adapted from part of the cpython test suite

### DIFF
--- a/src/runtime/super.cpp
+++ b/src/runtime/super.cpp
@@ -16,6 +16,7 @@
 
 #include <sstream>
 
+#include "capi/typeobject.h"
 #include "capi/types.h"
 #include "core/types.h"
 #include "gc/collector.h"
@@ -128,10 +129,10 @@ Box* superGetattribute(Box* _s, Box* _attr) {
         }
     }
 
-    Box* r = getattrInternalGeneric(s, attr->s(), NULL, true, false, NULL, NULL);
-    if (r)
-        return processDescriptor(r, s, s->cls);
-    raiseExcHelper(AttributeError, "'super' object has no attribute '%s'", attr->data());
+    Box* rtn = PyObject_GenericGetAttr(_s, _attr);
+    if (!rtn)
+        throwCAPIException();
+    return rtn;
 }
 
 Box* superRepr(Box* _s) {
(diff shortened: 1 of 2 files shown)


src/runtime/super.cpp
```diff
@@ -128,10 +128,10 @@ Box* superGetattribute(Box* _s, Box* _attr) {
...
     }
 
-    Box* r = typeLookup(s->cls, std::string(attr->s()), NULL);
-    // TODO implement this
-    RELEASE_ASSERT(r, "should call the equivalent of objectGetattr here");
-    return processDescriptor(r, s, s->cls);
+    Box* r = getattrInternalGeneric(s, attr->s(), NULL, true, false, NULL, NULL);
+    if (r)
+        return processDescriptor(r, s, s->cls);
+    raiseExcHelper(AttributeError, "'super' object has no attribute '%s'", attr->data());
```
kmod: Can we do this the way CPython does?

src/runtime/super.cpp
```diff
@@ -187,6 +187,36 @@ Box* superInit(Box* _self, Box* _type, Box* obj) {
...
+           call its type */
+        return runtimeCall(su->cls, ArgPassSpec(2, 0, false, false), su->type, obj, NULL, NULL, NULL);
+    } else {
+        /* Inline the common case */
+        BoxedClass* obj_type = supercheck(su->type, obj);
+        if (obj_type == NULL)
+            return NULL;
+        newobj = new (su->cls) BoxedSuper(NULL, NULL, NULL);
+        if (newobj == NULL)
+            return NULL;
```
kmod: These should be `throwCapiException()`.  I usually prefer though to just leave the CPython code unmodified, ie have this be a capi-style function and stick it into tp_descr_get.

toshok:
hm, sorry about that - I could have sworn I added a comment to this PR :/

to paraphrase - tp_descr_get inheritance isn't working for us.  subclasses of super() get tp_descr_get set to NULL.  I was planning to get the feature done using `__get__` then go back and fix tp_descr_get to work, and update things here.

Given that there were problems with the code partially ported to `__get__`, I'll go back and fix tp_descr_get inheritance and update this PR.

kmod:
We should call `add_operators(super_cls)` right before `super_cls->freeze()` -- I think that might be the source of the tp_ inheritance issue?  Sorry for the landmine; I think we've gotten to the point that freeze() could directly call add_operators, but I'm not sure.

PR #594, opened 2015-06-06
Some `__future__` import stuff

tjhance:
- Move `FutureFlags` from `BoxedModule` to `SourceInfo`.
- Handle `__future__` imports in exec statements.
- Handle the `division` import in cfg gen rather than later.

(This was prompted by a cpython test case that uses `from __future__ import division` in an exec statement.)

Seems like there is still a lot of cleaning up to do, but this is a step in the right direction. Notably, we have our own constants for the future imports, even though they're defined in cpython too.

### DIFF
--- a/from_cpython/Include/Python.h
+++ b/from_cpython/Include/Python.h
@@ -111,6 +111,8 @@
 
 #include "pyfpe.h"
 
+#include "code.h"
+
 #define Py_single_input 256
 #define Py_file_input 257
 #define Py_eval_input 258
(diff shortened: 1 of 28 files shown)


tjhance:
Did a bit more work here today:

- Got rid of our constants in order to use the cpython ones
- Get `compile` to handle the flags correctly, implement the `dont_inherit` parameter
- More tests

tjhance:
Fixed the build, hopefully

kmod:
cool :)

PR #595, opened 2015-06-09
Skip initializing exc fields inside JITed code

undingen:


src/runtime/objmodel.cpp
```diff
@@ -173,6 +173,11 @@ extern "C" Box* deopt(AST_expr* expr, Box* value) {
     // Should we only do this selectively?
     execution_point.cf->speculationFailed();
 
+    // Except of exc.type we skip initializing the exc fields inside the JITed code path (small perf improvement) that's
+    // why we have todo it now.
+    frame_state.frame_info->exc.traceback = NULL;
+    frame_state.frame_info->exc.value = NULL;
```
kmod: I think we should do this only when `exc.type == NULL`?  Or if that's always the case, maybe we should assert on it?

undingen: Oh good point!

PR #597, opened 2015-06-09
Add more external projects to our travis-ci test suite

undingen:
This are still a few left but they other fail currently.
Cheetah produces currently quite a lot of error because of a ```is``` comparison which fails.
(see: https://bitbucket.org/pypy/compatibility/wiki/cheetah)


kmod:
Looks great :)

If you're going to be poking around with this some more, one nice thing to do is list the revisions of the dependencies of the libraries we `pip install`; we ran into this where one of the dependencies of bcrypt got updated on pypi and then we started silently testing against that.

PR #604, opened 2015-06-10
New benchmarks

toshok:
adds minibenchmarks for a some simple sqlalchemy and pyxl usage.  also includes a combined runner (minibenchmarks/combined.py) which runs them all in succession.

cpython time for combined.py:
```
real	0m3.837s
user	0m3.784s
sys	0m0.050s
```

pyston_release time:
```
real	0m16.385s
user	0m15.953s
sys	0m0.416s
```


toshok:
incremental tracebacks only knocks out ~3 seconds here as well, so it looks like pyxl and sqlalchemy don't have the same cavalier approach to raising exceptions as django does:

``
real	0m13.305s
user	0m12.911s
sys	0m0.369s
```

kmod:
Awesome!  Just added to pyston-perf as well, so they should show up on speed.pyston.org once I run the benchmarks later tonight.

PR #605, opened 2015-06-10
Use BoxedString in codegen

kmod:
instead of using std::string.  the idea is that we will often end up converting to BoxedString in the runtime, so we might as well use those from the beginning to reduce on converting back and forth.  BoxedStrings are probably slightly more expensive than std::string, but I think it still works out.

Here's the perf data:
```
   django_migrate.py             2.2s (2)             2.2s (2)  -0.5%
 virtualenv_bench.py             8.7s (2)             8.7s (2)  -0.2%
  django-template.py             7.8s (6)             8.1s (6)  +3.0%
          interp2.py             5.4s (2)             5.0s (2)  -8.6%
         raytrace.py             6.8s (4)             6.4s (4)  -5.2%
            nbody.py             9.2s (2)             9.4s (2)  +3.0%
         fannkuch.py             7.8s (2)             7.4s (2)  -4.0%
            chaos.py            21.6s (2)            21.4s (2)  -0.9%
            fasta.py             5.5s (2)             5.3s (2)  -3.4%
         pidigits.py             6.1s (2)             6.3s (2)  +3.3%
         richards.py             2.7s (2)             2.6s (2)  -3.2%
        deltablue.py             1.9s (2)             1.8s (2)  -3.2%
             geomean                 5.8s                 5.7s  -1.7%
```

I don't like the 3% regression on django-template... it's hard to pin down exactly why it's happening, but some of it looks like it's coming from extra GC time.  We actually do fewer allocations and fewer gcs (87 vs 92), but I guess each one is longer due to the extra heap size.  I think this is something that we can address by being better about freeing the strings that this PR adds (currently they all become immortal), and by GC improvements in general (heap-size-dependent collection frequency?).

sorry for the huge diff


toshok:
it's an improvement on the incremental-traceback branch.  from ~5.990 to ~5.890

PR #606, opened 2015-06-11
factor out find_cf to pyston::binarySearch, and use to locate large/huge objects

toshok:
LargeArena::allocationFrom has been showing up in the perf traces forever, and it's only going to get worse with the finalizer passes.


toshok:
@rudi-c does this help with the finalizer perf?  I had to make some additional changes (and get rid of the linked list used for allocated objects.)

PR #607, opened 2015-06-13
cmake testing

dagar:
* add cmake option ENABLE_EXTRA_TESTS for tests that should only run on travis-ci
 * change `make check` and `make quick_check` testing to use ctest
  * this removes duplication, matches local testing to travis-ci, and provides a nice test summary

The exact `make check` and ctest test options have drifted a bit. This might be a good time to verify exactly what you want the testing to be.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -512,28 +512,12 @@ cpplint:
 .PHONY: check
 check:
 	@# These are ordered roughly in decreasing order of (chance will expose issue) / (time to run test)
-	$(MAKE) lint
-	$(MAKE) check_format
+
 	$(MAKE) pyston_dbg $(CHECK_DEPS)
-
-	$(PYTHON) $(TOOLS_DIR)/tester.py -R pyston_dbg -j$(TEST_THREADS) -k -a=-S $(TESTS_DIR) $(ARGS)
-	@# we pass -I to cpython tests & skip failing ones because they are sloooow otherwise
-	$(PYTHON) $(TOOLS_DIR)/tester.py -R pyston_dbg -j$(TEST_THREADS) -k -a=-S --exit-code-only --skip-failing -t30 $(TEST_DIR)/cpython $(ARGS)
-	$(PYTHON) $(TOOLS_DIR)/tester.py -R pyston_dbg -j$(TEST_THREADS) -k -a=-S --exit-code-only --skip-failing -t600 $(TEST_DIR)/integration $(ARGS)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -951,12 +888,12 @@ CMAKE_SHAREDMODS := sharedmods ext_pyston
 
 .PHONY: pyston_dbg pyston_release
 pyston_dbg: $(CMAKE_SETUP_DBG)
-	$(NINJA) -C $(CMAKE_DIR_DBG) pyston copy_stdlib copy_libpyston $(CMAKE_SHAREDMODS) ext_cpython $(NINJAFLAGS)
+	$(NINJA) -C $(CMAKE_DIR_DBG) pyston copy_stdlib copy_libpyston $(CMAKE_SHAREDMODS) ext_cpython clang-format unittests $(NINJAFLAGS)
```
kmod: Hmm this seems like the wrong place to place the dependency.  I assume it's for the check/quick_check targets?  Maybe we should have those build `check-deps` rather than getting it through the pyston_dbg dependency.  (ex I also wonder if `make check` would have the issue that these dependencies aren't listed for the release build.)

dagar: I'll change it to have the existing Makefile CHECK_DEPS build Debug check-deps for now.

dagar:
Fixed check-deps and rebased.

kmod:
Sorry if I'm misunderstanding, but won't the `( cd $(CMAKE_DIR_RELEASE) && ctest -V -R pyston )` line need to have the release-mode check-deps built?  ie won't it try to build the unittests, run clang-format, etc?

kmod:
Ah ok, didn't realize the -R flag was deselecting those types of tests.

kmod:
Hey dagar, could you help take a look at this build?
https://travis-ci.org/dropbox/pyston/jobs/67604256

Looks like the compile failed (actual bug in my PR I guess), but it got kind of hidden by the new fold, and then for some reason it still tried running all the tests but of course that failed since we didn't build the binary.  It looks like Travis-CI noticed the build failed, and it sounds like it's supposed to do the right thing [and not fold in this case](http://blog.travis-ci.com/2013-05-22-improving-build-visibility-log-folds/), so I'm not sure what happened here.

dagar:
That's weird, I thought it was supposed to stop after one failed step. It's easy enough to recombine minus the fold hiding verbose compile output. We now have to run ctest manually to get regular console output so that travis-ci won't timeout.

Fixed in PR #627

PR #608, opened 2015-06-15
Revert to using gettimeofday to have measurements in us instead of ticks.

rudi-c:
Also removing logic in Stats::dump that counts the total number of ticks
in timers. It relies on specific timer naming conventions that aren't
currently being followed. If, at a later time, we are interested in
having the sum of timer ticks, we should refactor the code to have a
separate class for counters that log time.

### DIFF
--- a/src/core/stats.cpp
+++ b/src/core/stats.cpp
@@ -138,41 +138,17 @@ void Stats::dump(bool includeZeros) {
 
     std::sort(pairs.begin(), pairs.end());
 
-    uint64_t ticks_in_main = 0;
-    uint64_t accumulated_stat_timer_ticks = 0;
     for (int i = 0; i < pairs.size(); i++) {
         uint64_t count = *pairs[i].second;
         if (includeZeros || count > 0) {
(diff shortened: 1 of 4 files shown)


src/core/stats.cpp
```diff
@@ -132,41 +132,13 @@ void Stats::dump(bool includeZeros) {
...
-                fprintf(stderr, "%s: %lu\n", pairs[i].first.c_str(), (uint64_t)(count / cycles_per_us));
-
-            } else
-                fprintf(stderr, "%s: %lu\n", pairs[i].first.c_str(), count);
-
-            if (startswith(pairs[i].first, "us_timer_"))
-                accumulated_stat_timer_ticks += count;
-
-            if (pairs[i].first == "ticks_in_main")
-                ticks_in_main = count;
```
kmod: I think we should keep the conversion even if we don't do the accumulation + assertion that it's the same as main.  I think the first part should be ok if we just change the checked prefix from `"us_"` to '"us_timer_"`?

PR #611, opened 2015-06-16
cache + reuse 1-length strings

kmod:
Seems like a no-brainer, but the perf results are less convincing than I would have liked:

```
       django_template.py             8.1s (6)             8.1s (6)  +0.4%
            pyxl_bench.py             5.7s (4)             5.6s (4)  -1.8%
 sqlalchemy_imperative.py             2.7s (4)             2.7s (4)  -0.4%
        django_migrate.py            2.1s (12)            2.1s (13)  +1.0%
      virtualenv_bench.py             8.5s (8)             8.5s (4)  +0.1%
```

### DIFF
--- a/microbenchmarks/str_find_ubench.py
+++ b/microbenchmarks/str_find_ubench.py
@@ -1,0 +1,9 @@
+def f():
+    s = "hello world!" * 500
+    count = 0
+    for i in xrange(10000):
+        for c in s:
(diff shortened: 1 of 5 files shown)


toshok:
yeah I've done this same optimization before and didn't bother with the pull request because the results were so less-than-spectacular :/

PR #636, opened 2015-06-24
Add debug symbols to release mode.

rudi-c:
It's useful to have.


kmod:
Nice, I was going to do this too.  We should make sure to strip them out when we do a release so that people don't complain about the binary size.

PR #637, opened 2015-06-24
Specify /bin/bash for the Makefile, as some distros may have other shells as /bin/sh

lameiro:
Specify /bin/bash for the Makefile, as some distros may have other shells (eg, ubuntu image for docker has /bin/sh pointing to /bin/dash).
This causes failures when bash extensions are used (in my case, the time command cannot run)


kmod:
good catch :)

PR #638, opened 2015-06-24
_U_dyn_register: Use binary search to find insert location

undingen:
Don't think this is currently showing up much, but if we start to use more runtime ICs (e.g. new JIT tier) this shows up as a serious bottleneck (and with this change it disappears from the top perf output).

Instead of making a patch for the patch I directly changed the libunwind patch...
Implementation is inspired by ```std::lower_bound``` http://en.cppreference.com/w/cpp/algorithm/lower_bound


kmod:
Man, reading a diff-of-patches is not the easiest thing :P

PR #642, opened 2015-06-25
Incremental tb improvements

toshok:
0fbafcf is the meat of this change.  brain dump follows:

We have two `*Unwinders` classes which deal with the two scenarios that call for unwinding:
1. `PythonUnwinder`: used for stack introspection/traceback generation on abort() (basically used only in unwindPythonStack).  Includes code specifically to trace through generators to their calling contexts.
2. `PythonExceptionUnwinder`: used for mapping between C++ unwinding and python unwinding.

These two use the same state class to tell if a given frame should be part of the traceback: `PythonUnwindState`.  `PythonUnwindState` has 1 piece of state, `skipNext`.  It has a method that it assumes will be called for all frames during a given unwind:

`bool getPythonFrame(unw_cursor_t* cursor, PythonFrameIteratorImpl* info) `

which will return true (and fill in *info) if the cursor corresponds to a python frame that the caller should deal with (i.e. it wasn't skipped due to re-raise or OSR.)

using `PythonUnwinder` + `PythonUnwindState` results in unwindPythonStack being pretty tiny now, just a loop over the frames it should be interested in.

For `PythonExceptionUnwinder` things are necessarily a little more complex, but this PR should simplify all the interactions it had with cxx_unwind.cpp:

Each thread has its own local `PythonExceptionUnwinder` instance and a boolean flag for whether a given C++ exception was a python re-raise.  I added the api `throwReraise` which sets this flag to true before doing the throw.  This gets rid of that additional bool in ExcInfo.

the api now corresponds more closely to the C++ unwinder:

1. `__cxa_allocate_exception` now calls (and returns the return value of) `pyston::allocatePythonExceptionStorage()`
2. `__cxa_throw` now calls `pyston::beginPythonException` before starting the unwind loop
3. `unwind_loop` calls `pyston::unwindingThroughFrame` for every C++ frame.
3. `unwind_loop` now calls `pyston::endPythonException` before branching to pa non-cleanup handler.

the `PythonExceptionUnwinder` uses `PythonUnwindState` as well, and resets the state upon every `beginPythonException` call (supplying the thread local reraise flag as the initial value of `skipNext`).

---

somewhat unrelated, but I got rid of the Box subclass used so we could scan the ExcInfo fields and added a handler function that can be used to scan roots.  This saves us 1 allocation for every getTopFrame call (36762 in django_template.py) and c++ exception thrown.

I'm surprised that the perf win from the reduced allocations disappeared after rebasing, but the change *definitely* cleans this up.

```
       django_template.py             5.5s (4)             5.5s (4)  +0.3%
            pyxl_bench.py             4.0s (4)             4.0s (4)  -0.3%
 sqlalchemy_imperative.py             2.2s (4)             2.2s (4)  +0.2%
        django_migrate.py             2.0s (4)             2.0s (4)  -0.4%
      virtualenv_bench.py             8.0s (4)             8.2s (4)  +1.5%
                  geomean                 3.8s                 3.8s  +0.2%
```

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -382,7 +382,7 @@ Box* ASTInterpreter::execJITedBlock(CFGBlock* b) {
     } catch (ExcInfo e) {
         AST_stmt* stmt = getCurrentStatement();
         if (stmt->type != AST_TYPE::Invoke)
-            throw e;
+            throwReraise(e);
 
         auto source = getCL()->source.get();
         exceptionCaughtInInterpreter(LineInfo(stmt->lineno, stmt->col_offset, source->fn, source->getName()), &e);
--- a/src/codegen/irgen/irgenerator.cpp
+++ b/src/codegen/irgen/irgenerator.cpp
@@ -110,8 +110,8 @@ static llvm::Value* getClosureElementGep(IREmitter& emitter, llvm::Value* closur
 
 static llvm::Value* getBoxedLocalsGep(llvm::IRBuilder<true>& builder, llvm::Value* v) {
     static_assert(offsetof(FrameInfo, exc) == 0, "");
-    static_assert(sizeof(ExcInfo) == 32, "");
-    static_assert(offsetof(FrameInfo, boxedLocals) == 32, "");
+    static_assert(sizeof(ExcInfo) == 24, "");
+    static_assert(offsetof(FrameInfo, boxedLocals) == 24, "");
     return builder.CreateConstInBoundsGEP2_32(v, 0, 1);
 }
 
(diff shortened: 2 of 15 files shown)


toshok:
rebased off master.  current perf comparison:

```
                           1ee49a674757637588:  db35996460f84821ac:
       django_template.py             4.6s (2)             4.5s (4)  -0.6%
            pyxl_bench.py             3.8s (2)             3.8s (4)  +0.2%
sqlalchemy_imperative2.py             5.0s (2)             5.0s (4)  -1.1%
        django_migrate.py             1.8s (2)             1.8s (4)  -1.0%
      virtualenv_bench.py             7.8s (2)             7.7s (4)  -0.1%
                  geomean                 4.1s                 4.1s  -0.5%
```

kmod:
Superseded by #815

PR #644, opened 2015-06-25
Include total-tests-run in our test-suite-results checking

kmod:
I'm seeing fewer failures locally than specified in the file, and
I'm guessing it's because I'm somehow skipping some tests, but I'm
not sure.


kmod:
Bleh was hoping it was this.

PR #646, opened 2015-06-25
Temporary fix for freeing-classes-before-instances bug

kmod:
Was running into it, so just added a quick fix that should hopefully
get us through until rudi gets in his real fix.

This commit just delays freeing the class objects until the end of the
sweep.  It doesn't deal with a number of corner cases such as if
a class and a metaclass get freed at the same time (they both get delayed
to the same point), or if we free a class with weakrefs to it (it just
gets treated like a normal weakreferent).

### DIFF
--- a/src/gc/collector.cpp
+++ b/src/gc/collector.cpp
@@ -375,10 +375,10 @@ void markPhase() {
 #endif
 }
 
-static void sweepPhase(std::vector<Box*>& weakly_referenced) {
+static void sweepPhase(std::vector<Box*>& weakly_referenced, std::vector<BoxedClass*>& classes_to_free) {
     // we need to use the allocator here because these objects are referenced only here, and calling the weakref
     // callbacks could start another gc
-    global_heap.freeUnmarked(weakly_referenced);
+    global_heap.freeUnmarked(weakly_referenced, classes_to_free);
 }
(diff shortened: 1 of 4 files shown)


kmod:
```
       django_template.py             5.7s (2)             5.7s (2)  +0.8%
            pyxl_bench.py             3.9s (2)             3.9s (2)  +0.4%
 sqlalchemy_imperative.py             2.3s (2)             2.2s (2)  -1.0%
        django_migrate.py             2.0s (2)             1.9s (2)  -0.3%
      virtualenv_bench.py             8.1s (2)             8.1s (2)  +0.1%
                  geomean                 3.8s                 3.8s  -0.0%
```

toshok:
just out of curiosity - was there a particular test where this was
happening (with the increase in collection frequency)?

I thought I had a good mental model of the problem and really thought we
were safe as long as we weren't memset'ing.

On Thu, Jun 25, 2015 at 4:47 PM, Kevin Modzelewski <notifications@github.com

kmod:
Oh, hmm yeah the classes should already be kept alive until the weakrefs phase, since they'll get included in someone else's `subclasses` list.  I wasn't able to reproduce the issues I was seeing before.

PR #651, opened 2015-06-27
Add a new JIT tier which is tightly coupled to the AST Interpreter

undingen:
This JIT is tightly coupled to the ASTInterpreter, at every CFGBlock* entry/exit
on can switch between interpreting and directly executing the generated code without having
to do any translations.
Generating the code is pretty fast compared to the LLVM tier but the generated code is not as fast
as code generated by the higher LLVM tiers.
But because the JITed can use runtime ICs, avoids a lot of interpretation overhead and
stores CFGBlock locals symbols inside register/stack slots, it's much faster than the interpreter.

Current perf numbers:
1. run:
```
              pyston (calibration)                      :    1.0s base1: 1.0 (+1.5%)
              pyston django_template.py                 :    5.2s base1: 6.7 (-23.0%)
              pyston pyxl_bench.py                      :    4.6s base1: 5.0 (-8.0%)
              pyston sqlalchemy_imperative2.py          :    5.7s base1: 6.8 (-14.9%)
              pyston django_migrate.py                  :    1.9s base1: 3.1 (-38.4%)
              pyston virtualenv_bench.py                :    6.2s base1: 8.1 (-22.9%)
              pyston interp2.py                         :    3.7s base1: 4.1 (-9.9%)
              pyston raytrace.py                        :    5.8s base1: 6.0 (-3.5%)
              pyston nbody.py                           :    7.3s base1: 7.0 (+4.1%)
              pyston fannkuch.py                        :    6.3s base1: 6.4 (-1.5%)
              pyston chaos.py                           :   17.5s base1: 18.1 (-3.6%)
              pyston fasta.py                           :    4.5s base1: 4.3 (+5.1%)
              pyston pidigits.py                        :    5.8s base1: 5.3 (+8.7%)
              pyston richards.py                        :    1.9s base1: 1.7 (+10.2%)
              pyston deltablue.py                       :    1.4s base1: 1.8 (-20.4%)
              pyston (geomean-10eb)                     :    4.6s base1: 5.1 (-9.5%)
```

best of 3 runs:
```
              pyston (calibration)                      :    1.0s base3: 1.0 (-0.5%)
              pyston django_template.py                 :    4.4s base3: 4.6 (-5.4%)
              pyston pyxl_bench.py                      :    3.9s base3: 3.8 (+2.5%)
              pyston sqlalchemy_imperative2.py          :    4.9s base3: 5.1 (-5.1%)
              pyston django_migrate.py                  :    1.7s base3: 1.7 (+1.7%)
              pyston virtualenv_bench.py                :    5.0s base3: 5.2 (-2.4%)
              pyston interp2.py                         :    3.6s base3: 4.1 (-12.4%)
              pyston raytrace.py                        :    5.5s base3: 5.4 (+1.4%)
              pyston nbody.py                           :    7.0s base3: 7.2 (-2.6%)
              pyston fannkuch.py                        :    6.1s base3: 6.3 (-2.0%)
              pyston chaos.py                           :   16.8s base3: 17.5 (-4.2%)
              pyston fasta.py                           :    4.4s base3: 4.1 (+5.8%)
              pyston pidigits.py                        :    5.7s base3: 5.3 (+8.2%)
              pyston richards.py                        :    1.7s base3: 1.5 (+11.3%)
              pyston deltablue.py                       :    1.2s base3: 1.4 (-12.9%)
              pyston (geomean-10eb)                     :    4.2s base3: 4.3 (-1.4%)


```

because of the increased malloc traffic this PR should benefit a lot from jemalloc...


undingen:
with jemalloc:
best of 3 runs:
```
              pyston (calibration)                      :    1.0s base3_je: 1.0 (+0.3%)
              pyston django_template.py                 :    3.9s base3_je: 4.3 (-9.5%)
              pyston pyxl_bench.py                      :    3.6s base3_je: 3.6 (-1.2%)
              pyston sqlalchemy_imperative2.py          :    4.2s base3_je: 4.6 (-7.8%)
              pyston django_migrate.py                  :    1.4s base3_je: 1.6 (-8.3%)
              pyston virtualenv_bench.py                :    4.1s base3_je: 4.7 (-12.7%)
              pyston interp2.py                         :    3.6s base3_je: 4.1 (-12.2%)
              pyston raytrace.py                        :    5.5s base3_je: 5.6 (-1.9%)
              pyston nbody.py                           :    6.9s base3_je: 6.9 (-1.2%)
              pyston fannkuch.py                        :    6.1s base3_je: 6.3 (-2.4%)
              pyston chaos.py                           :   16.6s base3_je: 17.2 (-4.0%)
              pyston fasta.py                           :    4.4s base3_je: 4.3 (+1.4%)
              pyston pidigits.py                        :    5.7s base3_je: 5.2 (+9.2%)
              pyston richards.py                        :    1.6s base3_je: 1.5 (+11.5%)
              pyston deltablue.py                       :    1.1s base3_je: 1.3 (-15.9%)
              pyston (geomean-10eb)                     :    3.9s base3_je: 4.1 (-4.2%)
```

undingen:
with jemalloc:
first run:
```
              pyston (calibration)                      :    1.0s base1_je: 1.0 (+0.1%)
              pyston django_template.py                 :    4.7s base1_je: 6.4 (-27.0%)
              pyston pyxl_bench.py                      :    4.4s base1_je: 4.9 (-10.8%)
              pyston sqlalchemy_imperative2.py          :    5.0s base1_je: 6.2 (-19.1%)
              pyston django_migrate.py                  :    1.6s base1_je: 2.9 (-45.3%)
              pyston virtualenv_bench.py                :    4.7s base1_je: 7.7 (-39.1%)
              pyston interp2.py                         :    3.6s base1_je: 4.1 (-12.4%)
              pyston raytrace.py                        :    5.7s base1_je: 5.9 (-3.8%)
              pyston nbody.py                           :    7.2s base1_je: 7.1 (+1.5%)
              pyston fannkuch.py                        :    6.2s base1_je: 6.4 (-3.3%)
              pyston chaos.py                           :   17.0s base1_je: 17.6 (-3.6%)
              pyston fasta.py                           :    4.6s base1_je: 4.4 (+5.3%)
              pyston pidigits.py                        :    5.7s base1_je: 5.2 (+9.5%)
              pyston richards.py                        :    1.7s base1_je: 1.6 (+6.6%)
              pyston deltablue.py                       :    1.3s base1_je: 1.7 (-20.4%)
              pyston (geomean-10eb)                     :    4.3s base1_je: 4.9 (-13.2%)
```

undingen:
I tried to reduce the number of GC allocated mem (to decrease the number of collections) by creating a wrapper around the``` ASTInterpreter``` class to just store a pointer to the ```ASTInterpreter``` so we only have to GC allocate ```sizeof(Box) + (sizeof(ASTInterpreter*))``` but it didn't increase the performance. (Even though a large portion of the GC allocated mem are ASTInterpreter instances)

I also tried to align the jump targets to 16byte adresses by emitting multi byte nop sequences but this also didn't help. (Would have surprised me if it would have helped - don't think there is much to gain in this area)

src/asm_writing/assembler.cpp
```diff
@@ -790,6 +790,8 @@ void Assembler::cmp(Indirect mem, Register reg) {
 }
 
 void Assembler::lea(Indirect mem, Register reg) {
+    RELEASE_ASSERT(mem.base != RSP, "We have so generate the SIB byte...");
```
kmod: I think we also need a SIB byte for R12.

src/codegen/baseline_jit.h
```diff
@@ -0,0 +1,249 @@
...
+//    ; Side exit (e.g. if <val> <block1> else <block2>
+//      movabs $0x1270014108,%rcx   ; rcx = True
+//      cmp    %rax,%rcx            ; rax == True
+//      jne    end_side_exit
+//      mov    %rax,0x10(%rsp)      ;
+//      movabs $0x215bb60,%rax      ; rax = CFGBlock* to call next (rax is also the 1. return reg)
+//      mov    0x8(%rax),%rsi       ; load CFGBlock->code
+//      test   %rsi,%rsi            ; CFGBlock->code == 0
+//      je     epilog               ; exit to interpreter if code == 0
+//      jmpq   *0x8(%rax)           ; jump to new jit fragment (e.g second_JitFragment)
```
kmod: optimization idea -- could we initialize `CFGBlock->code` to be epilog instead of NULL?

kmod: or maybe we could start by emitting these as unconditional direct jumps to epilog, and then when we generate the new fragment, go back and update the jump target

just some ideas for future changes, I don't think we should try to do that in this PR :)

undingen: yeah this should work and would remove the extra checks. We could just emit once the epilog and share it between all JITed functions...
Currently several places check if the code is already JITed by checking if ```code``` is set. This checks could get removed and replaced by a different check e.g. if ```entry_code``` is set.

undingen: also ```emitReturn``` etc should probably just directly emit a ```leave``` and ```ret``` and not jump to the epilog because the jump takes more bytes and is an additional branch.

undingen:
The main issues I see currently:
- ugly additional members inside CFGBlock*
- memory won't get freed
- ```JitFragment``` and ```JitCodeBlock``` are not the best name. baseline jit is also not too great for such a simple jit tier
- integration with the IC/rewriter classes feels a little messy...

src/codegen/baseline_jit.h
```diff
@@ -0,0 +1,249 @@
...
+class SetItemIC;
+class GetItemIC;
+class CompareIC;
+class AugBinopIC;
+class UnaryopIC;
+class RuntimeCallIC;
+
+class JitFragment;
+
+// Manages a fixed size memory block which stores JITed code.
```
kmod: Minor request -- I think it'd be nice to start with a little bit more detail here about the overall baseline-JIT strategy and how it relates to the interpreter.  Just a quick high-level overview of what the classes in this file are for, how they get used, what our strategy is for going back and forth between the interpreter/jitted code.

src/codegen/irgen/hooks.cpp
```diff
@@ -127,7 +127,7 @@ EffortLevel initialEffort() {
         return EffortLevel::MAXIMAL;
     if (ENABLE_INTERPRETER)
         return EffortLevel::INTERPRETED;
-    return EffortLevel::MINIMAL;
```
kmod: I think this change makes sense, but it does reduce the number of llvm JIT tiers from 3 to 2; I tried doing that before since it seems silly to have 3 jit tiers that are all pretty similar, but I didn't manage to do it without hurting performance.  I think it had something to do with not having enough time to do proper type speculations, which I guess shouldn't be an issue right now, but I guess my point is that there certainly are going to be a lot of effects from adding a new tier and I'm not surprised if we're not seeing all the benefits right away :)

src/core/cfg.h
```diff
@@ -43,6 +44,12 @@ class CFGBlock {
     CFG* cfg;
 
 public:
+    // Baseline JIT helper fields:
+    // contains address to the start of the code of this basic block
+    void* code;
+    // contains the address of the entry function
+    std::pair<CFGBlock*, Box*>(*entry_code)(void* interpeter, CFGBlock* block);
```
kmod: yeah having these here isn't great, but I think it's ok for now.  I haven't read through all the code yet, but maybe we could some other data structure whose fields are `CFGBlock* block; void* code; ... *entry_code;`, and then add them to the CLFunction?

src/core/types.h
```diff
@@ -291,6 +293,7 @@ class SourceInfo {
     CFG* cfg;
     bool is_generator;
     std::string fn; // equivalent of code.co_filename
+    std::shared_ptr<LivenessAnalysis> liveness_info;
```
kmod: Hmm I do remember this being a memory hit, but I think it's ok for now.  I think we have some options going forward for how to potentially not need it, such as not going through the hashtable for the baseline jit, or even just generating the relevant liveness info when we generate the cfg.  or eventually doing the work of migrating to a real bytecode format :P

src/runtime/types.h
```diff
@@ -560,6 +560,10 @@ class BoxedTuple : public BoxVar {
         memmove(&rtn->elts[0], elts, sizeof(Box*) * nelts);
         return rtn;
     }
+    static BoxedTuple* create0() {
+        BoxedTuple* rtn = new (0) BoxedTuple(0);
+        return rtn;
+    }
```
kmod: This should probably just return EmptyTuple

src/codegen/ast_interpreter.cpp
```diff
@@ -1244,53 +1481,193 @@ Value ASTInterpreter::visit_name(AST_Name* node) {
                 raiseExcHelper(NameError, "free variable '%s' referenced before assignment in enclosing scope",
                                node->id.c_str());
             }
```
kmod: Could this all become a call to derefHelper?

src/codegen/ast_interpreter.h
```diff
@@ -33,6 +34,41 @@ struct LineInfo;
...
+    static Box* getBoxedLocalsHelper(void* interp);
+    static Box* getLocalHelper(void* interp, InternedString id);
+    static Box* landingpadHelper(void* interp);
+    static Box* setExcInfoHelper(void* interp, Box* type, Box* value, Box* traceback);
+    static Box* uncacheExcInfoHelper(void* interp);
+    static Box* yieldHelper(void* interp, Box* val);
+    static void setItemNameHelper(void* interp, Box* str, Box* val);
+    static void setLocalClosureHelper(void* interp, InternedString id, Box* v);
+    static void setLocalHelper(void* interp, InternedString id, Box* v);
+};
```
kmod: Not sure if I'm a huge fan of this :P  I can see the point though and maybe we can find a better solution later -- I think we'll probably see some pretty tight coupling between the baseline jit and the interpreter, so maybe it makes sense to make these members of AstInterpreter and then make that class visible to the baseline jit?

src/codegen/baseline_jit.h
```diff
@@ -0,0 +1,249 @@
...
+//      leave
+//      ret
+//
+class JitCodeBlock {
+private:
+    static constexpr int scratch_size = 256;
+    static constexpr int code_size = 4096 * 2;
+    static constexpr int epilog_size = 2; // size of [leave, ret] in bytes
+
+    void* code;
```
kmod: Maybe this could be a unique_ptr?  I know we're not freeing any of the memory yet, but delineating the ownership should still be a good thing.

src/codegen/baseline_jit.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+JitCodeBlock::JitCodeBlock(llvm::StringRef name)
+    : code(malloc(code_size)),
+      entry_offset(0),
+      epilog_offset(0),
+      a((uint8_t*)code, code_size - epilog_size),
+      iscurrently_tracing(false),
+      asm_failed(false) {
+    static StatCounter num_jit_code_blocks("num_baselinejit_code_blocks");
+    num_jit_code_blocks.log();
```
kmod: I think it's ok that we don't free the memory for now, since we don't free the rest of the codegen structures either :(  Maybe we could just stick another StatCounter here that counts the number of instruction bytes that got allocated?

src/codegen/baseline_jit.h
```diff
@@ -0,0 +1,249 @@
...
+
+public:
+    JitCodeBlock(llvm::StringRef name);
+
+    std::unique_ptr<JitFragment> newFragment(CFGBlock* block, int jump_offset = 0);
+    bool shouldCreateNewBlock() const { return asm_failed || a.bytesLeft() < 128; }
+    void fragmentFinished(int size, bool not_enough_space);
+};
+
+class JitFragment : public Rewriter {
```
kmod: Maybe we could call it "JitFragmentWriter"?

I'm not super psyched that this is a subclass of Rewriter -- that doesn't feel quite right to me.  I might have said that it has-a rewriter, but it looks like we use some of the internal APIs so that might not be easy.  I think there might be some fundamental differences in what they need to do, so I guess has-a isn't really right either.  Maybe we should (in the future) refactor out the common aspects of them (register allocation, action registration, etc) and then make them both subclasses of some common class?

kmod:
I'll read some more of the code in a sec, but overall my feedback is: yeah there's some stuff that is a little bit quick-and-dirty, but I think that's ok for now.  I'd say, let's just be good about leaving notes on the things that we think could be better, then work on getting this in and iterating.  I'm ok with this area of the code being a little bit rough to start since it feels like we'll be experimenting with it a lot.

src/codegen/baseline_jit.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+    // emit prolog
+    a.push(assembler::RBP);
+    a.mov(assembler::RSP, assembler::RBP);
+
+    static_assert(scratch_size % 16 == 0, "stack aligment code depends on this");
```
kmod: I think we need to make sure that the eh_frame we emit matches the stack size that we ask for -- I think the eh_frame_template hardcodes a certain amount of scratch (48 bytes I think).

src/codegen/ast_interpreter.cpp
```diff
@@ -360,25 +402,66 @@ Value ASTInterpreter::executeInner(ASTInterpreter& interpreter, CFGBlock* start_
...
         interpreter.next_block = 0;
 
+        if (ENABLE_BASELINEJIT && !interpreter.jit) {
+            CFGBlock* b = interpreter.current_block;
+            if (b->entry_code) {
+                should_jit = true;
+
+                try {
+                    UNAVOIDABLE_STAT_TIMER(t0, "us_timer_in_baseline_jitted_code");
+                    std::pair<CFGBlock*, Box*> rtn = b->entry_code(&interpreter, b);
```
kmod: Ah ok this seems like a good way of handling exceptions, but it also terrifies me :)  It means that we have to make sure the code we generate corresponds to the eh_frame; I think the relevant parts of that are 1) we have to make sure the stack size matches up, and 2) we can't use any callee-save registers.  I think the rewriter already takes care of the second point, though I was hoping that we would be able to get away from that.

src/codegen/baseline_jit.h
```diff
@@ -0,0 +1,249 @@
...
+
+public:
+    JitFragment(CFGBlock* block, ICSlotRewrite* rewrite, int code_offset, int epilog_offset, void* entry_code,
+                JitCodeBlock& code_block);
+
+    RewriterVar* imm(uint64_t val);
+    RewriterVar* imm(void* val);
+
+
+    RewriterVar* emitAugbinop(Value lhs, Value rhs, int op_type);
```
kmod: Super minor, but do we ever use the actual value part of the arguments, ie could we just pass the RewriterVar part of them?

src/codegen/baseline_jit.cpp
undingen: That's why I choose to use frame pointers because IMHO if I emit the EH info for a function with frame pointers it doesn't contain the amount of bytes the stack has to be adjusted but just that the stack pointer has to be restored by copying the base pointer.
At least that's what I thought whats happening by comparing different EH table output.

src/codegen/ast_interpreter.cpp
undingen: I think the stack size is fine as long as we use frame pointers.
But you are right we aren't currently allowed to modify any callee-save registers... (and they could help in reducing the code size of the JITed code e.g. by storing the ```&ASTInterpreter->current_inst``` in one)

undingen:
I addressed all your immediate comments and but some of the non JIT changes into a separate request (#664) and rebased.
I didn't squash it yet in order to make reviewing easier.

src/codegen/baseline_jit.cpp
kmod: oh, yeah you're right :)

src/codegen/ast_interpreter.cpp
```diff
@@ -429,199 +527,266 @@ void ASTInterpreter::doStore(AST_expr* node, Value value) {
...
+        return NULL;
+
+    static StatCounter ast_osrs("num_ast_osrs");
+    ast_osrs.log();
 
-            if (generator)
-                sorted_symbol_table[source_info->getInternedStrings().get(PASSED_GENERATOR_NAME)] = generator;
+    // TODO: we will immediately want the liveness info again in the jit, we should pass
+    // it through.
+    std::unique_ptr<LivenessAnalysis> liveness = computeLivenessInfo(source_info->cfg);
```
kmod: Could this be getLiveness()?  I think the call to computeLivenessInfo() in irgen.cpp could also be changed to access the cached version.

src/codegen/ast_interpreter.cpp
```diff
@@ -630,6 +795,7 @@ Value ASTInterpreter::visit_invoke(AST_Invoke* node) {
         v = visit_stmt(node->stmt);
         next_block = node->normal_dest;
```
kmod: Should we call finishJITing here?  ie am I reading this right -- in the invoke case, we will jit across basic blocks?  Which seems pretty reasonable but I'm not sure if it's intentional or not.

undingen: oh yeah, there should be a finish call. I will add it.
I think even without it it should work fine but if the next basic block contains an instruction we don't support yet we will throw away more than necessary

src/core/types.h
undingen: I will change that one to a ```unique_ptr``` because we don't use ```shared_ptr``` at every use and having ```SourceInfo``` as the owner should be fine.

undingen: oh now I remember why I didn't use a ```unique_ptr``` it doesn't work with forward declared types...

undingen:
addressed your comments in the last commit

kmod:
I think the issue with unique_ptr's and forward-declared types is that the compiler will try to generate a default destructor (for the containing class) which won't work.  You can get around this by declaring a destructor (ie `~SourceInfo();``); the fun part is you might not have to define it since it probably doesn't get called anywhere :P

kmod:
I think we might need to be careful about reentrancy -- what if we are in the baseline jit, and we hit a function call that ends up triggering the baseline jit on the same function?  I haven't finished reading through the code, but assuming there isn't anything that explicitly handles this, I think there could be issues.  For instance, the JitFragmentWriter will save the `code_offset` at the beginning of the jit, but it will end up writing the assembly to whatever the current offset is in the code block.  I think this could mess up the offsets in the instructions, and also we will write the original code_offset into the CFGBlock even though some other assembly will live there :(

I think it might be enough to say "if the current codeblock offset at the end is not the same as when we started, abort the fragment"?  But I'm not sure if that's necessary (or enough).

undingen:
The ```JitCodeBlock``` has a ```iscurrently_tracing``` (which should probably be renamed) member which makes sure there is always only one JIT fragment active per JitCodeBlock.
But with multi threading this will need some lock...

kmod:
Ah ok, nice :)  Could we still have an assert in fragmentFinished/finishCompilation that asserts the fragmentwriter and the codeblock agree on where the code will end up?

Also I think it's safe to assume we're not getting rid of the GIL any time soon.  Probably not until after Windows and Python 3 support :/

undingen:
I will continue tomorrow with eliminating the shared_ptr and adding the check.
Can I stash it then or do you prefer to stash it later because you want to review it in more detail?

kmod:
ok finally read through it all :)  looks good to me; feel free to rebase/squash and we should be able to get this merged in tomorrow!

undingen:
I made the last changes and squashed. There should be no functional changes besides that I enabled again the callattr IC because I don't get the protobuf error any more. I don't think the issue is fixed but I also don't think it's hiding inside the JIT (more likely Runtime ICs/rewritter) if we encounter it again I will try to track it down.
I also added a new commit with a small micro optimization (change ASTInterpreter field offsets).
Should be ready for merging :-)

kmod:
Awesome!  having a baseline JIT will change a lot of things :)

undingen:
Cool :-)

PR #652, opened 2015-06-28
For issue #254, add lt, le, gt, ge to old-style class

Daetalus:
The first contribution target [#245](https://github.com/dropbox/pyston/issues/254). Add lt/le/gt/ge to old-style class. Pass the tests and format check.

### DIFF
--- a/src/runtime/classobj.cpp
+++ b/src/runtime/classobj.cpp
@@ -741,6 +741,22 @@ Box* _instanceBinary(Box* _inst, Box* other, const char* attr) {
     return runtimeCall(func, ArgPassSpec(1), other, NULL, NULL, NULL, NULL);
 }
 
+Box* instanceGt(Box* _inst, Box* other) {
+    return _instanceBinary(_inst, other, "__gt__");
+}
+
+Box* instanceGe(Box* _inst, Box* other) {
+    return _instanceBinary(_inst, other, "__ge__");
+}
+
+Box* instanceLt(Box* _inst, Box* other) {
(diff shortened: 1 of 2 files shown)


kmod:
Code looks good to me, but can we add some tests?  Looks like they could go with the `__eq__` and `__ne__` tests in test/tests/oldstyle_classes.py; just adding custom comparisons and showing that they get called should be good enough.  thanks!

kmod:
Actually sorry, can I request a couple other tests -- can we also make sure that we will try the reversed comparisons appropriately.  ie if a class only defines `__lt__`, but appears on the right-hand-side of a greater-than comparison, we should try calling the `__lt__` method.

Daetalus:
I will try to add some tests for testing all these four functions. Thanks for your comments!

kmod:
This is super minor, but in the test, could we have the functions print out the name of the function when they get called?  (ie `__lt__` could have something like `print "lt"`)  The way to resolve comparisons is pretty elaborate, and it's nice to verify the internal behavior if we can.  For example, in something else I copy-pasted and updated one "ge" to "gt" -- but missed the other one, and the tests just happened to pass since I didn't test the equality case.  I know this is nitpicky, but it's a good habit, and it will be relevant for the augassign stuff as well :)

Daetalus:
Ok, thanks for your comment! I should notice it earlier, because I already saw some `print` in test function, but didn't get enough attention. I will update it soon.

kmod:
Awesome, thanks!

PR #653, opened 2015-06-29
Don't guard types sometimes

tjhance:
We were doing some unnecessary guards, like guarding the types of constant objects. Fixing this saves about 10 bytes on average per IC, and has a comparable small improvement in performance:

```
       django_template.py             5.1s (6)             5.1s (8)  -1.6%
            pyxl_bench.py             4.2s (6)             4.1s (6)  -2.4%
 sqlalchemy_imperative.py             2.1s (6)             2.0s (6)  -2.5%
        django_migrate.py             2.0s (6)             1.9s (6)  -1.9%
      virtualenv_bench.py             8.4s (6)             8.3s (6)  -0.5%
                  geomean                 3.8s                 3.7s  -1.8%
```

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -719,7 +719,10 @@ Box* Box::getattr(llvm::StringRef attr, GetattrRewriteArgs* rewrite_args) {
                     REWRITE_ABORTED("");
                     rewrite_args = NULL;
                 } else {
-                    rewrite_args->obj->addAttrGuard(cls->attrs_offset + offsetof(HCAttrs, hcls), (intptr_t)hcls);
+                    if (!(rewrite_args->obj->isConstant() && cls == type_cls
+                          && static_cast<BoxedClass*>(this)->is_constant)) {
+                        rewrite_args->obj->addAttrGuard(cls->attrs_offset + offsetof(HCAttrs, hcls), (intptr_t)hcls);
+                    }
                     if (hcls->type == HiddenClass::SINGLETON)
                         hcls->addDependence(rewrite_args->rewriter);
                 }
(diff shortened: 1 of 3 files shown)


tjhance:
Hm so the build is failing, and the culprit appears to be this change:

```
+ if (!cls->is_constant) {
      rewrite_args->obj->addAttrGuard(cls->attrs_offset + HCATTRS_HCLS_OFFSET, (intptr_t)hcls);
+ }
```

The goal was to not guard on the hidden class of builtin types (you can't set attributes on them). I figure if a class is constant, the hidden class shouldn't ever change. I guess this assumption is wrong. Any idea why?

tjhance:
oh, do we change the hidden class to be `DICT_BACKED` in the event that someone accesses the `__dict__`? That's a shame :\

tjhance:
or does that only happen when someone sets the __dict__? (in which case my previous comment wouldn't apply)

tjhance:
OK I think I fixed the problem, I was checking `is_constant` on the wrong thing.

The perf improvement is not as large (but who knows, really, because noise). The IC size improvement is still significant, so I think there will be an even better improvement when we actually get around to making the IC slot sizes smaller.

src/runtime/objmodel.cpp
```diff
@@ -961,6 +964,7 @@ Box* typeLookup(BoxedClass* cls, llvm::StringRef attr, GetattrRewriteArgs* rewri
                 assert(rewrite_args->obj == obj_saved);
             } else {
                 rewrite_args->obj = rewrite_args->rewriter->loadConst((intptr_t)base, Location::any());
+                rewrite_args->obj_cls_guarded = true;
```
kmod: Is this true?  I think you might be allowed to to change the metaclass of a base class...  Maybe this could be `obj_cls_guarded = base->cls->is_constant`?

kmod:
I think this is a bit confusing since this PR is based off a commit (adding "-a") that got rewritten and then merged; would you mind rebasing?  I think this applies to some of your other PRs as well (#654, #656, #659).  otherwise lgtm!

tjhance:
I rebased and added a check that the base class is a constant class.

Can you double check that all my assumptions look sane and then merge this in?

PR #654, opened 2015-06-29
Don't do non-visible actions whose results are never used

tjhance:
Not too sure if this is worth it. There are definitely actions (like getattrs) whose results are never used, so this does PR does improve the assembly size by a tiny bit. The perf change, if any, is small enough that it's difficult to tell if it's worth it. Just posting in case you guys have thoughts.


kmod:
Sounds good to me.  I think I saw that you added some stats for total-rewrite-bytes -- I think if that improves, it's worth checking in the change even if the overall perf impact isn't that big.  I guess I think these changes are useful even if the overall effect for now is low, since 1) the other sections of time will go down, 2) once Marius's thing gets in we'll be using the rewriter a lot more, and 3) we'll be able to trim the size of patchpoints which should give an additional benefit.

I also have some project ideas in other areas though; let's chat after the meeting tomorrow?

PR #656, opened 2015-06-29
Fix lea

tjhance:
It turns out we weren't doing the `lea` thing as often as we could because we were releasing the constant RewriterVars too early.

### DIFF
--- a/src/asm_writing/rewriter.cpp
+++ b/src/asm_writing/rewriter.cpp
@@ -593,6 +593,8 @@ assembler::Register RewriterVar::getInReg(Location dest, bool allow_constant_in_
                 assembler::Register dest_reg = dest.asRegister();
                 assert(dest_reg != reg); // should have been caught by the previous case
 
+                rewriter->allocReg(dest);
+
                 rewriter->assembler->mov(reg, dest_reg);
                 rewriter->addLocationToVar(this, dest_reg);
                 return dest_reg;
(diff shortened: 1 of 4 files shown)


undingen:
Not sure if this is the travis ci problem: I think I implemented the LEA inst encoding wrong.
At least [RSP+displ] is wrong because it misses the SIB byte...

tjhance:
Thanks for the heads up! We'll see if that fixes it

tjhance:
Actually it probably won't, since this only applies to RSP and R12. Oh well :\

tjhance:
Okay I think I got it working (let's see if the travis build passes)

Here's what I did (some of this is from my ramblings on gitter)

I finally realized I need to focus on initializePatchpoint3 since that’s where all the spilling and such is. So it seems like the live_outs fall into 3 camps: the callee save registers, the regs_to_spill, and the regs_to_reload.

So the first camp, callee save registers, is easy, since we basically don’t touch those.

The regs_to_reload are also easy. Since the continue_addr is now before these registers get unspilled, I can just remove them from the live_outs, as we discussed

I’m unsure about the regs_to_spill camp; it seems like they have to go in the live_outs (the “batch push” which spills them doesn’t run before the IC does, and even if it did, it uses the scratch space, the same space as the ICs) which means we still need to add some restoreLiveOuts (analagous to restoreArgs, I guess) logic.

So I added the logic to restore live outs in `restoreArgs`. Initially, this had the problem that the subsequent `mov`s would just un-move the live outs.

I figure that since `mov`s don't affect the flags, it should be fine to put `restoreArgs` in between the cmp and the jump. In the future, though, we could be smarter about register allocation.

kmod:
lgtm :)  I agree that we might want to move to spilling more stuff globally, but it's not clear since I think the getattr/setattr/getGlobal ics are small enough that they might be able to get away with very little spilling.

kmod:
doh this had a non-textual merge conflict with another PR I just merged; I'll fix it.

PR #659, opened 2015-06-30
Use less large constants

tjhance:
The incl, decl, and call can use the smaller instructions sometimes instead of loading large constants.

I also now allocate the slot before emitting assembly, to make it easier to emit assembly which depends on its address. (This was an easy change to make, since we emit all the assembly in the commit, anyway.)

Average IC size goes down from 143 bytes to 130 bytes. 

```
       django_template.py            5.1s (14)             5.1s (4)  -0.8%
            pyxl_bench.py            4.2s (14)             4.1s (4)  -2.4%
 sqlalchemy_imperative.py            2.1s (14)             2.1s (4)  -0.4%
        django_migrate.py            1.9s (14)             1.9s (4)  -1.1%
      virtualenv_bench.py            8.3s (14)             8.3s (4)  -0.1%
                  geomean                 3.7s                 3.7s  -0.9%
```

### DIFF
--- a/src/asm_writing/assembler.cpp
+++ b/src/asm_writing/assembler.cpp
@@ -670,7 +670,24 @@ void Assembler::decl(Indirect mem) {
     }
 }
 
+void Assembler::incl(Immediate imm) {
+    emitByte(0xff);
+    emitByte(0x04);
+    emitByte(0x25);
+    emitInt(imm.val, 4);
+}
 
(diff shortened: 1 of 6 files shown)


kmod:
lgtm but could use a rebase :)

tjhance:
rebased

undingen:
cool, the new jit tier is hitting a lot the new shorter call sequence.

PR #660, opened 2015-06-30
gc less frequently, and make collections faster

toshok:
This is not quite read for merging.  a little refactoring needs to be done.

---

Found this while investigating slowdown/speedup of benchmarks due to the run allocator work.

This change adds two knobs to configure gc frequency - heap occupancy percentage, and time
between collections.

The idea is that the collector tries to collect no more frequently than every QUICK_GC_FREQUENCY(500)
milliseconds.  If it NUM_QUICK_GC(2) collections in a row, it increases the size of the heap.

We also resize the heap if after a collection memory usage is above a certain threshold (right now it's
the same percentage as what triggered the collection.  should probably tune this more when/if we get
benchmarks when a lot of live objects after a collection - right now none of them actually hit this.)

further work might be to shrink the heap if collections start taking too much time.

### DIFF
--- a/src/core/stats.cpp
+++ b/src/core/stats.cpp
@@ -102,9 +102,6 @@ void Stats::clear() {
 }
 
 void Stats::startEstimatingCPUFreq() {
-    if (!Stats::enabled)
-        return;
-
     clock_gettime(CLOCK_REALTIME, &Stats::start_ts);
     Stats::start_tick = getCPUTicks();
 }
--- a/src/gc/collector.cpp
+++ b/src/gc/collector.cpp
@@ -96,6 +96,7 @@ public:
 
     void push(void* p) {
         GC_TRACE_LOG("Pushing %p\n", p);
+        assert(((intptr_t)p) % 8 == 0);
         GCAllocation* al = GCAllocation::fromUserData(p);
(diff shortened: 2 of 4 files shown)


toshok:
```
                           88eedd39ce0745f6e0:  1d0f49400582eeb2c8:
       django_template.py             4.7s (2)             4.4s (2)  -5.3%
            pyxl_bench.py             3.9s (2)             4.0s (2)  +1.1%
sqlalchemy_imperative2.py             5.1s (2)             5.0s (2)  -0.7%
        django_migrate.py             1.8s (2)             1.8s (2)  -1.0%
      virtualenv_bench.py             7.8s (2)             7.6s (2)  -2.5%
                  geomean                 4.2s                 4.1s  -1.7%
```

the pyxl_bench.py behavior seems to be due to the increased work per collection to build a quickly searchable vector of large objects.  of the three single process benchmarks, django_template has most live large objects at the start of each gc (going as high as 12000 per gc toward the end).  next is sqlalchemy_imperative2, with ~300 per gc.  and next is pyxl_bench.py, with ~100-200.

I'll investigate turning off the vector lookup for those cases where we can identify the number of large objects alive (we should be able to do this.)

PR #662, opened 2015-06-30
Some refactors in GC code + class-freed-before-instance bug fix.

rudi-c:
This version of the bug fix handle cases like metaclasses.

No significant performance change.
```
pyston (calibration)                      :    1.0s baseline: 1.0 (-0.4%)
pyston django_template.py                 :    4.6s baseline: 4.6 (+0.1%)
pyston pyxl_bench.py                      :    3.7s baseline: 3.7 (+1.0%)
pyston sqlalchemy_imperative2.py          :    5.1s baseline: 5.1 (-0.9%)
pyston django_migrate.py                  :    1.7s baseline: 1.7 (-0.4%)
pyston virtualenv_bench.py                :    5.2s baseline: 5.1 (+0.6%)
pyston interp2.py                         :    3.7s baseline: 3.7 (+0.5%)
pyston raytrace.py                        :    5.4s baseline: 5.5 (-1.6%)
```


rudi-c:
Had to change the solution for class-freed-before-instance because the class was being kept around but things it referenced (like `attr_list`) were not. Travis CI didn't catch this because it ran in release mode and we got lucky that accessing invalid memory didn't cause a crash.

The new solution I found keeps the class around for only one more collection which has minimal performance hit.
```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+1.6%)
              pyston django_template.py                 :    4.5s baseline: 4.5 (-0.0%)
              pyston pyxl_bench.py                      :    3.7s baseline: 3.6 (+1.0%)
              pyston sqlalchemy_imperative2.py          :    5.1s baseline: 5.1 (+0.3%)
              pyston django_migrate.py                  :    1.7s baseline: 1.7 (+0.6%)
              pyston virtualenv_bench.py                :    5.1s baseline: 5.1 (+0.1%)
```

src/gc/collector.cpp
```diff
@@ -48,6 +48,31 @@ namespace gc {
 FILE* trace_fp;
 #endif
 
+unsigned classesAllocatedSinceCollection = 0;
```
toshok: seems like the classesAllocatedSinceCollection stuff snuck into this PR?  I can't see a use of it in this change.

src/gc/collector.cpp
```diff
@@ -333,50 +375,62 @@ void markPhase() {
...
+    trace_fp = fopen("gc_trace.txt", "w");
+#endif
+#endif
+    GC_TRACE_LOG("Starting collection %d\n", ncollections);
+
+    GC_TRACE_LOG("Looking at roots\n");
+    TraceStack stack(roots);
+    GCVisitor visitor(&stack);
+
+    getMarkPhaseRoots(visitor);
```
toshok: bikeshedding: this doesn't really "get" the roots - it marks them.

src/gc/heap.cpp
```diff
@@ -118,13 +117,20 @@ void _bytesAllocatedTripped() {
         return;
 
     threading::GLPromoteRegion _lock;
+
     runCollection();
 }
 
+__attribute__((always_inline)) void markClass(GCVisitor& visitor, GCAllocation* al) {
```
toshok: this doesn't appear to be used?

src/gc/collector.cpp
```diff
@@ -115,8 +140,12 @@ class TraceStack {
             pop_chunk();
             assert(cur == end);
             return *--cur; // no need for any bounds checks here since we're guaranteed we're CHUNK_SIZE from the start
+        } else {
+            // We emptied the stack, but we should prepare a new chunk in case another item
```
toshok: does this actually help?  it guarantees we have one additional `get_chunk()` call at the end of the last use of the trace stack.

src/gc/collector.cpp
```diff
@@ -283,30 +302,47 @@ void GCVisitor::visitPotentialRange(void* const* start, void* const* end) {
...
 
-static int ncollections = 0;
-
-void markPhase() {
-#ifndef NVALGRIND
-    // Have valgrind close its eyes while we do the conservative stack and data scanning,
-    // since we'll be looking at potentially-uninitialized values:
-    VALGRIND_DISABLE_ERROR_REPORTING;
-#endif
+static inline void visitByGCKind(void* p, GCVisitor& visitor) {
```
toshok: I do worry that the compiler will choose at some point in the future to *not* inline function, even with additional specifiers/attributes/etc.  Not sure how we can ensure that without some post processing, though :/

For now I've verified that it does get inlined, and we can deal with the non-inline case later if it pops up (since I'm pretty sure it'll cause a perf regression if it does.)

toshok:
otherwise lgtm

src/gc/collector.cpp
rudi-c: Yeah, this isn't needed anymore.

src/gc/heap.cpp
rudi-c: Similarly.

src/gc/collector.cpp
rudi-c: Yes, because in the mark phase I added an extra step where some classes get added to the TraceStack after one traversal, and then do another traversal, so I'm reusing the same TraceStack. I could also create another TraceStack, but this makes the TraceStack implementation more "correct" and it was pretty hard to figure out that the problem was in the TraceStack because it would manifest as memory corruption elsewhere in the program.

src/gc/collector.cpp
rudi-c: Yeah, might as well use alwaysinline.

src/gc/collector.cpp
toshok: Ah yes.  The TraceStack was very much operating under the assumption that after the final pop it was done.

rudi-c:
Rebased -- ready for review again.

src/gc/collector.cpp
```diff
@@ -48,6 +39,29 @@ namespace gc {
 FILE* trace_fp;
 #endif
 
+static std::unordered_set<void*> roots;
+static std::vector<std::pair<void*, void*>> potential_root_ranges;
+
+// BoxedClasses in the program that are still needed.
+static std::unordered_set<BoxedClass*> class_objects;
+
+static std::unordered_set<void*> nonheap_roots;
```
kmod: random thought: maybe at some point we could change these to use llvm::DenseSet

src/gc/collector.cpp
```diff
@@ -198,7 +207,7 @@ bool isValidGCObject(void* p) {
     return al->user_data == p && (al->kind_id == GCKind::CONSERVATIVE_PYTHON || al->kind_id == GCKind::PYTHON);
 }
 
-void setIsPythonObject(Box* b) {
+void registerPythonObject(Box* b) {
```
kmod: just fyi that we don't call this for all python objects.

src/gc/collector.cpp
```diff
@@ -333,50 +357,62 @@ void markPhase() {
...
+            classes_to_remove.push_back(cls);
+            visitor.visit(cls);
         }
     }
 
+    graphTraversalMarking(stack, visitor);
+
+    for (BoxedClass* cls : classes_to_remove) {
+        class_objects.erase(cls);
+    }
```
kmod: Hmm if I'm understanding this, if a class and metaclass become garbage in the same collection, we'll defer them both until the next collection at which point we'll still free them at the same time :/

kmod:
mostly looks good to me, but I'm not sure about the class-freeing thing.  it might be an easy fix; it might just be something like an additional
```C++
for (BoxedClass* cls : classes_to_remove) class_objects.insert(cls->cls);
```
but I'm not sure :/  I don't think we've been able to even reproduce this issue, so maybe it makes sense to separate the class-freeing work from the finalizer work?

src/gc/collector.cpp
rudi-c: Ah you're right. I just need to split visiting unreachable classes and adding them to the remove-list in separate loops.

rudi-c: Wait no that's not true, it'll just retain the classes forever. Your solution seems better.

src/gc/collector.cpp
rudi-c: Yeah. For the class objects, it's possible that this whole class-freeing problem goes away with a moving collector (which won't have a sweep phase in which to look at classes) and in most programs the sets won't contain more than a few hundred objects so I'll put that off for now.

src/gc/collector.cpp
rudi-c: I'm adding an assert in `DEFAULT_CLASS_SIMPLE` for class objects which you mention in #672

rudi-c:
Ok I rebased and added an assert for the `registerPythonObject` comment and fixed the metaclass edge case.

rudi-c:
Erm, timeout?

toshok:
yeah that timeout problem is worrying.  happened randomly for other (not gc touching, not perf impacting) PR's too.  are we just sharing hardware with other travis-ci vms that are eating up all the cpu?

toshok:
:+1:

PR #666, opened 2015-07-02
Preparation for finalizer code, refactoring some simple_destructor logic.

rudi-c:
Mostly moving simple_destructor functions into tp_dealloc instead of their own field. Replace the field with a boolean to indicate whether the tp_dealloc has the safety properties of a simple_destructor. Assign simple_destructor to more builtin types.

Depends on gcfinalizers2 PR.

### DIFF
--- a/from_cpython/Modules/itertoolsmodule.c
+++ b/from_cpython/Modules/itertoolsmodule.c
@@ -4080,6 +4080,34 @@ static PyMethodDef module_methods[] = {
     {NULL,              NULL}           /* sentinel */
 };
 
+// Pyston change: These are types defined in this file that I manually
+// checked. The ones that aren't commented out have a `tp_dealloc` that
+// doesn't do anything in Pyston as we switched to garbage collection and
+// the finalizer logic in Pyston wants to know that for optimization purposes.
+PyTypeObject* Itertool_SafeDealloc_Types[] = {
+    // &combinations_type,
+    // &cwr_type,
+    &cycle_type,
+    &dropwhile_type,
+    &takewhile_type,
+    &islice_type,
+    &starmap_type,
+    &imap_type,
+    &chain_type,
+    &compress_type,
+    &ifilter_type,
+    &ifilterfalse_type,
+    &count_type,
+    &izip_type,
+    &iziplongest_type,
+    // &permutations_type,
+    // &product_type,
+    &repeat_type,
+    &groupby_type,
+    NULL
+};
+
+
 PyMODINIT_FUNC
 inititertools(void)
 {
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -3132,6 +3132,10 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
     // this should get automatically initialized to 0 on this path:
     assert(cls->attrs_offset == 0);
 
+    if (Py_TPFLAGS_BASE_EXC_SUBCLASS & cls->tp_flags) {
+        exception_types.push_back(cls);
+    }
+
     return 0;
 }
 
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -1152,11 +1152,6 @@ static void proxy_to_tp_traverse(GCVisitor* v, Box* b) {
     b->cls->tp_traverse(b, call_gc_visit, v);
 }
 
-static void proxy_to_tp_clear(Box* b) {
-    assert(b->cls->tp_clear);
-    b->cls->tp_clear(b);
-}
-
 // This probably belongs in tuple.cpp?
 extern "C" void tupleGCHandler(GCVisitor* v, Box* b) {
     boxGCHandler(v, b);
@@ -2932,15 +2927,11 @@ extern "C" PyUnicodeObject* _PyUnicode_New(Py_ssize_t length) noexcept {
 }
 
 // Normally we don't call the Python tp_ slots that are present to support
-// CPython's reference-counted garbage collection. However, for a few types
-// like weak references, we still rely on reference-counting to some extent.
+// CPython's reference-counted garbage collection.
 static void setTypeGCProxy(BoxedClass* cls) {
     cls->tp_alloc = PystonType_GenericAlloc;
+    cls->tp_free = default_free;
     cls->gc_visit = proxy_to_tp_traverse;
-
-    // We can't use the original tp_dealloc here, the dealloc method of some
-    // types like ProxyType explicitely frees itself without using tp_free.
-    cls->tp_dealloc = proxy_to_tp_clear;
     cls->has_safe_tp_dealloc = true;
     cls->is_pyston_class = true;
 }
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -94,6 +94,8 @@ void setupGC();
 
 bool IN_SHUTDOWN = false;
 
+std::vector<BoxedClass*> exception_types;
+
 void FrameInfo::gcVisit(GCVisitor* visitor) {
     visitor->visit(boxedLocals);
     visitor->visit(exc.traceback);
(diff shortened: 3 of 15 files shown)


rudi-c:
```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.7%)
              pyston django_template.py                 :    4.5s baseline: 4.5 (-0.2%)
              pyston pyxl_bench.py                      :    3.7s baseline: 3.6 (+1.2%)
              pyston sqlalchemy_imperative2.py          :    5.1s baseline: 5.1 (+0.6%)
              pyston django_migrate.py                  :    1.7s baseline: 1.7 (-0.3%)
              pyston virtualenv_bench.py                :    5.2s baseline: 5.2 (+0.3%)
```

rudi-c:
Oh nice I have PR 666 :smiling_imp:

from_cpython/Modules/itertoolsmodule.c
```diff
@@ -4080,6 +4080,30 @@ static PyMethodDef module_methods[] = {
     {NULL,              NULL}           /* sentinel */
 };
 
+PyTypeObject* Itertool_SafeDealloc_Types[] = {
+    // &combinations_type,
+    // &cwr_type,
```
toshok: are these commented out because they're not actually safe?

src/runtime/types.cpp
```diff
@@ -2910,6 +2912,51 @@ extern "C" PyUnicodeObject* _PyUnicode_New(Py_ssize_t length) noexcept {
     return unicode;
 }
 
+static void setTypeGCProxy(BoxedClass* cls) {
```
toshok: comments as to the purpose of these two functions (particularly `setTypeGCNone` would be useful)

src/capi/typeobject.cpp
```diff
@@ -2993,6 +2993,17 @@ void commonClassSetup(BoxedClass* cls) {
     assert(cls->tp_dict && cls->tp_dict->cls == attrwrapper_cls);
 }
 
+static void checkIfExceptionType(BoxedClass* cls) {
```
toshok: is there a constant time check we can perform here using `Py_TPFLAGS_BASE_EXC_SUBCLASS` ?  I'm not sure if we do the right setup 100% of the time for it, but if we do, it should be quite a bit quicker.

src/gc/collector.cpp
```diff
@@ -333,50 +372,62 @@ void markPhase() {
...
+    markRoots(visitor);
+
+    graphTraversalMarking(stack, visitor);
+
+    // Some classes might be unreachable. Unfortunately, we have to keep them around for
+    // one more collection, because during the sweep phase, instances of unreachable
+    // classes might still end up looking at the class. So we visit those unreachable
+    // classes and remove it from the list of class objects so that it can be freed
+    // in the next collection.
+    std::vector<BoxedClass*> classes_to_remove;
```
toshok: it might be more efficient to create a new `class_objects` vector here (maybe reserve the same size as the old `class_objects` vector to save allocations), and only push_back the classes onto the new one if they're marked.

src/gc/collector.cpp
```diff
@@ -333,50 +372,62 @@ void markPhase() {
...
+        GCAllocation* al = GCAllocation::fromUserData(cls);
+        if (!isMarked(al)) {
+            classes_to_remove.push_back(cls);
+            visitor.visit(cls);
         }
     }
 
+    graphTraversalMarking(stack, visitor);
+
+    for (BoxedClass* cls : classes_to_remove) {
```
toshok: which would let us get rid of this loop and all the copying.

src/gc/collector.cpp
```diff
@@ -283,30 +299,47 @@ void GCVisitor::visitPotentialRange(void* const* start, void* const* end) {
...
-    GC_TRACE_LOG("Looking at roots\n");
-    TraceStack stack(roots);
-    GCVisitor visitor(&stack);
+    GCKind kind_id = al->kind_id;
+    if (kind_id == GCKind::UNTRACKED) {
+        // Nothing to do here.
+    } else if (kind_id == GCKind::CONSERVATIVE || kind_id == GCKind::CONSERVATIVE_PYTHON) {
+        uint32_t bytes = al->kind_data;
+        if (DEBUG >= 2) {
+            global_heap.assertSmallArenaContains(p, bytes);
```
toshok: won't this cause usually cause assertions?  long objects do large conservative allocations, as will densemaps/vectors/etc.

from_cpython/Modules/itertoolsmodule.c
rudi-c: I copied the static array from `PyTypeObject *typelist[]` that has all the types in the file, then commented out those that I wasn't sure was safe yeah.

src/gc/collector.cpp
rudi-c: In most applications there should be no classes to remove, I think it'd be more work to copy all the existing classes every collection (but either way it should be a negligible cost).

src/gc/collector.cpp
rudi-c: This can actually be removed, `DEBUG` is never `>= 2` in `collector.cpp`

from_cpython/Modules/itertoolsmodule.c
toshok: definitely add that to the comment (maybe with a `TODO` to take a further look).  confusing otherwise.

rudi-c:
Updated -- ready for review again. That `threading_local` test still fails but that's fine right?

kmod:
lgtm (pending #662) though I think it'd be nice to address proxy_to_tp_clear.  I kicked travis-ci so it's retesting and should hopefully not hit that threading_local issue a second time :)

rudi-c:
Ok, I replaced that to use the default `tp_dealloc` (which does the same thing as `tp_clear` for weakrefs in addition to a few CPython GC-things that we don't run).

rudi-c:
Actually no that's not true. `tp_dealloc` and `tp_clear` for proxy are different and we can't use `tp_dealloc`, I'm reverting that.

toshok:
what does "address proxy_to_tp_clear" mean?  missing context :)

rudi-c:
Replacing `tp_dealloc = proxy_to_tp_clear` with just the original `tp_dealloc`. I don't think that works though.

toshok:
hm, I think it will if `tp_free` is set to `default_free`.

the weakref gc_dealloc is just:

```
    PyObject_GC_UnTrack(self);
    clear_weakref((PyWeakReference *) self);
    Py_TYPE(self)->tp_free(self);
```

`gc_clear` only does the `clear_weakref` call.

The wildcard is `PyObject_GC_UnTrack`.  that's either a macro that does nothing or a function in `gcmodule.c` that definitely won't work for us.

rudi-c:
That one is fine, I tried setting `tp_del`. The problem is:
```
static void                                       
proxy_dealloc(PyWeakReference *self)              
{                                                 
    if (self->wr_callback != NULL)                
        PyObject_GC_UnTrack((PyObject *)self);    
    clear_weakref(self);                          
    PyObject_GC_Del(self);                        
}                                                 
```

src/runtime/types.cpp
```diff
@@ -2929,6 +2931,56 @@ extern "C" PyUnicodeObject* _PyUnicode_New(Py_ssize_t length) noexcept {
     return unicode;
 }
 
+// Normally we don't call the Python tp_ slots that are present to support
+// CPython's reference-counted garbage collection. However, for a few types
+// like weak references, we still rely on reference-counting to some extent.
+static void setTypeGCProxy(BoxedClass* cls) {
+    cls->tp_alloc = PystonType_GenericAlloc;
+    cls->gc_visit = proxy_to_tp_traverse;
+    cls->tp_dealloc = proxy_to_tp_clear;
```
toshok: might be good to add a comment here "we can't use the normal tp_dealloc here, since proxy_dealloc explicitly frees self but doesn't go through tp_free, etc..." so we don't forget and ask the questions again :)

kmod:
I think we're a bit off-track on the proxy_to_tp_clear issue -- tp_clear is the wrong attribute for us to be relying on, and tp_dealloc is the right one.  There's nothing particularly special about the proxy_dealloc method, and while it's ok for us to make some restrictions on the c extension code we'll support, it seems like this is a pretty bad workaround for a case that we need to support anyway.

toshok:
nod.  I mean we can just change the proxy tp_dealloc to do the same tp_free invocation that weakref_dealloc does (and set tp_free to default_free).  The only reason I didn't feel too strongly either way about this is that this particular method is already very specialized and only used for the weakref types.

is there any reason we couldn't just change PyObject_GC_Del(o) to go through tp_free?

toshok:
@rudi-c okay, so steps forward from here are, I think:

1. verify that PyObject_GC_Del is actually hitting gc_compat_free (I can't remember if that's dependent on #defines we aren't specifying..  maybe it's a no-op for us?)
2. depending on 1, either leave PyObject_GC_Del in, comment it out, or change to the tp_free call like weakref_dealloc uses (the latter will take some additional work since you'll have to change alloc sites to be proper).  Then likely get rid of proxy_to_tp_clear + setTypeGCProxy, and switch the weakref types to using the same method as the other builtins.
2. check other extension tp_deallocs (at least in our builtin extension modules) to make sure they aren't calling PyObject_GC_Del and are otherwise safe.
3. add an issue for auditing the tp_deallocs across all the source we need to run to make sure they aren't doing anything funny.

sound good?

toshok:
:+1:

PR #667, opened 2015-07-02
gc less frequently

toshok:
simpler logic for determining gc frequency, based on how long it took to collect.  For cpus with consistent tick rates this should be a tighter bound than wall-clock time, although given cache effects it won't be entirely deterministic.  That said, I don't think determinism is something we can continue to rely on.

this change (optimistically) increases the amount of headroom we give the mutator after each collection,
except if the collection took more than 50ms, at which point we decrease the headroom.  The factors for growth/shrinking are 1.85 and 0.35 respectively.

In practice this causes the collector to oscillate around that desired time.

pyxl_bench.py is most sensitive to this work, for reasons that are unclear.  I've actually seen collection times *drop* after increasing allocBytesForNextCollection in pyxl_bench.py.  More investigation there would be useful - it is most sensitive to any change in GC parameters.

```
       django_template.py             4.6s (2)             4.4s (4)  -4.4%
            pyxl_bench.py             3.8s (2)             3.8s (4)  -0.5%
sqlalchemy_imperative2.py             5.0s (2)             4.9s (4)  -2.2%
        django_migrate.py             1.8s (2)             1.7s (4)  -2.1%
      virtualenv_bench.py             7.8s (2)             7.6s (4)  -2.3%
                  geomean                 4.1s                 4.0s  -2.3%
```

### DIFF
--- a/src/core/stats.cpp
+++ b/src/core/stats.cpp
@@ -65,7 +65,7 @@ std::unordered_map<uint64_t*, std::string>* Stats::names;
 bool Stats::enabled;
 
 timespec Stats::start_ts;
-uint64_t Stats::start_tick;
+uint64_t Stats::start_tick = 0;
 
 StatCounter::StatCounter(const std::string& name) : counter(Stats::getStatCounter(name)) {
 }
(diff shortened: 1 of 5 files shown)


kmod:
Chris and I chatted about this irl, but I'm pretty against having our GC be non-deterministic like this :/  I think tuning the collection frequency is still a good thing though.

PR #670, opened 2015-07-06
rewriter: use lea to allocate stack space

undingen:
+ add small cleanup: add helper for mode calculation


tjhance:
cool, looks good

PR #671, opened 2015-07-06
add babel to our extra tests

undingen:


undingen:
the ```setup.py import_cldr``` step is slow, caused by a lot of gc collections (probably because it's using a lot of generators and they register ```INITIAL_STACK_SIZE``` memory (=4096*8 bytes)).
#667 helps to decrease the time taken from 275secs to 213secs.

kmod:
Cool :)  these test suites are giving us way more coverage and have exposed a number of issues.  not the easiest to debug though :P

PR #672, opened 2015-07-06
Call finalizers during garbage collection with ordering

rudi-c:
Attempt #2 at a PR implementing PyPy's finalization order algorithm. Depends on PR #662 and #666 

As far as I can tell, 1/3 of the slowdown for django_template comes from the extra switch statement added in TraceStack which now has 3 different modes, and the other 2/3 of the slowdown comes from always adding unwind info in `doSafePoint`. `fasta.py` is suffering a massive slowdown because of large numbers of `PyCapsule` objects which don't seem to be safe to turn into simple destructors.

```
              pyston (calibration)                      :    1.0s baseline: 1.0 (+0.1%)
              pyston django_template.py                 :    4.6s baseline: 4.5 (+2.1%)
              pyston pyxl_bench.py                      :    3.7s baseline: 3.6 (+3.6%)
              pyston sqlalchemy_imperative2.py          :    5.2s baseline: 5.1 (+2.8%)
              pyston django_migrate.py                  :    1.8s baseline: 1.7 (+3.3%)
              pyston virtualenv_bench.py                :    5.3s baseline: 5.2 (+3.4%)
              pyston interp2.py                         :    3.9s baseline: 3.9 (-0.2%)
              pyston raytrace.py                        :    5.7s baseline: 5.3 (+7.4%)
              pyston nbody.py                           :    7.1s baseline: 6.8 (+5.8%)
              pyston fannkuch.py                        :    6.1s baseline: 5.8 (+5.1%)
              pyston chaos.py                           :   18.0s baseline: 16.0 (+12.8%)
              pyston fasta.py                           :    5.0s baseline: 4.1 (+22.2%)
              pyston pidigits.py                        :    4.9s baseline: 4.7 (+4.7%)
              pyston richards.py                        :    1.5s baseline: 1.4 (+1.3%)
              pyston deltablue.py                       :    1.4s baseline: 1.3 (+4.0%)
              pyston (geomean-10eb)                     :    4.3s baseline: 4.1 (+5.5%)
```


kmod:
fyi I think the PyCapsule objects are coming from PyArg_ParseTuple, which creates capsules to do cleanups on error (it creates capsule objects with destructors that frees memory, appends those capsules to a PyList, and then frees the list at the end of the function).  We could probably handle this case directly -- we could convert the system away from Python data structures, or have some special handling of capsule objects (if they don't have an instance-level destructor set, the class-level destructor is safe to run [or skip I guess]).

kmod:
Ok, looked through this and the other finalizer PRs and they look pretty good :)  assigning them back to chris for final review.  I don't fully understand the entire ordering procedure, so I'm going to advocate for having lots of testing, especially of all these edge cases (what happens if you add a `__del__` to a subclass of a builtin [I think you said that that crashes, but maybe we should test that], etc).

kmod:
oh, and quick question -- what happens to reference cycles that have finalizers involved?

rudi-c:
For cycles: http://doc.pypy.org/en/release-2.0-beta2/discussion/finalizer-order.html
```
This basic idea fails when there are cycles. It’s not a good idea to keep the objects alive forever or to never call any of the finalizers. The model we came up with is that in this case, we could just call the finalizer of one of the objects in the cycle – but only, of course, if there are no other objects outside the cycle that has a finalizer and a reference to the cycle.
```

This is bad but not much better than CPython, which ignores the whole cycle and puts it in a global variable for the programmer to deal with.

kmod:
just to make a note of it... let's definitely get a test for that :)  both that we free the memory (allocate enough that it would cause the test to fail), and that some finalizers got called.

rudi-c:
Ok I implemented the revisions. For the case of class with `__del__` subclassing a built-in it doesn't crash so much as assert if someone tries to do that. Beyond the assert, a crash would be theoretically possible if resurrection is involved in combination with subclassing the builtin but I'm not sure how I'd get one to happen.

kmod:
yeah, I guess I was saying that maybe we should have an `expected: fail` test that tries to subclass a builtin and then add a `__del__` and make sure it fails -- otherwise that code path will be untested.

rudi-c:
Ok I fixed that obscure bug that was hard to reproduce. See last commit.

rudi-c:
`gc_compat_free` is only called through `PyObject_Free` and `PyMem_Free`, `StlCompatAllocator` uses `gc_free`. I figured I'd leave an empty `gc_compat_free` just because the name of the function is somewhat self-documenting, but I don't feel strongly about it either way.

toshok:
ah, quite right.  my memory failed me wrt usage of gc_compat_free.  That said I think breaking the link at the cpython api layer is a lot cleaner.

rudi-c:
Sure -- updated.

toshok:
:shipit: woohoo

kmod:
nice!

PR #674, opened 2015-07-06
Be more consistent about "ip" while unwinding

kmod:
I think the ip we receive is the return address to that stack frame
(at least in the non-signal case).  This makes things a bit weird
since the "is this ip in this range" tests have a different half-openness
than they normally do.  At some point an "ip = ip - 1" snuck in, which
I think was to address this issue, but I think it's better to not
change the ip -- ie the resulting address is not independently useful (it's
in the middle of an instruction); we could pass it around as "ip_minus_one",
but instead just try converting the tests to be better.

### DIFF
--- a/src/codegen/unwinding.cpp
+++ b/src/codegen/unwinding.cpp
@@ -404,12 +404,12 @@ static unw_word_t getFunctionEnd(unw_word_t ip) {
 
 static bool inASTInterpreterExecuteInner(unw_word_t ip) {
     static unw_word_t interpreter_instr_end = getFunctionEnd((unw_word_t)interpreter_instr_addr);
-    return ((unw_word_t)interpreter_instr_addr <= ip && ip < interpreter_instr_end);
+    return ((unw_word_t)interpreter_instr_addr < ip && ip <= interpreter_instr_end);
 }
 
 static bool inGeneratorEntry(unw_word_t ip) {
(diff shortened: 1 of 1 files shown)


kmod:
@toshok, does this match your understanding of the purpose of the "-1"?

toshok:
yeah, the ip while unwinding points to the next instruction to execute after the call (the ret instruction pops IP from the stack.)  so the -1 should hopefully give us an ip corresponding to the same line number as the call itself.

toshok:
the only issue I can see with using <= for the end range check is that that ip might potentially point to a completely valid instruction.

might it make sense to use < for both?

kmod:
Yeah, I guess the right check would depend on what was meant when the range was generated, but I think if the range includes the call instruction, they probably meant to say that while we're in the call, we're in the relevant range.

ie I'm thinking of something like
```
bytes 0-9:   mov $0xdeadbeef, %r11
bytes 10-12: callq %r11
```
and the caller (llvm I think) specifies a byte range of "offset 0, length 13".  If we unwind, we see that we will return to byte 13 (the next instruction), which I think we want to consider as inside the range.  ip will point to a valid instruction which should be fine, since in this case we're concerned with what happens before ip (ie what we are currently doing) rather than what happens after it.

PR #678, opened 2015-07-08
Refactor offset macros to use the C++ built-in instead.

rudi-c:
#define SOMETHING_OFFSET -> offsetof(...)

Generates the same assembly, more standard and less `#define`s


kmod:
nice, thanks for cleaning this up

PR #679, opened 2015-07-08
runtime ICs: add support for passing args on the stack (=func with >6 args)

undingen:
This is important for callattr and runtimeCall ICs.
Supports ICs with and without frame pointers.
In addition I had to increase the scratch size and the patchpoint size a little bit.

Sadly looks like this doesn't have much influence on the perf.


undingen:
I'm closing this because #693 replaced this approach with a faster and cleaner one.

PR #683, opened 2015-07-10
add long.__float__ attribute

Daetalus:
#569 seems didn't continuing. So I improve it according the #569 conversations.

### DIFF
--- a/src/runtime/long.cpp
+++ b/src/runtime/long.cpp
@@ -338,7 +338,15 @@ extern "C" long PyLong_AsLongAndOverflow(Box* vv, int* overflow) noexcept {
 extern "C" double PyLong_AsDouble(PyObject* vv) noexcept {
     RELEASE_ASSERT(PyLong_Check(vv), "");
     BoxedLong* l = static_cast<BoxedLong*>(vv);
-    return mpz_get_d(l->n);
+
+    double result = mpz_get_d(l->n);
+
+    if (std::isinf(result)) {
+        PyErr_SetString(PyExc_OverflowError, "long int too large to convert to float");
+        return -1;
+    }
+
+    return result;
 }
 
 /* Convert the long to a string object with given base,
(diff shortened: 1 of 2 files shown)


src/runtime/long.cpp
```diff
@@ -688,6 +688,18 @@ Box* longInt(Box* v) {
         return new BoxedInt(n);
 }
 
+Box* longFloat(BoxedLong* v) {
+    if (!isSubclass(v->cls, long_cls))
+        raiseExcHelper(TypeError, "descriptor '__float__' requires a 'long' object but received a '%s'",
+                       getTypeName(v));
+    double result;
+    result = PyLong_AsDouble(v);
+    if (std::isinf(result)) {
```
kmod: Can we move this check into PyLong_AsDouble?  Here's CPython's implementation:
```C
double
PyLong_AsDouble(PyObject *v)
{
    Py_ssize_t exponent;
    double x;

    if (v == NULL || !PyLong_Check(v)) {
        PyErr_BadInternalCall();
        return -1.0;
    }
    x = _PyLong_Frexp((PyLongObject *)v, &exponent);
    if ((x == -1.0 && PyErr_Occurred()) || exponent > DBL_MAX_EXP) {
        PyErr_SetString(PyExc_OverflowError,
                        "long int too large to convert to float");
        return -1.0;
    }
    return ldexp(x, (int)exponent);
}
```
ie it contains overflow-checking which ours doesn't have :/

Also, you're running into the distinction we have between C-API exceptions (such as what gets set by `PyErr_SetString`, and gets returned by C API functions like PyLong_AsDouble) and our C++ exceptions (which you get with `raiseExcHelper` and this method should throw).  PyLong_AsDouble signals an exception by returning -1 and setting the C-api exception (or well, it should but currently doesn't :), which we will need to check with something like
```C
if (result == -1)
    checkAndThrowCAPIException();
```

kmod:
looks like you had the misfortune of updating this PR while our build was broken, but it looks like there's also a formatting issue with the change.  would you mind updating it?

otherwise it lgtm :)

kmod:
Awesome!  thanks for the contribution :)

PR #686, opened 2015-07-10
Specify older 'mock' version to unbreak sqlalchemy

undingen:


kmod:
Ah nice, thanks for fixing this :)

@rudi-c, it looks like you'll need to convince travis-CI to re-test your #687 to include this change.  it should do this automatically on the next push to that branch, or you can close+reopen the PR -- though that generates some extra emails for everyone so it could be better to just make a formatting change and push that :P

PR #687, opened 2015-07-10
Add docs and tips for new contributors

rudi-c:
The changes are mainly doc changes, so they are easier to read from my fork https://github.com/rudi-c/pyston than the diff itself.

Mostly of the changes are in `CONTRIBUTING.md` and `docs/TIPS.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,20 +1,52 @@
 ## Contributing to Pyston
 
-#### Pull Requests
+### Pull Requests
 
 Before a pull request can be merged, you need to to sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/).
 
-Please make sure to run the tests (`make check` and `make check_format`) on your changes.
+Please make sure to run at least the basic tests (`make quick_check`) on your changes. Travis CI will run the full test suite (`make check`) on each pull request.
 
 ##### Formatting
 
-Please make sure `make check_format` passes on your commits.  If it reports any issues, you can run `make format` to auto-apply
-the project formatting rules.  Note that this will format your files in-place with no built-in undo, so you may want to
-create a temporary commit if you have any unstaged changes.
+Please make sure `make check_format` passes on your commits.  If it reports any issues, you can run `make format` to auto-apply the project formatting rules.  Note that this will format your files in-place with no built-in undo, so you may want to create a temporary commit if you have any unstaged changes.
 
-### Projects
+Adding a pre-commit hook can be useful to catch formatting errors earlier. i.e. have in `~/pyston/.git/hooks/pre-commit`:
 
-If you don't know where to start:
-- Check out our [Github issues list](https://github.com/dropbox/pyston/issues), especially those marked ["probably easy"](https://github.com/dropbox/pyston/labels/probably%20easy)
+```
+#!/bin/sh
+exec make -C ~/pyston check_format
+```
+
+### Getting started with the codebase
+
+The easiest way to contribute to Pyston is to help us improve our compatibility with CPython. There are many small tasks to do such as built-in functions that are not yet implemented, edge cases that are not being handled by Pyston or where our output is slightly different than CPython, etc. The fix will often involve a local change, giving a smooth start to learning the codebase. One of Python's greatest strengths is that it comes ["batteries included"](https://xkcd.com/353/), but this means that there is a long long tail of these little tasks that needs to be driven through - your help is immensely valuable there!
+
+The command `make quick_check` will first run our Pyston tests (great way to make sure everything is in order) and then the default CPython tests. You will get an output that looks like this:
+
+```
+                  ...
+                  test_bastion.py    Correct output (125.7ms)
+                 test_unittest.py    (skipped: expected to fail)
+                     test_json.py    (skipped: expected to fail)
+                  test_future3.py    Correct output (952.8ms)
+                  ...
+```
+
+Notice that a large number tests are currently marked as failing (marked with an `# expected: fail` comment at the top of the file). Just pick any that you think is interesting and get it to pass! Remove the `#expected: fail` flag and run `make check_TESTNAME` (without `.py`) to compare the result to CPython's (the command will search for TESTNAME in the `test/` directory). If the test is crashing, the easiest way to start debugging is to use `make dbg_TESTNAME` which is essentially `make check_TESTNAME` inside GDB.
+
+This kind of work will often happen where native libraries are defined (e.g. `src/runtime/builtin_modules/builtins.cpp`), implementation of types (e.g. `src/runtime/str.cpp`) and debugging may involve tracing through the interpreter (`src/codegen/ast_interpreter.cpp`). The code in those files should be relatively straightforward. Code that involve the JIT (function rewriting, assembly generation, etc) is more intricate and confusing in the beginning (e.g. `src/asm_writing/rewriter.cpp`). Keep in mind, it's perfectly fine to ask for help!
+
+To save you some time, the cause of failures for some of the tests [may have already been identified](test/cpython/NOTES.org). Do note, however, that not all of CPython's behavior can be matched exactly. For example, by nature of having a garbage collector over reference counting, the freeing of objects is non-deterministic and we can't necessarily call object finalizers in the same order as CPython does.
+
+[Some tips on challenges you might run into and debugging tips](docs/TIPS.md).
+
+You can also check out our [Github issues list](https://github.com/dropbox/pyston/issues), especially those marked ["probably easy"](https://github.com/dropbox/pyston/labels/probably%20easy).
+
+### Communicating
+
+- We use a [gitter chat room](https://gitter.im/dropbox/pyston) for most of our discussions. If you need help with figuring out where to start or with the codebase, you should get a response there fairly quickly. If you found a small project to work on already and are eager to start, by all means get started! It is still a good idea to drop us a note - we might some suggestions or we can think of an edge case or two.
 - Email the [pyston-dev mailing list](http://lists.pyston.org/cgi-bin/mailman/listinfo/pyston-dev), or [browse the archives](http://lists.pyston.org/pipermail/pyston-dev/)
-- Join us on [#pyston](http://webchat.freenode.net/?channels=pyston) on freenode.
+
+### Bigger projects
+
+There are many big areas where Pyston can use improvements. This includes, for example, a better garbage collector, better performance profiling tools (including finding more benchmarks), finding new libraries to add to our test suite, etc. These can be very involved - if you are interested bigger projects (e.g. as part of graduate studies), please contact us directly.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ Notice that a large number tests are currently marked as failing (marked with an
 
 This kind of work will often happen where native libraries are defined (e.g. `src/runtime/builtin_modules/builtins.cpp`), implementation of types (e.g. `src/runtime/str.cpp`) and debugging may involve tracing through the interpreter (`src/codegen/ast_interpreter.cpp`). The code in those files should be relatively straightforward. Code that involve the JIT (function rewriting, assembly generation, etc) is more intricate and confusing in the beginning (e.g. `src/asm_writing/rewriter.cpp`). Keep in mind, it's perfectly fine to ask for help!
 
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -1,20 +1,45 @@
...
 
-Please make sure to run the tests (`make check` and `make check_format`) on your changes.
+Please make sure to run the tests (`make check`) on your changes.
 
 ##### Formatting
 
-Please make sure `make check_format` passes on your commits.  If it reports any issues, you can run `make format` to auto-apply
-the project formatting rules.  Note that this will format your files in-place with no built-in undo, so you may want to
-create a temporary commit if you have any unstaged changes.
+Please make sure `make check_format` passes on your commits.  If it reports any issues, you can run `make format` to auto-apply the project formatting rules.  Note that this will format your files in-place with no built-in undo, so you may want to create a temporary commit if you have any unstaged changes.
```
toshok: might be good to mention that a pre-commit hook might help, either to run `check_format` or `format`.

e.g. I have in ~/pyston/.git/hooks/pre-commit:

```
#!/bin/sh
exec make -C ~/pyston check_format
```

CONTRIBUTING.md
```diff
@@ -1,20 +1,52 @@
 ## Contributing to Pyston
 
-#### Pull Requests
+### Pull Requests
 
 Before a pull request can be merged, you need to to sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/).
 
-Please make sure to run the tests (`make check` and `make check_format`) on your changes.
+Please make sure to run the tests (`make check`) on your changes.
```
kmod: This might be a good place to mention the existence of `make quick_check` (which also checks formatting, which it didn't when we wrote this doc) and that we usually rely on travis-ci to run the full test suite anyway.

CONTRIBUTING.md
```diff
@@ -1,20 +1,52 @@
...
+```
+                  ...
+                  test_bastion.py    Correct output (125.7ms)
+                 test_unittest.py    (skipped: expected to fail)
+                     test_json.py    (skipped: expected to fail)
+                  test_future3.py    Correct output (952.8ms)
+                  ...
+```
+
+Notice that a large number tests are currently marked as failing (marked with an `# expected: fail` comment at the top of the file). Just pick any that you think is interesting and get it to pass! Remove the `#expected: fail` flag and run `make check_TESTNAME` (without `.py`) to compare the result to CPython's (the command will search for TESTNAME in the `test/` directory). If the test is crashing, the easiest way to start debugging is to run `gdb --args ./pyston_dbg test/cpython/TESTNAME.py`.
```
kmod: There's also `make dbg_TESTNAME` :P

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+
+In general, boxes are allocated in the heap (but not all - e.g. classes of built-in types) in one of the three heaps. Pointers in the heap have a hardcoded range of allowed values (see `src/gc/heap.h`) - values that fall outside this range (and are not in the stack) could potentially indicate memory errors.
+
+The C++ stacktrace can be quite difficult to read. Every Python function call will result in half a dozen C++ function calls along the lines of `getAttrInternal`. This is why the `_printStacktrace()` function, which can be used inside GDB, can come in very handy.
```
kmod: Hmm this function doesn't seem to work for me, since I think it tries to crawl the gdb stack.  I tend to use `signal SIGUSR1`.

Also, it might be worth mentioning the `dump()` method, which you can call on any pointer and it will try to print some simple info about an object (class name, as well as some type-specific info for some supported builtin types).

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+
+##### Problem: The tests fail in release mode
+
+For example, `tests/sys_argv` might output `['/mnt/rudi/pyston/test/tests/sys_argv.py']` instead of `['/mnt/rudi/pyston/./test/tests/sys_argv.py']`. If you switch between testing in debug mode and release mode, try running `touch test/tester.py` in-between runs.
```
kmod: I think this is due to switching between the cmake-based test targets (check and quick_check I believe) and the makefile-based ones (check_dbg, check_release).

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+}
+
+void initCodegen() {
+    ...
+    signal(SIGSEGV, &handle_sigsegv);
+    ...
+}
+```
```
kmod: Could be worth mentioning `PYSTON_RUN_ARGS=G` and `=P`.  `-P` should probably do the segfault thing you mention here :)

kmod:
this is awesome, thanks for putting this together!

docs/TIPS.md
rudi-c: What do those flags do?

rudi-c:
Should be ready now.

test/CPYTHON_TEST_NOTES.md
```diff
@@ -1,54 +1,20 @@
...
-- pypa-parser.cpp: readName: assert(e.type == pypa::AstType::Name)
-- src/runtime/util.cpp: parseSlice: assert(isSubclass(start->cls, int_cls) || start->cls == none_cls)
+- missing `__name__` attributes on capifuncs
+- missing `sys.__stdout__` attribute
+- `serialize_ast.cpp`: `writeColOffset: assert(v < 100000 || v == -1)` gets tripped
+- `pypa-parser.cpp`: readName: `assert(e.type == pypa::AstType::Name)`
+- `src/runtime/util.cpp`: `parseSlice`: `assert(isSubclass(start->cls, int_cls) || start->cls == none_cls)`
 
-** list of files & why they're failing
+## List of files & why they're failing
```
toshok: are these all up-to-date?  instead of going through the list to make sure as part of this PR, I'd suggest writing a python script in tools/ (list_test_failures.py) that essentially does a `grep -A1 "expected: fail"`, along with other patterns for the tests we skip, and formats it nicely.  Then this part of the doc could go away (aside from the script reference) and hopefully it wouldn't go out of date.

CONTRIBUTING.md
```diff
@@ -1,20 +1,52 @@
...
+                  test_future3.py    Correct output (952.8ms)
+                  ...
+```
+
+Notice that a large number tests are currently marked as failing (marked with an `# expected: fail` comment at the top of the file). Just pick any that you think is interesting and get it to pass! Remove the `#expected: fail` flag and run `make check_TESTNAME` (without `.py`) to compare the result to CPython's (the command will search for TESTNAME in the `test/` directory). If the test is crashing, the easiest way to start debugging is to use `make dbg_TESTNAME` which is essentially `make check_TESTNAME` inside GDB.
+
+This kind of work will often happen where native libraries are defined (e.g. `src/runtime/builtin_modules/builtins.cpp`), implementation of types (e.g. `src/runtime/str.cpp`) and debugging may involve tracing through the interpreter (`src/codegen/ast_interpreter.cpp`). The code in those files should be relatively straightforward. Code that involve the JIT (function rewriting, assembly generation, etc) can be more intricate (e.g. `src/asm_writing/rewriter.cpp`). If you start running into that area of the code, it's perfectly fine to ask for help!
```
toshok: I'd rewrite "If you start running into that area of the code, it's perfectly fine to ask for help!" to something like "Keep in mind, it's perfectly fine to ask for help!" - it's always okay to ask for help, regardless of what part of the code you're in :)

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+### Exceptions and stack unwinding
+
+Having a custom unwinder, Pyston has a few special constraints - see [exception safety](EXCEPTION-SAFETY.md) and [unwinding](UNWINDING.md).
+
+### Make shortcuts
+
+The main Make targets you will use will be
+- `make`: to compile in debug mode and generate the `pyston_dbg` executable
+- `make pyston_release`: to compile in release mode and generate the `pyston_release` executable
+- `make check`: run the tests
```
toshok: I'd switch `quick_check` and `check` - `quick_check` here, `check` down in the larger list.

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+
+Some benchmarks are particularly prone to variations caused by trivial changes. [http://speed.pyston.org/](http://speed.pyston.org/timeline) shows the performance graphs of many of the benchmarks. If a benchmark already shows a fair amount of up and downs, it is not as big a deal if the PR causes a performance drop in that benchmark. For example, [django_template](http://speed.pyston.org/timeline/#/?exe=1,3&base=none&ben=django_template&env=2&revs=1000&equid=off) is a benchmark that performs better and better, but [nbody](http://speed.pyston.org/timeline/#/?exe=1,3&base=none&ben=nbody&env=2&revs=1000&equid=off) has the tendency to be affected by the most unrelated changes.
+
+### Profiling tools
+
+The general go-to tool for profiling that we use is `perf`, which shows the hot functions in which the program spends the most time and their callers. It can be used as follows:
+
+```
+perf record -g -- ./pyston_release ../pyston-perf/benchmarking/benchmark_suite/django_template.py
```
toshok: I'd put `-p -q` here in this command line.  it's so useful that I don't think we want to suggest people should do a perf run without it.

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+This will create a `.perf` file in the same directory that can be viewed using `perf report -n`. We recommend using the `-n` flag which shows the number of samples per function as well as the percentage, which is useful when comparing two perf runs.
```
toshok: `perf.data` file

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+Adding `-p -q` to `./pyston_release` can output some more information about jitted frames that perf can use. Perf has a few other useful flags, such as `-e page-faults` which counts the number of page faults during an execution. While we optimize some of it away, Python intrinsically allocates massive amounts of memory and we are definitely at the point where it is worth thinking about caching, at all levels of abstraction.
+
+For some notes on other profiling tools, see [PROFILING](PROFILING).
+
+## Travis CI
+
+We use Travis CI for continuous integration. Pull requests to the main repository will be run automatically against our full test suite upon creation and updates.
+
+##### Problem: My code passes all the tests locally but fails on Travis CI
+/spp
```
toshok: vim typo?

docs/TIPS.md
undingen: or directly recommend to use ```make perf_foo```

docs/TIPS.md
```diff
@@ -0,0 +1,251 @@
...
+
+There are a lot of ways a pointer to an object could remain on the stack despite being conceptually unreachable. The pointer could have been placed on the stack when it was used as the argument to a function, but newer function calls never overwrote that value due to gaps in or between stack frames.
+
+The pointer value could have also been generated randomly. How likely is it that an integer value just happens to be a valid pointer? More often than you would think, and for the most obscure reasons. For example, you have a valid pointer `0x127021df48` to some random object, such as an xrange iterator which the compiler stores in register `r14` as an optimization. Then, the register gets reused, but in such a way that the `mov` instruction operates only in the lower 8 bits of `r14` to store the value `1`. Now, `r14` contains the value `0x127021e001`, a valid interior pointer to an otherwise unreachable object. Then, `r14` gets pushed onto the stack by callee-save conventions. Now, a pointer to this object exists on the stack so it doesn't get freed until that stack frame is popped.
+
+It can get really obscure.
```
toshok: really like this file, btw.  really useful stuff.

src/codegen/ast_interpreter.cpp
```diff
@@ -75,6 +75,12 @@ class RegisterHelper {
     static void deregister(void* frame_addr);
 };
 
+/*
+ * ASTInterpreters exist per function frame - there's no global interpreter object that executes
+ * all non-jitted code!
+ *
+ * The ASTInterpreter inherits from Box as part of generator support.
```
toshok: this likely won't hold true upon rebasing (or at least in the future) - with the GCWrapper stuff @undingen is working on.  But I'd s/generator support/GC support/ or something to that effect.

toshok:
Finally made it through it all -- worked *much* better reading the documents straight through from your branch (https://github.com/rudi-c/pyston/tree/docs) than reading just the diffs.

:+1:

docs/TIPS.md
rudi-c: derp that always happens

rudi-c:
Ok, done fixing those ups. Thanks for taking the time to read through that wall of text!

toshok:
awesome to have more/better docs :)

PR #693, opened 2015-07-12
baseline jit: directly emit patchpoints instead of using runtime ICs

undingen:
In order to speedup baseline JITed code directly emit patchpoints.
- reduces number of calls
- reduces number of EH frames
- supports stack args (currently hard coded to 8args)
- maybe slightly faster unwinding because of of reduced call nesting  

I choose to directly add the stack args support to the baseline jit instead of the rewriter because I think to get this right for all cases is harder and in the baseline jit I have special knowledge about the stack layout which makes it easier.

```
pyston django_template.py                 :    3.7s base_3: 3.9 (-4.0%)
pyston pyxl_bench.py                      :    3.4s base_3: 3.6 (-6.0%)
pyston sqlalchemy_imperative2.py          :    4.6s base_3: 4.7 (-2.0%)
pyston django_migrate.py                  :    1.7s base_3: 1.7 (-0.8%)
```

### DIFF
--- a/src/asm_writing/rewriter.cpp
+++ b/src/asm_writing/rewriter.cpp
@@ -970,10 +970,12 @@ void Rewriter::_call(RewriterVar* result, bool has_side_effects, void* func_addr
         assert(assembler->hasFailed() || asm_address == (uint64_t)assembler->curInstPointer());
     }
 
-    assert(vars_by_location.count(assembler::RAX) == 0);
-    result->initializeInReg(assembler::RAX);
+    if (!failed) {
+        assert(vars_by_location.count(assembler::RAX) == 0);
+        result->initializeInReg(assembler::RAX);
 
-    assertConsistent();
+        assertConsistent();
+    }
 
     result->releaseIfNoUses();
 }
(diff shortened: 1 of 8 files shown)


undingen:
Cool could even remove more lines than I added :-)

src/asm_writing/rewriter.cpp
```diff
@@ -950,6 +945,16 @@ void Rewriter::_call(RewriterVar* result, bool has_side_effects, void* func_addr
...
+}
+
+void Rewriter::_call(RewriterVar* result, bool has_side_effects, void* func_addr, const RewriterVar::SmallVector& args,
+                     const RewriterVar::SmallVector& args_xmm) {
+    assembler->comment("_call");
+
+    // RewriterVarUsage scratch = createNewVar(Location::any());
+    assembler::Register r = allocReg(assembler::R11);
+
+    _setupCall(result, has_side_effects, args, args_xmm);
```
kmod: do we need to worry about if _setupCall ended up using R11?  I don't think it will, but it might be nice to assert something like `vars_by_location.count(r) == 0`.

undingen:
added the assert and rebased

PR #694, opened 2015-07-13
fix analysis unittests by not asserting clfunc is non-NULL

tjhance:
the analysis unittest creates an OSREntryDescriptor with clfunc=NULL.

Maybe the 'correct' fix is to change the unittest to not create an invalid OSREntryDescriptor, but I don't see an easy way to do that.


kmod:
shoot sorry, I created+merged my pr before seeing this one :/

tjhance:
XD it's fine, looks like your fix was better anyway

PR #697, opened 2015-07-13
travis-ci build both Release and Debug

dagar:


kmod:
Hey, sorry I ended up resurrecting #522 (but I adopted your naming).  I ended up configuring it to not run the "extra" tests in debug mode and the testing time is about the same as release mode, at least once I let ccache warm up.

dagar:
Sounds good. I didn't know about #522.

PR #699, opened 2015-07-14
Temporarily backout #656

kmod:
It had a merge conflict with #693, but even after fixing that (at least
superficially), I'm still getting test failures.  Let's just back it out
for now to make the tree green again.


kmod:
I think the build failures on master are unrelated; closing this.

PR #701, opened 2015-07-14
fix null check

toshok:
sorry about that - for some reason I had it in my head that travis-ci still did quick_check against pyston_dbg


kmod:
can we move the null check into the assert?

toshok:
yeah, that'd work too, sec.

PR #706, opened 2015-07-15
Slightly speedup ASTInterpreter::initArguments

undingen:
This function can get quite hot because now with the baseline jit we execute it much more often.
I'm not really excited about adding the ```InternedString``` fields to ```ParamNames``` but I don't know where to add them else.
```
                           e648044ce2266a3fc6:  b8f5647281593fc5ed:
       django_template.py             3.5s (4)             3.4s (4)  -0.7%
            pyxl_bench.py             3.3s (4)             3.3s (4)  -0.6%
sqlalchemy_imperative2.py             4.1s (4)             4.1s (4)  -0.7%
                  geomean                 3.6s                 3.6s  -0.7%
```

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -35,6 +35,7 @@
 #include "core/stats.h"
 #include "core/thread_utils.h"
 #include "core/util.h"
+#include "gc/roots.h"
 #include "runtime/generator.h"
 #include "runtime/import.h"
 #include "runtime/inline/boxing.h"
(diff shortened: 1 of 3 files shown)


undingen:
I added another optimization (cache lookup type) and changed the ```ParamNames``` to contain ```AST_Name``` nodes. (Could be nicer any ideas?)

perf is now:
```

       django_template.py             3.5s (4)             3.4s (4)  -1.1%
            pyxl_bench.py             3.3s (4)             3.3s (4)  -1.1%
sqlalchemy_imperative2.py             4.1s (4)             4.1s (4)  -1.2%
                  geomean                 3.6s                 3.6s  -1.1%

```

src/core/types.h
```diff
@@ -147,6 +148,9 @@ struct ParamNames {
     std::vector<llvm::StringRef> args;
     llvm::StringRef vararg, kwarg;
 
+    std::vector<AST_Name*> arg_names;
+    AST_Name* vararg_name, *kwarg_name;
```
kmod: Could we get rid of the StringRef versions?  or at least have a note about when one should be used vs the other (I assume that the stringref ones are now deprecated but it'd be nice to have a comment).

src/codegen/ast_interpreter.cpp
```diff
@@ -285,32 +285,35 @@ ASTInterpreter::ASTInterpreter(CLFunction* clfunc)
     assert(scope_info);
 }
 
-void ASTInterpreter::initArguments(int nargs, BoxedClosure* _closure, BoxedGenerator* _generator, Box* arg1, Box* arg2,
-                                   Box* arg3, Box** args) {
+void ASTInterpreter::initArguments(BoxedClosure* _closure, BoxedGenerator* _generator, Box* arg1, Box* arg2, Box* arg3,
+                                   Box** args) {
+    RootedBox arg1_root(arg1);
+    RootedBox arg2_root(arg2);
+    RootedBox arg3_root(arg3);
```
kmod: have you made much progress on figuring out why these are needed?  I wouldn't mind punting on that issue for now, but we should have some sort of comment explaining why we have these RootedBoxes here.

src/core/types.h
undingen: We use ```ParamNames``` also for specifying argument names for runtime functions where we don't need and don't have a ```InternedStringPool``` available which we would need to completely switch to ```AST_Name``` nodes.
Maybe it's best to split this struct into two? Or just add a comment that the AST_Names will only get set if source is available...

kmod: ah ok, makes sense.  I think the comment sounds good :)  splitting them into separate structs probably makes more sense, but for now this fits in with our "stuff things into the AST until we get around to doing a bytecode system" strategy :P

src/codegen/ast_interpreter.cpp
undingen: Concerning the ```RootedBoxes``` I'm trying to remove them again. It looked like it works without them. And I can't reason why the are needed inside this function (in a debug build).
But the next build will tell more...

undingen:
please review :-)

PR #708, opened 2015-07-15
rewriter: release allocated scratch space

undingen:
This patch makes scratch space allocated with ```Rewriter::allocate``` available when the ```RewriterVar``` becomes unused.

I'm not really happy about this patch, I couldn't add the release logic into ```bumpUse``` because it get's called to early (the scratch space will get used afterwards but we could spill vars in between which would overwrite the argument)
@tjhance any ideas?

It looks like we run only rarely out of scratch space and in non of our important benchmarks.

### DIFF
--- a/src/asm_writing/rewriter.cpp
+++ b/src/asm_writing/rewriter.cpp
@@ -551,6 +551,12 @@ void Rewriter::_setAttr(RewriterVar* ptr, int offset, RewriterVar* val) {
     }
 
     ptr->bumpUse();
+
+    // If the value is a scratch allocated memory array we have to make sure we won't release it immediately.
+    // Because this setAttr stored a reference to it inside a field and the rewriter can't currently track this uses and
+    // will think it's unused.
+    if (val->hasScratchAllocation())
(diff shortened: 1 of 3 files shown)


tjhance:
I haven't taken a close look but if bumpUse is being called too early, then the call is probably incorrect and we should move it later

undingen:
The problem is that we call bumpUse before emitting a call instruction after we put the args in place.
Afterwards we spill all the args which are clobbered by the call -> this may use one of the scratch slots we just have released but which will be used by the call...

But now that I think about it, I probably just have to call bumpUse after saving the caller-saved regs and only save the ones which are used afterwards...

tjhance:
Sounds right

undingen:
@tjhance should work now, could you please review it
it failed travis ci checks at first because of some pyicu error. Rebuild helped not sure if this PR caused it...

kmod:
btw Travis if you're happy with it you should feel free to merge the pr -- I think we should go ahead and say that's part of the role of the "assignee".

PR #709, opened 2015-07-15
Mercurial fixes

toshok:
couple of minor changes that enable us to run mercurial (using their pure implementations of some modules.  the native modules don't build yet.)  enough to clone the mercurial source repo.

it's very, very slow at the moment.  cpython does it in about 30s, and we do it in 4.5m

### DIFF
--- a/src/runtime/file.cpp
+++ b/src/runtime/file.cpp
@@ -480,6 +480,9 @@ static PyObject* file_write(BoxedFile* f, Box* arg) noexcept {
     if (!f->writable)
         return err_mode("writing");
     if (f->f_binary) {
+        // NOTE: this call will create a new tuple every time we write to a binary file. if/when this becomes hot or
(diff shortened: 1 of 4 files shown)


toshok:
I think instead of exposing `getbuffer`, I'd rather make a `PyArgs_ParseTuple` that takes the single arg (or maybe refactor `PyArgs_ParseTuple` to take length + Box*, so we can pass `&arg` from here, and `_ParseTuple` can pass the pointer to the tuple's data).  That way we shouldn't get caught with more cases we aren't handling.

sound good?

kmod:
ah ok that sounds like a better approach (I'm partial to the taking a length+Box*).  it looks like this might be a non-issue though since our file.write time is dominated by other work (saving thread state, syscall).  maybe we can just leave a note+todo?

toshok:
added TODO and rebased.

PR #713, opened 2015-07-17
optimize regex handling

kmod:
I think these are good changes, but the perf results are exactly the same as my non-perf-related PR:

```
       django_template.py             3.6s (2)             3.6s (2)  -0.2%
            pyxl_bench.py             3.5s (2)             3.4s (2)  -1.2%
sqlalchemy_imperative2.py             4.3s (2)             4.4s (2)  +0.4%
                  geomean                 3.8s                 3.8s  -0.4%
```

It speeds up some simple regex microbenchmarks by 4x though.

### DIFF
--- a/from_cpython/Modules/_sre.c
+++ b/from_cpython/Modules/_sre.c
@@ -1768,6 +1768,9 @@ state_init(SRE_STATE* state, PatternObject* pattern, PyObject* string,
     if (!ptr)
         return NULL;
 
+    // Pyston change:
+    assert(charsize == 1 || charsize == sizeof(Py_UNICODE));
+
     /* adjust boundaries */
     if (start < 0)
         start = 0;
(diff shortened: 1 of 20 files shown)


kmod:
It looks like Travis-CI tested the wrong commit, but I ran the tests manually and it seems ok.

Here are the stats on the benchmarks I was looking at:
```
         django_lexing.py             2.0s (2)             1.8s (2)  -11.6%
      re_split_ubench2.py             2.9s (2)             1.8s (2)  -38.7%
       re_split_ubench.py             3.1s (2)             0.8s (2)  -75.3%
```
django_lexing is the lexing part of the template parsing; re_split_ubench2 is the regex portion of the lexing, and re_split_ubench is just a tiny regex (something like `re.split("banana" * 100, "ana")`).

PR #714, opened 2015-07-17
Convert "a in (b, c)" to "a == b or a == c"

kmod:
Not super common, but can seems to help since we are particularly bad at handling `a in (b, c)`.

```
       django_template.py             3.6s (2)             3.6s (2)  -0.5%
            pyxl_bench.py             3.5s (2)             3.5s (2)  -0.4%
sqlalchemy_imperative2.py             4.3s (2)             4.3s (2)  -0.9%
                  geomean                 3.8s                 3.8s  -0.6%
```

### DIFF
--- a/src/codegen/compvars.cpp
+++ b/src/codegen/compvars.cpp
@@ -455,6 +455,10 @@ public:
         return new ConcreteCompilerVariable(UNKNOWN, rtn, true);
     }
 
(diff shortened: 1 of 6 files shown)


tjhance:
I haven't looked too closely at your diff but the transformation in your subject line is incorrect, e.g.,

```
>>> n = float('nan')
>>> n in (n,n)
True
>>> n == n or n == n
False
```

kmod:
oh doh that's true

PR #715, opened 2015-07-17
Get test_iter.py to work

tjhance:
I need to clean this up, maybe write some tests for ourselves, but this nearly gets `test/cpython/test_iter.py` working.

There was a function `calliter_hasnext` in `from_cpython/Objects/iterobject.c`. It needs to throw exceptions, so I moved it to our own `iter.cpp`.

The behavior of `filter` depends on the type of its input. For filtering strings, I used cpython's code; for filtering tuples I wrote my own. I should probably pick a style. I guess `test_iter.py` doesn't implement filtering unicode, but I guess I should do that too.

The test is still throwing one exception:

```
d = {"one": 1, "two": 2, "three": 3, 1j: 2j}
for k in d:                                 
    self.assertIn(k, d)                     
    self.assertNotIn(k, d.itervalues())     
for v in d.values():                        
    self.assertIn(v, d.itervalues())
    self.assertNotIn(v, d)        
```

I have no idea why. But it's 2am, so I'll just put it up here to run tests and finish it up tomorrow.

### DIFF
--- a/from_cpython/Objects/iterobject.c
+++ b/from_cpython/Objects/iterobject.c
@@ -66,7 +66,7 @@ calliter_next(calliterobject *it)
                 Py_CLEAR(it->it_sentinel);
             }
         }
-        else if (PyErr_ExceptionMatches(PyExc_StopIteration) || PyErr_ExceptionMatches(PyExc_IndexError)) {
+        else if (PyErr_ExceptionMatches(PyExc_StopIteration)) {
             PyErr_Clear();
             Py_CLEAR(it->it_callable);
             Py_CLEAR(it->it_sentinel);
--- a/from_cpython/Objects/iterobject.c
+++ b/from_cpython/Objects/iterobject.c
@@ -4,15 +4,6 @@
 
 #include "Python.h"
 
-extern PyTypeObject PyCallIter_Type;
-typedef struct {
-    PyObject_HEAD
-    PyObject *it_callable; /* Set to NULL when iterator is exhausted */
-    PyObject *it_sentinel; /* Set to NULL when iterator is exhausted */
-    // Pyston changes:
-    PyObject *it_nextvalue; /* Set to non-null when iterator is advanced in __hasnext__ */
-} calliterobject;
-
 PyObject *
 PyCallIter_New(PyObject *callable, PyObject *sentinel)
 {
@@ -52,7 +43,7 @@ calliter_traverse(calliterobject *it, visitproc visit, void *arg)
 
 // Pyston change: extract most of the body of calliter_iternext here
 // so we can use it from both calliter_iternext and calliter_hasnext
-static PyObject *
+PyObject *
 calliter_next(calliterobject *it)
 {
     if (it->it_callable != NULL) {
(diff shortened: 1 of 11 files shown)


tjhance:
yay, got this working (fixed, the dict iterator problem I mentioned, fixed a few other problems, got the test case passing)

from_cpython/Objects/iterobject.c
```diff
@@ -75,7 +66,7 @@ calliter_next(calliterobject *it)
                 Py_CLEAR(it->it_sentinel);
             }
         }
-        else if (PyErr_ExceptionMatches(PyExc_StopIteration)) {
+        else if (PyErr_ExceptionMatches(PyExc_StopIteration) || PyErr_ExceptionMatches(PyExc_IndexError)) {
```
kmod: Hmm this feels a bit fishy -- why do we need it when cpython doesn't?

src/codegen/compvars.cpp
```diff
@@ -1086,7 +1086,10 @@ class IntType : public ConcreteCompilerType {
     }
 
     CompilerVariable* contains(IREmitter& emitter, const OpInfo& info, VAR* var, CompilerVariable* lhs) override {
-        return makeBool(false);
+        llvm::CallSite call
+            = emitter.createCall(info.unw_info, g.funcs.raiseNotIterableError, embedConstantPtr("int", g.i8_ptr));
+        call.setDoesNotReturn();
+        return new ConcreteCompilerVariable(BOOL, llvm::UndefValue::get(BOOL->llvmType()), true);
```
kmod: doh thanks for fixing these

src/runtime/inline/list.cpp
```diff
@@ -39,21 +39,42 @@ Box* listiterHasnext(Box* s) {
     assert(s->cls == list_iterator_cls);
     BoxedListIterator* self = static_cast<BoxedListIterator*>(s);
 
-    return boxBool(self->pos < self->l->size);
+    if (!self->l) {
+        raiseExcHelper(StopIteration, "");
```
kmod: Could we return false here and rely on the caller to throw the StopIteration if they want one?

src/runtime/objmodel.cpp
```diff
@@ -4120,7 +4120,7 @@ Box* compareInternal(Box* lhs, Box* rhs, int op_type, CompareRewriteArgs* rewrit
             if (result < 0)
                 throwCAPIException();
             assert(result == 0 || result == 1);
-            return boxBool(result);
+            return boxBool(op_type == AST_TYPE::NotIn ? !result : result);
```
kmod: doh good catch

from_cpython/Objects/iterobject.c
tjhance: Hm that does feel super fishy. I don't remember what prompted this but I'll take another look.

kmod: I noticed there are some other parts of this pr that deal with IndexErrors (the seqiter stuff) which makes sense; maybe it was just a testing change?  it looks like test_iter still passes if this line is reverted.

kmod:
Cool :)  I'm surprised but neither the label change or title change triggered emails.

PR #718, opened 2015-07-18
make attributes interned BoxedStrings

kmod:
instead of using llvm::StringMaps.  I guess the downside is additional gc pressure (django_template accumulates 15k interned strings) -- but this is memory that would get allocated anyway, and we probably save overall by de-duplicating between hiddenclasses.  We also have the ability to un-intern them when they become garbage ("mortal" interned strings).

```
       django_template.py             3.6s (2)             3.6s (2)  -2.3%
            pyxl_bench.py             3.4s (2)             3.4s (2)  -2.0%
sqlalchemy_imperative2.py             4.3s (2)             4.1s (2)  -6.7%
                  geomean                 3.8s                 3.6s  -3.7%
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,8 +81,8 @@ else
 endif
 
 TOOLS_DIR := ./tools
-TEST_DIR := ./test
-TESTS_DIR := ./test/tests
+TEST_DIR := $(abspath ./test)
+TESTS_DIR := $(abspath ./test/tests)
 
 GPP := $(GCC_DIR)/bin/g++
 GCC := $(GCC_DIR)/bin/gcc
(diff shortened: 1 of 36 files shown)


undingen:
Really nice perf change! :+1:

undingen:
there may be a few ```giveAttr``` calls which could also use a immortal static string if the ```internStringMortal``` ends up becoming hot.

kmod:
thanks for the comments :)  most of the calls to internStringMortal with constant strings are just me being lazy :P  I changed the ones that will be called more than once to immortal strings.  my thinking is that the ones that are only called once (ex copying an attr during initialization) shouldn't be that much more expensive (if at all) than what we had before, since we do the same number StringMap lookups, just now it's into the interned_strings map instead of the attr_offsets map.  but yeah something to keep an eye on :)

PR #719, opened 2015-07-18
Pow

Daetalus:
Complete rewrite the `int` and `long` power implementation. 

- `intPow` only handle all parameters are `int`. Otherwise call `longPow` or `floatPow`.
- `longPow` only handle all parameters are `int` or `long`. Otherwise call `floatPow`.

- use `addFunc` function to add `__pow__` attribute in `int` and `long`, to let tenary function to choose which implementation will be called.

- add completed error checking in `intPow` and `longPow`, the error messages order consist with CPython message order:
 **If has mod**:
 - check second argument whether is negative: 
`TypeError: pow() 2nd argument cannot be negative when 3rd argument specified`
 - Second all argument must not be float:
`TypeError: pow() 3rd argument not allowed unless all arguments are integers`
 - Third argument must not be zero:
`ValueError: pow() 3rd argument cannot be 0`

* according the GMP documentation:

>(In pow calculation) The sign of the divisor is ignored; the result is always non-negative.

So in `longPow` implementation, get the pow result of `rhs` and `lhs` first, then call `longMod` to get the final result. Then the result is consist with CPython.


test/tests/float.py
```diff
@@ -88,3 +88,16 @@ class F2(float):
...
 except OverflowError as e:
     print e.message
+
+all_args = [None, 1, 1L, -1, -1L, 2.0, 0.5, 0]
+for lhs in all_args:
+    for rhs in all_args:
+        for mod in all_args:
+            try:
+                pow(lhs, rhs, mod)
+            except ValueError as e:
```
kmod: Can we print out the type of the exceptions here?  Maybe something like
```
except Exception as e:
  print type(e), e
```
It looks like we are generating some ValueErrors where cpython generates TypeErrors.

Also, can we print out the result of the `pow` call?  Also let's add some more entries to all_args while we're at it :)  Maybe add `["", 0.0, -0.0, -0.5, float('nan'), float('inf')]` (looks like `-0.5` generates the wrong error message and `-0.0` produces a slightly bad result [-0.0 instead of 0.0]).

src/runtime/long.cpp
```diff
@@ -1191,49 +1192,81 @@ Box* longRTrueDiv(BoxedLong* v1, Box* _v2) {
...
-        return r;
+    } else if (mpz_sgn(rhs_long->n) == -1) {
+        BoxedFloat* rhs_float = static_cast<BoxedFloat*>(longFloat(rhs_long));
+        BoxedFloat* lhs_float = static_cast<BoxedFloat*>(longFloat(lhs));
+        return boxFloat(pow_float_float(lhs_float->d, rhs_float->d));
+    }
+
+    if (mod_long) {
+        mpz_pow_ui(r->n, lhs->n, mpz_get_ui(rhs_long->n));
+        return longMod(r, mod_long);
```
kmod: I think we need to be careful here if the long doesn't fit in an integer, ie `mpz_get_ui` will return just the low bits of the long.  For 1 and -1 as the base, we have the same behavior as CPython since it's only the lowest bit that matters.  For 0 though, we give the wrong result: we say `0 ** (1<<100)` is 1 (since we treat `1<<100` as 0),, but CPython correctly says it is 0.

For other bases, CPython will crash (will try to do the exponentiation even though there is not enough memory), and we will just silently return the wrong value.  ex `2 ** (1<<100)` for us is 1.

My suggestion is that we should call `mpz_fits_ulong_p`, and if that is true then do this path.  If it doesn't fit, then we check+handle 0/1/-1 cases, and for the others we abort() (not sure it's safe to throw an OverflowError when CPython would have crashed).

For the pow case with a mod, I think it's important that we do the pow+mod as a single step, rather than pow-then-mod -- integrating the two means that the result can be computed even if the pow would not fit in memory.  ex `pow(1<<100, 1<<100, 1000)` works just fine and should return 376.  The modulus-sign rules sound tricky, and I can never remember how python does those, but we should either try to apply them to the result of mpz_powm, or we should abort with an "unimplemented" message :/

test/tests/float.py
```diff
@@ -88,3 +88,16 @@ class F2(float):
...
+                print e.message
+            except TypeError as e:
+                print e.message
+            except ZeroDivisionError as e:
+                print e.message
```
kmod: Here are some more tests that currently don't work:
`pow(1<<30, 1<<30, 127)`
`pow(1L<<30, 1L<<30, 127)`

It's probably safe to leave those as unimplemented right now, but currently those will cause us to run out of memory.

src/runtime/int.cpp
```diff
@@ -663,36 +663,67 @@ extern "C" Box* intMul(BoxedInt* lhs, Box* rhs) {
...
+            raiseExcHelper(ValueError, "pow() 3rd argument cannot be 0");
+        }
     }
+
+    Box* rtn = pow_i64_i64(lhs->n, rhs_int->n);
+    if (mod == None)
+        return rtn;
+    if (isSubclass(rtn->cls, long_cls))
+        return longInt(binop(rtn, mod, AST_TYPE::Mod));
+    return binop(rtn, mod, AST_TYPE::Mod);
```
kmod: Same comment, I don't think it's safe to do the pow and mod separately if the pow overflows; ex `pow(1<<30, 1<<30, 127)`.  Maybe one option here is to change pow_i64_i64 to return an int64, but also take an additional `bool& overflowed` flag that gets set if the operation overflowed, in which case we just punt to longPow (instead of pow_i64_i64 do the overflow-handling itself).

src/runtime/long.cpp
```diff
@@ -1191,49 +1192,93 @@ Box* longRTrueDiv(BoxedLong* v1, Box* _v2) {
...
+    } else if (isSubclass(rhs->cls, int_cls)) {
+        rhs_long = boxLong(static_cast<BoxedInt*>(rhs)->n);
     } else {
         return NotImplemented;
     }
+
+    if (mod != None) {
+        if (isSubclass(mod->cls, float_cls)) {
+            raiseExcHelper(TypeError, "pow() 3rd argument not "
+                                      "allowed unless all arguments are integers");
```
kmod: This branch shouldn't get hit, right?

src/runtime/long.cpp
```diff
@@ -1191,49 +1192,93 @@ Box* longRTrueDiv(BoxedLong* v1, Box* _v2) {
...
+
+    if (mpz_sgn(rhs_long->n) == -1) {
+        BoxedFloat* rhs_float = static_cast<BoxedFloat*>(longFloat(rhs_long));
+        BoxedFloat* lhs_float = static_cast<BoxedFloat*>(longFloat(lhs));
+        return boxFloat(pow_float_float(lhs_float->d, rhs_float->d));
+    }
+
+    if (mod != None) {
+        mpz_powm(r->n, lhs->n, rhs_long->n, mod_long->n);
+        if (mpz_sgn(r->n) == 0)
```
kmod: should this be `== 1`?  looks like it's ok since it will fall through to `return r` at the bottom, but I think the sign should never be 0 here.

src/runtime/long.cpp
```diff
@@ -1191,49 +1192,93 @@ Box* longRTrueDiv(BoxedLong* v1, Box* _v2) {
...
+        if (mpz_fits_ulong_p(rhs_long->n)) {
+            uint64_t n2 = mpz_get_ui(rhs_long->n);
+            mpz_pow_ui(r->n, lhs->n, n2);
+        } else {
+            if (mpz_cmp_si(lhs->n, 1l) == 0) {
+                mpz_set_ui(r->n, 1l);
+            } else if (mpz_sgn(lhs->n) == 0) {
+                mpz_set_ui(r->n, 0l);
+            } else if (mpz_cmp_si(lhs->n, -1l) == 0) {
+                mpz_set_si(r->n, -1l);
```
kmod: This should return return something like `if_even ? 1 : -1`

src/runtime/long.cpp
kmod: oh sorry misread, I thought that was looking at `mod_long`

test/tests/long.py
```diff
@@ -57,6 +57,10 @@ def test(a, b):
 print -(1L)
 print 1L**2L
 print 1L**2
+print 0 ** (1 << 100)
+print pow(1 << 30, 1 << 30, 127)
+print pow(1L << 30, 1L << 30, 127)
+print pow(1 << 100, 1 << 100, 1000)
```
kmod: couple more tests:
```
print pow(-1, (1<<100))
print pow(-1, (1<<100) + 1)
print pow(0, (1<<100))
print pow(1, (1<<100))
print pow(5, 3, -7L)
print pow(-5,  3, 7L)
print pow(-5,  3, -7L)
```
(I don't think we had any tests that returned negative numbers)

kmod:
also, please make sure to ping the PR when you'd like it to get reviewed again -- otherwise I'm not sure when it's ready :)

kmod:
awesome!

PR #720, opened 2015-07-18
Allocate all ASTInterpreter instances on the stack and remove the interpreter map.

undingen:
Not having the ASTInterpreter GC allocated improves performance.
I had to add a small asm function in order to produce a special stack frame where we can easily retrieve the ASTInterpreter*,
to replace s_interpreterMaps job. This also make sure that this function really does not get inlined.
The s_interpreterMap was hard to understand and produced several times problems (duplicate entries,...)

This patch contains a hack which limits the number of variables inside a function to 512.
Because we have to make sure the are all on the stack and can't dynamically add more space.
An upcoming patch will remove this limitation and replace it with a stack alloca of the size of the actual number of variables the function uses.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -34,6 +34,7 @@ add_library(PYSTON_OBJECTS OBJECT ${OPTIONAL_SRCS}
 		capi/object.cpp
 		capi/typeobject.cpp
 		codegen/ast_interpreter.cpp
+		codegen/ast_interpreter_exec.S
 		codegen/baseline_jit.cpp
 		codegen/codegen.cpp
 		codegen/compvars.cpp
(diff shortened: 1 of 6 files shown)


undingen:
```
                           e648044ce2266a3fc6:  46dec97fccdef6c0ef:
       django_template.py             3.5s (4)             3.4s (4)  -1.5%
            pyxl_bench.py             3.3s (4)             3.3s (4)  -2.4%
sqlalchemy_imperative2.py             4.1s (4)             4.0s (4)  -2.4%
                  geomean                 3.6s                 3.6s  -2.1%
```

undingen:
Rebased and tried two builds --> booth succeeded.
I think this change should be fine.

undingen:
please review :-)
even if you don't like merging in the hard limit of numbers of variables, it helps because this will be the basis for #736 which gets ride of the limit.

kmod:
lgtm, though yeah I would prefer to hold off on merging this until #736 is ready since it seems like that's close.

kmod:
(closed by #736)

PR #723, opened 2015-07-20
update libpypa

tjhance:


tjhance:
Uh I don't know why this is failing :| `virtualenv_test.py` works fine on my local machine.

kmod:
Ok I have another theory for the parsing failures, and addressed them in #732.  Not sure why the format thing showed up -- looks like it's coming from the network retry code, which might not get exercised that much.  anyway, once I merge #732 I guess we should just try rebasing this again and hoping that it helps.  one flaky test at a time :)

PR #728, opened 2015-07-20
Fix some issues found during a search for another bug

undingen:


kmod:
good catches :)

src/runtime/generator.cpp
```diff
@@ -407,7 +407,7 @@ extern "C" void generatorGCHandler(GCVisitor* v, Box* b) {
 
     if (g->running) {
         v->visitPotentialRange((void**)&g->returnContext,
-                               ((void**)&g->returnContext) + sizeof(g->returnContext) / sizeof(void*));
+                               ((void**)&g->returnContext) + sizeof(*g->returnContext) / sizeof(void*));
```
kmod: Should this be `(void**)g->returnContext`?  ie returnContext is a `Context*`, so scanning `&returnContext` scans the pointer itself when really we want to be scanning its contents.  I'm really not sure, though it feels odd that there is a two-levels-of-indirection difference between the pointer and the size.

undingen: Oh good catch!
I created #746.
Feels a little bit scary that this didn't show up.

PR #731, opened 2015-07-21
A fix to the ast interpreter gc visitor

kmod:


kmod:
I think I introduced this with https://github.com/dropbox/pyston/commit/ef71d4538c66ab3adc462521857edf2830d714db a couple days ago.

kmod:
ok travis-ci is running behind so I'm going to push this without ci-results, since I think this might be causing some of the sporadic gc failures, ex with #723

PR #732, opened 2015-07-21
Fixes

kmod:


kmod:
Merging this since I think it has fewer flakes than master

PR #733, opened 2015-07-21
Revert #708 in order to check if this caused the build failures

undingen:
Revert "rewriter: release allocated scratch space after last use" #708
This reverts commit 9bbc7eb7a21e06dd7543741d211c59fa31478678.


undingen:
Closing, the issue is not caused by this commit

PR #734, opened 2015-07-21
try reverting these

kmod:


kmod:
erf I didn't want to revert these, but if it unbreaks the build...

PR #736, opened 2015-07-21
assign fixed slots (vregs) to the symbols.

undingen:
This removes a bottleneck of the interpreter/bjit:
most var accesses introduced a DenseMap lookup, with this change we use a fixed offset per var.
The bjit stores the pointer to the vregs array inside r14 for fast accesses.

It is based on #720 (=the first commit is the same) and removes the fixed limit of variables.

This improves perf by: (this includes the perf gained by #720)
```
                           f1129a8463c7fdbd56:  7974141245e82dbd99:
      django_template2.py             4.5s (6)             4.1s (4)  -8.6%
            pyxl_bench.py            3.2s (10)             3.1s (4)  -3.9%
sqlalchemy_imperative2.py            3.8s (10)             3.6s (4)  -4.6%
                  geomean                 3.8s                 3.6s  -5.7%
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -34,6 +34,7 @@ add_library(PYSTON_OBJECTS OBJECT ${OPTIONAL_SRCS}
 		capi/object.cpp
 		capi/typeobject.cpp
 		codegen/ast_interpreter.cpp
+		codegen/ast_interpreter_exec.S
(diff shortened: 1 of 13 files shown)


undingen:
This should be ready for a first review.
The first commit is form #720 only the second one is new.

kmod:
awesome!

PR #739, opened 2015-07-22
Min max

Daetalus:
Let `min` and `max` could accept keyword parameter. Extract common part into `min_max`, just as CPython did. Also add some `min` and `max` test.

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -145,30 +145,77 @@ extern "C" Box* any(Box* container) {
     return boxBool(false);
 }
 
-extern "C" Box* min(Box* arg0, BoxedTuple* args) {
+Box* min_max(Box* arg0, BoxedTuple* args, BoxedDict* kwargs, int opid) {
     assert(args->cls == tuple_cls);
+    if (kwargs)
+        assert(kwargs->cls == dict_cls);
 
-    Box* minElement;
+    Box* key_func = nullptr;
+    Box* extremElement;
     Box* container;
+    Box* extremVal;
+
+    if (kwargs && kwargs->d.size()) {
+        static BoxedString* key_str = static_cast<BoxedString*>(PyString_InternFromString("key"));
+        auto it = kwargs->d.find(key_str);
+        if (it != kwargs->d.end() && kwargs->d.size() == 1) {
+            key_func = it->second;
+        } else {
+            if (opid == Py_LT)
+                raiseExcHelper(TypeError, "min() got an unexpected keyword argument");
+            else
+                raiseExcHelper(TypeError, "max() got an unexpected keyword argument");
+        }
+    }
+
     if (args->size() == 0) {
-        minElement = nullptr;
+        extremElement = nullptr;
+        extremVal = nullptr;
         container = arg0;
     } else {
-        minElement = arg0;
+        extremElement = arg0;
+        if (key_func != NULL) {
+            extremVal = runtimeCall(key_func, ArgPassSpec(1), extremElement, NULL, NULL, NULL, NULL);
+        } else {
+            extremVal = extremElement;
+        }
         container = args;
     }
 
+    Box* curVal = nullptr;
     for (Box* e : container->pyElements()) {
-        if (!minElement) {
-            minElement = e;
+        if (key_func != NULL) {
+            if (!extremElement) {
+                extremVal = runtimeCall(key_func, ArgPassSpec(1), e, NULL, NULL, NULL, NULL);
+                extremElement = e;
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -145,30 +145,68 @@ extern "C" Box* any(Box* container) {
...
+    bool pass_kwargs = (kwargs && kwargs->d.size());
+    Box* key_func;
+    Box* extremElement;
     Box* container;
+    Box* extremVal;
+
+    if (pass_kwargs) {
+        static BoxedString* key_str = static_cast<BoxedString*>(PyString_InternFromString("key"));
+        auto it = kwargs->d.find(key_str);
+        key_func = it->second;
```
kmod: we should have a check here that it was in fact "key" that was passed.  ex `min(1, a=1)` will segfault.  cpython checks both that the size of kwargs is 1 and that we found "key" in it.

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -145,30 +145,68 @@ extern "C" Box* any(Box* container) {
...
+
     if (args->size() == 0) {
-        minElement = nullptr;
+        extremElement = nullptr;
+        extremVal = nullptr;
         container = arg0;
     } else {
-        minElement = arg0;
+        extremElement = arg0;
+        if (pass_kwargs) {
```
kmod: As a minor stylistic thing, I think it'd be nicer if this was `if (key_func)` -- it just so happens that the conditions will be true at the same time, but checking `key_func` seems like better one to be checking.

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -145,30 +145,68 @@ extern "C" Box* any(Box* container) {
...
-                throwCAPIException();
-            if (r)
-                minElement = e;
+            if (!extremElement) {
+                extremVal = e;
+                extremElement = e;
+            }
+            curVal = e;
+        }
+        int r = PyObject_RichCompareBool(extremVal, curVal, opid);
```
kmod: If `extremElement`, we probably shouldn't call PyObject_RichCompareBool.  Maybe we could take the `!extremElement` check out the above if/else, and if it's true then `continue`?

test/tests/math_test.py
```diff
@@ -16,6 +16,79 @@
...
+print(max(1, 2, 3, 4, key=lambda x: -x))
+print(max(4, 3, 2, 1, key=lambda x: -x))
+
+print min([[1, 2], [3, 4], [9, 0]], key=lambda x: x[1])
+print min(1.2, 6.3, 6.9, key=int)
+print min("moon", "sun", "earth", key=len)
+
+print max([[1, 2], [3, 4], [9, 0]], key=lambda x: x[1])
+print max(1.2, 6.3, 6.9, key=int)
+print max("moon", "sun", "earth", key=len)
```
kmod: Can we get a test for the invalid-keywords case, and also a test that has a `key` function that prints when it gets evaluated?

kmod:
looks pretty good :) just some minor comments

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -145,30 +145,77 @@ extern "C" Box* any(Box* container) {
...
+        static BoxedString* key_str = static_cast<BoxedString*>(PyString_InternFromString("key"));
+        auto it = kwargs->d.find(key_str);
+        if (it != kwargs->d.end()) {
+            key_func = it->second;
+        } else {
+            if (opid == Py_GT)
+                raiseExcHelper(TypeError, "min() got an unexpected keyword argument");
+            else
+                raiseExcHelper(TypeError, "max() got an unexpected keyword argument");
+        }
```
kmod: We need a check here that the size of kwargs was 1, so that we can reject things like `min([1], key=lambda x:x, extra_arg=1)`.

test/tests/math_test.py
kmod: Sorry, I meant something like:
```python
class C(object):
    def __init__(self, n):
        self.n = n
    def __lt__(self, rhs):
        print "lt", self.n, rhs.n
        return self.n < rhs.n
    def __gt__(self, rhs):
        print "gt", self.n, rhs.n
        return self.n > rhs.n

def key(x):
    print "key", x.n
    return C(-x.n)
print min([C(1), C(3), C(2)], key=key).n
print max([C(1), C(3), C(2)], key=key).n
```

which shows which function calls we make.  looks like we have a small discrepancy with CPython, where our max calls `__lt__` (and min calls `__gt__`) whereas CPython's max calls `__gt__` -- shouldn't really matter that much but you never know if someone is relying on it.

kmod:
awesome!  thanks for sticking through all the corner cases :)

PR #741, opened 2015-07-22
Maybe travis-ci doesn't expand env vars here

kmod:


kmod:
nevermind, looks like it does (pretty cool!).

PR #743, opened 2015-07-22
bjit: omit frame pointer + use R12 for ASTInterpreter*

undingen:
This reduces the number of stack access, because previously when we accessed a field of the interpreter we always had to load it first from stack into a reg - and now we have a dedicated reg.
This is currently only a very small perf change but when #736 lands this becomes more important.

```
       django_template.py             3.3s (4)             3.3s (4)  -0.3%
            pyxl_bench.py             3.2s (4)             3.2s (4)  -0.4%
sqlalchemy_imperative2.py             3.7s (4)             3.8s (4)  +0.4%
                  geomean                 3.4s                 3.4s  -0.1%
```


undingen:
this one is too ready for review

PR #746, opened 2015-07-23
Fix gc visiting code for running generators

undingen:
I verified with gdb that this now really visits all entries of the return context struct


kmod:
nice :)

Issue #749, opened 2015-07-23
Memory leak in finalization ordering


kmod:
@rudi-c 

A malloc leak not a gc leak :P  Here's a test case (make sure to run inside a ulimit):
```python
for i in xrange(1000000):
    unicode("Hello world", "utf8", "strict")
```

I think there are two things happening:
- TraceStack has a leak since we now have one chunk left at the end which we don't free
- we create a large number of TraceStacks, looks like two per `FinalizationState::UNREACHABLE` object.

I played around with it and this seems to work ok:
```C++
~TraceStack() {
    assert(end - cur == CHUNK_SIZE);
    free(start);
}
```

But we might want to also try to reuse the TraceStacks so that we aren't constantly allocating+freeing the chunks.

kmod:
fixed by #752

PR #752, opened 2015-07-23
Release leftover chunk in the TraceStack.

rudi-c:
This also makes the implementation of TraceStack more correct. Though in practice we always a non-zero number of elements to the TraceStack before using it, the old TraceStack would leak memory if it was created but never used.

Addresses https://github.com/dropbox/pyston/issues/749

### DIFF
--- a/src/gc/collector.cpp
+++ b/src/gc/collector.cpp
@@ -120,6 +120,13 @@ public:
             push(p);
         }
     }
+    ~TraceStack() {
+        RELEASE_ASSERT(end - cur == CHUNK_SIZE, "destroying non-empty TraceStack");
+
+        // We always have a block available in case we want to push items onto the TraceStack,
+        // but that chunk needs to be released after use to avoid a memory leak.
+        release_chunk(start);
+    }
 
     void push(void* p) {
         GC_TRACE_LOG("Pushing %p\n", p);


kmod:
Thanks :)  we should keep an eye out for whether the mallocs end up being expensive -- not sure how they compare to the rest of the finalization work.  I found this test since I was trying to optimize it, part of which will involve reducing the number of pycapsules; ie the solution to "finalization involves too many mallocs" might just be "reduce the amount of finalization we have to do" :P

rudi-c:
The TraceStack is stack-allocated and the blocks are always re-used, there should no longer be a performance hit from mallocs.

You're right though, the less finalization work the better.

kmod:
sorry, not sure I follow -- we're still doing a malloc+free per tracestack that we instantiate, and we will instantiate quite a few of them (one per call to `finalizationOrderingFindReachable`, which is called per object in the `UNREACHABLE` state).

rudi-c:
A certain number of chunks (`MAX_FREE_CHUNKS`) are kept around in a free list.

kmod:
oh shoot ok I didn't see that you changed my code to call release_chunk instead of free, you're right that should avoid extra calls to malloc/free.

PR #754, opened 2015-07-23
Release assert for more debugging info.

rudi-c:


kmod:
fyi there's also an ASSERT that takes the same format string but doesn't run in release mode.

kmod:
I cherry-picked this into #755 since I got the error in the travis-ci run; didn't get the error a second time, but for better or worse it seems like just a matter of time until it shows up again :P

PR #756, opened 2015-07-24
fix up our install targets

toshok:
add cmake install() rules for things, and fix up the various probing the from_cpython code
uses to probe for buildir vs. prefix.  with this we're able to install to a prefix and
run pyston from that prefix and get a working repl that can import things.

not ready for merging, just want to see if the longer running tests work.


toshok:
as per our talk, the method (somehow getting ahold of the location of the pyston we built), the method I suggested won't work, since the executable is copied into the virtualenv.

But, this is a problem for all other python implementations, and there's a site.py solution:  virtualenv writes out a file call "orig-prefix.txt".  site.py knows to read that file in if it exists, and sets sys.real_prefix to  the contents.

e.g.:

for the pypy version I downloaded and created a virtualenv of, orig-prefix.txt contains: `/home/toshok/pypy-2.6.0-linux64`

for the system cpython, it contains:
`/usr`

Will try to get this implemented tonight/tomorrow.

PR #757, opened 2015-07-24
Use cpythons lock implementation

undingen:
This switches the thread lock implementation to use a semaphore instead of a mutex.
I hope this gets rid of the threading_local.py error on travis-ci.


undingen:
It's hard to tell if this change really fixed the issue.
At least with this patch applied #720 works without problems...

kmod:
Oh hmm yeah pthreads seems pretty emphatic that you shouldn't unlock a mutex from a different thread -- surprised we've gotten away with this!  We've seen other issues with threading_local.py so maybe this will clean those up too :)

PR #759, opened 2015-07-26
Fix the errors that report in `test format` and re-enable `test format`.

Daetalus:


src/capi/abstract.cpp
```diff
@@ -2008,6 +2008,13 @@ extern "C" PyObject* PyNumber_Long(PyObject* o) noexcept {
     if (o->cls == int_cls)
         return PyLong_FromLong(((BoxedInt*)o)->n);
 
+    PyObject* long_func = PyObject_GetAttrString(o, "__long__");
+    if (long_func) {
+        PyObject* long_obj = PyEval_CallObject(long_func, NULL);
+        Py_DECREF(long_func);
+        return long_obj;
+    }
+
```
kmod: Can we copy CPython's implementation of PyNumber_Long?  They have a number of features in there that, while I'm sure we can port, seems like it'd be nice to just copy.  (ex throwing an exception if `__long__` returned something that wasn't a long)

kmod:
awesome, thanks :)

PR #760, opened 2015-07-26
bjit: add support for most common missing nodes and don't JIT compile cold blocks

undingen:
- don't jit cold blocks after doing a OSR JIT compilation
 - previously after doing a OSR JIT compilation we continued to JIT every block outside of the loop.
This doesn't show up as a perf change but reduces the number of JITed code / makes it slightly smaller.
- add the most common missing AST nodes to the baseline JIT
 - with this PR I'm seeing only 6 aborts for ```django_template.py```, 1 abort for ```pyxl_bench.py``` and none for ```sqlalchemy_imperative.py```


undingen:
```
      django_template2.py             4.0s (4)             4.0s (4)  -0.1%
            pyxl_bench.py             3.0s (4)             3.0s (4)  -1.4%
sqlalchemy_imperative2.py             3.3s (4)             3.2s (4)  -0.8%
                  geomean                 3.4s                 3.4s  -0.8%
```

undingen:
mmh travis-ci doen't want to check this PR???

PR #763, opened 2015-07-27
Numeric binary operator support for old style class

Daetalus:


src/runtime/classobj.cpp
```diff
@@ -735,6 +735,69 @@ static Box* instanceNext(BoxedInstance* inst) {
...
+            return NULL;
+        PyErr_Clear();
+        PyErr_Format(TypeError, "unsupported operand type(s) for ** or pow(): '%s' and '%s'", getTypeName(self),
+                     getTypeName(arg1));
+        return NULL;
+    }
+    res = PyEval_CallObject(func, args);
+    Py_DECREF(func);
+    return res;
+}
```
kmod: I'm not sure I understand the role of these functions -- they seem named after the typeobject.c functions, but 
behave a bit more like instance_pow.  Regardless, maybe in this situation we should just copy instance_pow and instance_ipow from CPython?  Your slot_nb_power seems reasonable (though a bit hard to verify all the corner cases); I think inplace_power is missing the fallback to `__pow__` and `__rpow__`.

src/runtime/classobj.cpp
```diff
@@ -735,6 +735,182 @@ static Box* instanceNext(BoxedInstance* inst) {
...
+    if (args == NULL) {
+        Py_DECREF(func);
+        return NULL;
+    }
+    result = PyEval_CallObject(func, args);
+    Py_DECREF(args);
+    Py_DECREF(func);
+    return result;
+}
+
```
Daetalus: Is that ok to copy these code? Or use a "Pyston way" function? Such as `generic_binary_op` could replaced by `_instanceBinary` function. But for other functions, such as `half_binop` and `do_pow`, it contains logic. So either rewrite it ( like the last version of commit), or call the CPython version.

kmod:
I think copying the cpython code is the right approach here -- we can go back and make changes to it if we need to, but basically it just saves so many headaches to use code that we know is right :)  We might want at some point to reduce the duplication by having our functions call the cpython ones you just copied.

PR #765, opened 2015-07-27
implementing _ast.__version__ for installing 'flask'

corona10:
implementing _ast.__version__ for installing 'flask'

### DIFF
--- a/src/runtime/builtin_modules/ast.cpp
+++ b/src/runtime/builtin_modules/ast.cpp
@@ -30,6 +30,7 @@
 #include "runtime/types.h"
 #include "runtime/util.h"
 
+#define AST_VERSION "82160"
 namespace pyston {
 
 static BoxedClass* AST_cls;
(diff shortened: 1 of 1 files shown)


corona10:
same error occured both Cpython and pyston
```
 Exception:
    File "/home/corona/pyston/test/integration/virtualenv/virtualenv_support/pip-6.0.8-py2.py3-none-any.whl/pip/basecommand.py", line 232, in main
    File "/home/corona/pyston/test/integration/virtualenv/virtualenv_support/pip-6.0.8-py2.py3-none-any.whl/pip/commands/install.py", line 339, in run
    File "/home/corona/pyston/test/integration/virtualenv/virtualenv_support/pip-6.0.8-py2.py3-none-any.whl/pip/req/req_set.py", line 435, in prepare_files
    File "/home/corona/pyston/test/integration/virtualenv/virtualenv_support/pip-6.0.8-py2.py3-none-any.whl/pip/_vendor/pkg_resources/__init__.py", line 2496, in requires
    File "/home/corona/pyston/test/integration/virtualenv/virtualenv_support/pip-6.0.8-py2.py3-none-any.whl/pip/_vendor/pkg_resources/__init__.py", line 2552, in __getattr__
  AttributeError: _dep_map
```

what is wrong.. i don't want to say unit test is wrong..

kmod:
Hmm I guess it might be hard to do this just for flask for now -- I think the issue is something like distutils will try to use the ast module, but if it doesn't exist it will fall back on another one (which we support).  We could probably patch our distutils, or maybe we could just do our testing with a modified flask?

Issue #689, opened 2015-07-10
Infinite loops in recursive lists


rudi-c:
e.g.
```
l = range(5)
l[0] = l
print l
```

prints something like
```
[[...], 1, 2, 3, 4]
```
in CPython.

Currently segfaults in Pyston due to stack overflow.

kmod:
I'm marking this "probably easy" since CPython has a pretty straightforward way of handling this that I think we can copy.

kmod:
cool, looks like list is taken care of :)  some grepping shows that cpython does this also for dict, set, and (surprisingly, though they have an explanation) tuple.

kmod:
I think this is fixed now.

PR #766, opened 2015-07-27
implement recursive printing of set and tuples.

corona10:
#689

implement recursive printing of set and tuples.

### DIFF
--- a/src/runtime/set.cpp
+++ b/src/runtime/set.cpp
@@ -99,6 +99,14 @@ static Box* _setRepr(BoxedSet* self, const char* type_name) {
     std::string O("");
     llvm::raw_string_ostream os(O);
 
+    int status = Py_ReprEnter((PyObject*)self);
+    if (status != 0) {
+        if (status < 0)
+            return boxString(os.str());
+
+        os << type_name << "(...)";
+        return boxString(os.str());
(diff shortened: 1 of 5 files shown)


kmod:
Can we get some tests?  The tuple case is a bit hard to reproduce but I think this works:
```python
class C(object):
    def __init__(self):
        self.t = (self,)
    def __repr__(self):
        return repr(self.t)
print repr(C())
```

I think the set one can be triggered using the same test case but switching the tuple to a set.

corona10:
@kmod 
in my guess set can be infinate set theorically but Cpython implementation prohibit this kind of set.

kmod:
Ah ok nice, catch, didn't realize that we currently allowed hashing sets :)  Can we have a test for that though?

Also, I think we can test the set-recursion case using the same thing as the tuple test but with a set:
```python
class C(object):
    def __init__(self):
        self.t = set([self])
    def __repr__(self):
        return repr(self.t)
print repr(C())
```

kmod:
looks good!  just a minor comment -- in the future let's avoid using `# should_error` and we could probably just include these tests in one of the existing test files.

fhaseen419-byte:
Hello what you say about github

PR #767, opened 2015-07-28
Import the ctypes module.

rudi-c:
This PR allows us to compile the ctypes module and run `import ctypes`. Functionality has not been tested. This is part of WIP to run NumPy.

### DIFF
--- a/from_cpython/Modules/_ctypes/_ctypes.c
+++ b/from_cpython/Modules/_ctypes/_ctypes.c
@@ -2044,13 +2044,21 @@ PyCSimpleType_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
             break;
         }
 
-        if (ml) {
+        // TODO: Pyston change:
+        // For now, don't run this code path because we don't support some of
+        // the descriptor CAPI functions.
+        // We do this to be able to run `import ctypes`, but this will need to be
+        // enabled again once we want CType to actually work.
+        // if (ml) {
+        if (false) {
 #if (PYTHON_API_VERSION >= 1012)
             PyObject *meth;
             int x;
+            /*
             meth = PyDescr_NewClassMethod(result, ml);
             if (!meth)
                 return NULL;
+            */
 #else
 #error
             PyObject *meth, *func;
--- a/from_cpython/Modules/_ctypes/cfield.c
+++ b/from_cpython/Modules/_ctypes/cfield.c
@@ -1291,7 +1291,9 @@ s_get(void *ptr, Py_ssize_t size)
      */
     slen = strlen(PyString_AS_STRING(result));
     size = min(size, (Py_ssize_t)slen);
-    if (result->ob_refcnt == 1) {
+
+    // Pyston change: no ob_refcnt
+    if (false /*result->ob_refcnt == 1*/) {
         /* shorten the result */
         _PyString_Resize(&result, size);
         return result;
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -3236,10 +3236,9 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
                        | Py_TPFLAGS_DICT_SUBCLASS | Py_TPFLAGS_BASE_EXC_SUBCLASS | Py_TPFLAGS_TYPE_SUBCLASS;
     RELEASE_ASSERT((cls->tp_flags & ~ALLOWABLE_FLAGS) == 0, "");
     if (cls->tp_as_number) {
-        RELEASE_ASSERT(cls->tp_flags & Py_TPFLAGS_CHECKTYPES, "Pyston doesn't yet support non-checktypes behavior");
+        // RELEASE_ASSERT(cls->tp_flags & Py_TPFLAGS_CHECKTYPES, "Pyston doesn't yet support non-checktypes behavior");
     }
 
-    RELEASE_ASSERT(cls->tp_descr_set == NULL, "");
     RELEASE_ASSERT(cls->tp_free == NULL || cls->tp_free == PyObject_Del || cls->tp_free == PyObject_GC_Del, "");
     RELEASE_ASSERT(cls->tp_is_gc == NULL, "");
     RELEASE_ASSERT(cls->tp_mro == NULL, "");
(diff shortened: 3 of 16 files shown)


from_cpython/Modules/_ctypes/cfield.c
```diff
@@ -719,6 +719,9 @@ vBOOL_get(void *ptr, Py_ssize_t size)
 }
 #endif
 
+// Pyston change: We don't run CPython's ./configure so we didn't have this defined.
+#define SIZEOF__BOOL 1
```
kmod: Could we just put this in our pyconfig.h?

rudi-c: True, that is a better place.

src/capi/typeobject.cpp
```diff
@@ -3316,8 +3316,24 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
...
+
+        // If this class is a metaclass defined in a C extension, chances are that some of its
+        // instances may be hardcoded in the C extension as well. Those instances will call
+        // PyType_Ready and expect their class (this metaclass) to have a place to put attributes.
+        // e.g. CTypes does this.
+        assert(cls->instancesHaveHCAttrs() || cls->instancesHaveDictAttrs());
+    } else {
+        // this should get automatically initialized to 0 on this path:
+        assert(cls->attrs_offset == 0);
+    }
```
kmod: Maybe we should just always inherit attrs_offset?  I think this will affect any case that an extension class inherits from a Pyston/Python class.

src/runtime/objmodel.cpp
```diff
@@ -5446,6 +5420,100 @@ Box* typeNew(Box* _cls, Box* arg1, Box* arg2, Box** _args) {
...
+// {
+//     ...
+//     result = (PyTypeObject *)PyType_Type.tp_new(type, args, kwds);
+//     ...
+// }
+//
+// Assigned to the tp_new of PyCSimpleType, a metaclass. By calling PyType_Type.tp_new,
+// we don't want to do an attribute lookup for __new__, because that would end up calling
+// the tp_new of the subclass PyCSimpleType (PyCSimpleType_new again) and end up in a loop.
+Box* typeNewType(BoxedClass* metatype, Box* args, Box* kwds) noexcept {
```
kmod: I think this should be called `type_new` (ie using the cpython convention for cpython-like functions)

src/capi/typeobject.cpp
```diff
@@ -3235,10 +3235,10 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
                        | Py_TPFLAGS_DICT_SUBCLASS | Py_TPFLAGS_BASE_EXC_SUBCLASS | Py_TPFLAGS_TYPE_SUBCLASS;
     RELEASE_ASSERT((cls->tp_flags & ~ALLOWABLE_FLAGS) == 0, "");
     if (cls->tp_as_number) {
-        RELEASE_ASSERT(cls->tp_flags & Py_TPFLAGS_CHECKTYPES, "Pyston doesn't yet support non-checktypes behavior");
+        // RELEASE_ASSERT(cls->tp_flags & Py_TPFLAGS_CHECKTYPES, "Pyston doesn't yet support non-checktypes behavior");
     }
 
-    RELEASE_ASSERT(cls->tp_descr_set == NULL, "");
+    // RELEASE_ASSERT(cls->tp_descr_set == NULL, "");
```
kmod: I think we can delete this if we just copy the TPSLOT defs for tp_descr_set.

src/capi/typeobject.cpp
```diff
@@ -3235,10 +3235,10 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
                        | Py_TPFLAGS_DICT_SUBCLASS | Py_TPFLAGS_BASE_EXC_SUBCLASS | Py_TPFLAGS_TYPE_SUBCLASS;
     RELEASE_ASSERT((cls->tp_flags & ~ALLOWABLE_FLAGS) == 0, "");
     if (cls->tp_as_number) {
-        RELEASE_ASSERT(cls->tp_flags & Py_TPFLAGS_CHECKTYPES, "Pyston doesn't yet support non-checktypes behavior");
+        // RELEASE_ASSERT(cls->tp_flags & Py_TPFLAGS_CHECKTYPES, "Pyston doesn't yet support non-checktypes behavior");
```
kmod: This worries me :)  would it be possible to change the ctypes code that relies on having this support?

from_cpython/Modules/_ctypes/_ctypes.c
```diff
@@ -2044,13 +2044,20 @@ PyCSimpleType_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
             break;
         }
 
-        if (ml) {
+        // TODO: Pyston change:
+        // For now, don't run this code path because we don't support descriptors.
```
kmod: fyi we support descriptors in general, it looks like we just don't support a couple of the capi functions for creating some special types of them.  though we do support those types (members, getsets, and classmethods).

kmod:
could we get a test that just imports ctypes?

src/capi/typeobject.cpp
rudi-c: Is it hard to support checktypes behavior? That would be nicer than patching ctypes.

rudi-c:
Alright, all changes made, tests pass again.

from_cpython/Modules/_ctypes/_ctypes.c
```diff
@@ -2044,13 +2044,21 @@ PyCSimpleType_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
             break;
         }
 
-        if (ml) {
+        // TODO: Pyston change:
```
kmod: I think it's best if we can avoid these kinds of "soft" failures, since it'll potentially be much harder to debug why we get a `no attribute "from_param"` than a "hard" error like "Py_FatalError: unimplemented".  I tried, though, to think of some other way to get around this, and I couldn't think of anything better so let's just go with this and hope we remember that this is here :)

kmod:
Nice, this now supports the small usage we have in the dropbox codebase :)

PR #770, opened 2015-07-28
bjit: fix problems when setting bjit thresholds to 1

undingen:
- we hit an assert when we ran out of scratch space
- 'classobjSetattr' returned void but all runtime funcs must return a Box* object.
- sneak in a change which removes a unused arg from '_setupCall'
- skip retrying to jit very large blocks
- register iterator types with the correct size
 - I checked all calls to ```BoxedHeapClass::create```

We may want to add a test run which forces to use the baseline JIT like we do for the other tiers...

This shows a perf change with our default thresholds which surprises me:
```
                           75e203a2dfac5578fc:  dbc4b9d2ff6f1b31a2:
      django_template3.py             3.6s (4)             3.3s (4)  -7.2%
            pyxl_bench.py             2.9s (4)             2.9s (4)  +1.4%
sqlalchemy_imperative2.py             3.3s (4)             3.2s (4)  -0.4%
                  geomean                 3.2s                 3.2s  -2.1%
```

### DIFF
--- a/src/asm_writing/rewriter.cpp
+++ b/src/asm_writing/rewriter.cpp
@@ -819,7 +819,7 @@ RewriterVar* Rewriter::call(bool has_side_effects, void* func_addr, const Rewrit
     return result;
 }
 
-void Rewriter::_setupCall(RewriterVar* result, bool has_side_effects, const RewriterVar::SmallVector& args,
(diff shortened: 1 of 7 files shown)


kmod:
nice :)

PR #772, opened 2015-07-28
dictionary recursive printing

corona10:
might be last code for #689

### DIFF
--- a/src/runtime/dict.cpp
+++ b/src/runtime/dict.cpp
@@ -31,7 +31,22 @@ using namespace pyston::ExceptionStyle;
 using pyston::ExceptionStyle::ExceptionStyle;
 
 Box* dictRepr(BoxedDict* self) {
+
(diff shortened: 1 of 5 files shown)


kmod:
looks good, though I took a closer look and I think there are a couple things we need to do -- if `PyRepr_Enter` returns a negative number, that means that some sort of CAPI exception was thrown, so we should call `throwCAPIException()` instead of returning the current partial repr contents.

Also, we should add a try-catch block (or better, use C++ destructors) to make sure that we call PyRepr_Leave if an exception was thrown between PyRepr_Enter and PyRepr_Leave.

corona10:
i added try-catch block on all chnaged files and remove raw_ostream.(it might be better performance..)
but build failed because of http://llvm.org is down.. haha might be try later
```

The command "curl -sSL "http://llvm.org/apt/llvm-snapshot.gpg.key" | sudo -E apt-key add -" failed and exited with 2 during .
```

PR #774, opened 2015-07-29
Recursive printing

corona10:
add try catch block and remove raw_ostream

### DIFF
--- a/src/runtime/dict.cpp
+++ b/src/runtime/dict.cpp
@@ -31,57 +31,48 @@ using namespace pyston::ExceptionStyle;
 using pyston::ExceptionStyle::ExceptionStyle;
 
 Box* dictRepr(BoxedDict* self) {
-
-    try {
-
-        std::vector<char> chars;
-        int status = Py_ReprEnter((PyObject*)self);
-
-        if (status != 0) {
+    std::vector<char> chars;
+    int status = Py_ReprEnter((PyObject*)self);
+    if (status != 0) {
+        try {
             if (status < 0)
                 throwCAPIException();
-
-            chars.push_back('{');
-            chars.push_back('.');
-            chars.push_back('.');
-            chars.push_back('.');
-            chars.push_back('}');
-
-            return boxString(llvm::StringRef(&chars[0], chars.size()));
+        } catch (ExcInfo e) {
+            Py_ReprLeave((PyObject*)self);
+            throw e;
         }
 
-
         chars.push_back('{');
+        chars.push_back('.');
+        chars.push_back('.');
+        chars.push_back('.');
+        chars.push_back('}');
+        return boxString(llvm::StringRef(&chars[0], chars.size()));
+    }
 
-        bool first = true;
-
-        for (const auto& p : self->d) {
-
-            if (!first) {
-                chars.push_back(',');
-                chars.push_back(' ');
-            }
-
-            first = false;
-
+    chars.push_back('{');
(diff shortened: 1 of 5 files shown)


kmod:
sorry I might have mislead you -- the try-catch should be around the part of the code where it calls `repr()` on all of the elements.  ie once we see that PyRepr_Enter had succeeded, we need to make sure that we call the final PyRepr_Leave even if the code in between throws an exception.

Also, the catch block should call `throw e` and not `setCAPIException(e); return NULL`.  Sorry for the complexity around exception handling, hopefully we will clean that up soon or add some linters to help.

corona10:
@kmod thanks, i did as you commented :-)

src/runtime/dict.cpp
```diff
@@ -31,25 +31,57 @@ using namespace pyston::ExceptionStyle;
...
 
 Box* dictRepr(BoxedDict* self) {
-    std::vector<char> chars;
-    chars.push_back('{');
-    bool first = true;
-    for (const auto& p : self->d) {
-        if (!first) {
-            chars.push_back(',');
+
+    try {
```
kmod: Minor, but the try blocks should start after the `if (status != 0)` block is done -- ie we should only do it if the Py_ReprEnter call was hit and succeeded.  Technically the try block should also end right before the call to PyRepr_Leave in case that or boxString throw exceptions (though since those functions happen to not through exceptions it's less important).

Also, can we delete some of these new blank lines?

corona10: @kmod i think that i have some misunderstanding about exception mechanism... in conclusion you mean try catch block only need when Py_ReprEnter failed or call repr()??

corona10:
i changed some code..

src/runtime/dict.cpp
```diff
@@ -32,6 +32,24 @@ using pyston::ExceptionStyle::ExceptionStyle;
...
     std::vector<char> chars;
+    int status = Py_ReprEnter((PyObject*)self);
+    if (status != 0) {
+        try {
+            if (status < 0)
+                throwCAPIException();
+        } catch (ExcInfo e) {
+            Py_ReprLeave((PyObject*)self);
+            throw e;
+        }
```
kmod: I don't think we should have this try-catch block (will call Py_ReprLeave when we shouldn't).

src/runtime/list.cpp
```diff
@@ -63,36 +61,41 @@ extern "C" PyObject* PyList_AsTuple(PyObject* v) noexcept {
...
-    os << '[';
     for (int i = 0; i < self->size; i++) {
-        if (i > 0)
-            os << ", ";
 
+        if (i > 0) {
+            chars.push_back(',');
+            chars.push_back(' ');
+        }
         Box* r = self->elts->elts[i]->reprICAsString();
```
kmod: This can throw so we need to put it in a try-catch block -- I'd start the try right after the end of the `if (status != 0)` block, and put the end (ie the catch) right before the normal call to Py_ReprLeave.

corona10:
Ah.. I understand what you saying.
 sorry for misunderstanding you commented.
maybe I should study about error handling mechanism.

I m soldier now(maybe return to university next year) so I should return to base today night so I will fix it tonight  before return to base and that will be my last commit for few month haha

corona10:
@kmod i re-commit them sorry :-)

kmod:
awesome, thanks :)

PR #783, opened 2015-07-29
Some work on the NumPy test

rudi-c:
Depends on #767 

These are just a few small changes that help run NumPy which should be useful independently of the rest of the NumPy work.

### DIFF
--- a/from_cpython/Include/Python.h
+++ b/from_cpython/Include/Python.h
@@ -134,6 +134,13 @@ PyObject* PyGC_AddRoot(PyObject*) PYSTON_NOEXCEPT;
 //      PyGC_AddRoot(obj);
 // to reduce any chances of compiler reorderings or a GC somehow happening between the assignment
 // to the static slot and the call to PyGC_AddRoot.
+//
+
+// Similarly, some extension modules have static non-type objects that our GC thinks are errors
+// since they are not in the heap, unless we root them. This function should be used purely as
+// a temporary patching mechanism - we want to have a better way of dealing with such objects
+// in the future (even if it's an invalid use of CPython APIs).
+PyObject* PyGC_AddNonHeapRoot(PyObject* obj, int size) PYSTON_NOEXCEPT;
 
 // Pyston change : expose these type objects
 extern PyTypeObject Pattern_Type;
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -3278,6 +3278,19 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
 
     assert(cls->tp_name);
 
+    // Inherit some special protocols. Normally methods are automatically inherited,
+    // when a Python class is declared, but that may not be the case with C extensions.
+    if (base != NULL) {
+        if (cls->tp_as_number == NULL)
+            cls->tp_as_number = base->tp_as_number;
+        if (cls->tp_as_sequence == NULL)
+            cls->tp_as_sequence = base->tp_as_sequence;
+        if (cls->tp_as_mapping == NULL)
+            cls->tp_as_mapping = base->tp_as_mapping;
+        if (cls->tp_as_buffer == NULL)
+            cls->tp_as_buffer = base->tp_as_buffer;
+    }
+
     if (cls->tp_call) {
         cls->tpp_call.capi_val = tppProxyToTpCall<CAPI>;
         cls->tpp_call.cxx_val = tppProxyToTpCall<CXX>;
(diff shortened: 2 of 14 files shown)


rudi-c:
@kmod This got rebased on the ctypes changes and should be ready for review now.

src/capi/typeobject.cpp
```diff
@@ -3267,6 +3267,19 @@ extern "C" int PyType_Ready(PyTypeObject* cls) noexcept {
...
+    if (base != NULL) {
+        if (cls->tp_as_number == NULL)
+            cls->tp_as_number = base->tp_as_number;
+        if (cls->tp_as_sequence == NULL)
+            cls->tp_as_sequence = base->tp_as_sequence;
+        if (cls->tp_as_mapping == NULL)
+            cls->tp_as_mapping = base->tp_as_mapping;
+        if (cls->tp_as_buffer == NULL)
+            cls->tp_as_buffer = base->tp_as_buffer;
+    }
```
kmod: hmm this is a bit odd to me -- why do we have to do this when cpython doesn't?

rudi-c: It is in PyType_Ready, typeobject.cpp:4123

src/runtime/dict.cpp
```diff
@@ -397,6 +400,18 @@ Box* dictDelitem(BoxedDict* self, Box* k) {
     return None;
 }
 
+// Analoguous to CPython's, used for sq_ slots.
+static int dict_ass_sub(PyDictObject* mp, PyObject* v, PyObject* w) {
+    Box* res;
+    if (w == NULL) {
+        res = dictDelitem((BoxedDict*)mp, v);
+    } else {
+        res = dictSetitem((BoxedDict*)mp, v, w);
```
kmod: We should wrap these in a try-catch-setCapiException.  it's also nice to put a "noexcept" specifier on the method as a whole, since that will get the compiler to enforce that the function doesn't end up throwing anything (by adding try-catch-std::terminate() wrappers).

src/runtime/list.cpp
```diff
@@ -253,6 +255,12 @@ extern "C" Box* listGetslice(BoxedList* self, Box* boxedStart, Box* boxedStop) {
     return _listSlice(self, start, stop, 1, stop - start);
 }
 
+// Analoguous to CPython's, used for sq_ slots.
+static PyObject* list_item(PyListObject* a, Py_ssize_t i) {
+    BoxedList* self = (BoxedList*)a;
+    return listGetitemUnboxed(self, i);
```
kmod: same here with the exception handling.

kmod:
Sorry, we really should get around to adding an exception-style linter :/

btw, how hard would it be to get it to the point that we can actually enable the numpy test?

src/capi/typeobject.cpp
kmod: ah ok, right (not sure why I didn't see that).

rudi-c:
There's almost nothing to do besides adding a line to the script that applies a patch to NumPy.

rudi-c:
That being said the NumPy test currently only tests if NumPy crashes, not if it produces the same output.

src/codegen/parser.cpp
```diff
@@ -384,6 +384,13 @@ AST_DictComp* read_dictcomp(BufferedReader* reader) {
     return rtn;
 }
 
+AST_Ellipsis* read_ellipsis(BufferedReader* reader) {
+    // TODO:
```
kmod: Is there actually anything to do here?  I think we should avoid putting this in if it will silently handle ellipses the wrong way, but it looks like this might actually be all we need and we can remove the TODO?  Same for the serializer.

src/core/cfg.cpp
```diff
@@ -1932,6 +1932,11 @@ class CFGVisitor : public ASTVisitor {
         return true;
     }
 
+    bool visit_ellipsis(AST_Ellipsis* node) override {
+        // TODO:
+        return true;
+    }
```
kmod: My recollection of the cfg code isn't great, but I think Ellipsis should be handled in remapExpr rather than as a top-level visit function (I'm surprised you didn't have to add that).

kmod: Or you could just not add it there and then we will abort when we try to execute a function containing it, which sounds like it is good enough for now.  But I think we don't need this visit function.

from_cpython/Include/Python.h
```diff
@@ -134,6 +134,8 @@ PyObject* PyGC_AddRoot(PyObject*) PYSTON_NOEXCEPT;
 //      PyGC_AddRoot(obj);
 // to reduce any chances of compiler reorderings or a GC somehow happening between the assignment
 // to the static slot and the call to PyGC_AddRoot.
+//
```
kmod: Can we add a quick comment here saying that this is a pyston change and what it's for?

kmod:
Ok I merged the testsuite change, could you update this PR to include the new submodule commit?

rudi-c:
Cool this is passing again.

kmod:
awesome!

PR #784, opened 2015-07-29
Refactors and cleanup in objmodel.cpp

rudi-c:
Mostly removing some old functions and moving a few functions and hidden classes out of objmodel.cpp


kmod:
Thanks for the cleanup :)

PR #785, opened 2015-07-30
object.__str__ use capi slot in order to increase compatibility with cpython

undingen:
With this change we print True for:
```
class C(object):
    pass
print C.__str__ is object.__str__
```
Cheetah depends on this behavior (see https://bitbucket.org/pypy/compatibility/wiki/cheetah)

### DIFF
--- a/src/runtime/builtin_modules/thread.cpp
+++ b/src/runtime/builtin_modules/thread.cpp
@@ -163,6 +163,17 @@ public:
             PyThread_free_lock(self->lock_lock);
         }
     }
+
+    static Box* locked(Box* _self) {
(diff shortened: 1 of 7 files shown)


kmod:
Oh goodie:
```
>> print type(None).__str__ is None.__str__
True
```

I think the issue is that we have a problem binding descriptors to None, since we use `object == None` to check whether the object was passed or not.  We might need to switch from using the `__get__` attribute to using tp_descr_get, which could take NULL.

ex:
```
print None.__str__ # binds to None
print type(None).__str__.__get__(None, type(None)) # doesn't bind to None!
```

undingen:
I updated this one, I'm not sure why I had to implement ```locked()``` but without this I always got a build failure...

PR #787, opened 2015-07-30
Add some function to old style class

Daetalus:
**Note:**

* The `__complex__` could call `complexNew` in `complex.cpp` correctly. But `complexNew` need some improvements to support `str` and custom type.
* The behaviors of `__unicode__`, `__get__`, `__set__`, `__delete__` are exact same as in CPython. IMHO, there is no need to write additional code in `classobj.cpp`.
* `__missing__` only work in subclass of dict. So no need to write it in `classobj.cpp`.
* `__coerce__` in old style class could call the new added `coerce` builtin function correctly. But the `coerce` need further improvements.

**Update**
* According `import types; print sorted(types.InstanceType.__dict__.keys())` in CPython. Pyston has extra `__base__`.
* `__doc__`, `__new__` seems as instance's class attributes in Pyston. If these three attribute were fine, then the old style class support was completed.
* `__getslice__` and some other functions didn't change. Because if we copy `instance_slice` from CPython, then apply some change, it will very similar as the current one. But I will try to copy CPython code in next PR as much as possible.

### DIFF
--- a/src/runtime/classobj.cpp
+++ b/src/runtime/classobj.cpp
@@ -488,6 +488,15 @@ Box* instanceGetitem(Box* _inst, Box* key) {
     return runtimeCall(getitem_func, ArgPassSpec(1), key, NULL, NULL, NULL, NULL);
 }
 
+Box* instanceReversed(Box* _inst) {
+    RELEASE_ASSERT(_inst->cls == instance_cls, "");
+    BoxedInstance* inst = static_cast<BoxedInstance*>(_inst);
+
+    static BoxedString* reversed_str = internStringImmortal("__reversed__");
+    Box* reversed_func = _instanceGetattribute(inst, reversed_str, true);
+    return runtimeCall(reversed_func, ArgPassSpec(0), NULL, NULL, NULL, NULL, NULL);
+}
+
 Box* instanceSetitem(Box* _inst, Box* key, Box* value) {
     RELEASE_ASSERT(_inst->cls == instance_cls, "");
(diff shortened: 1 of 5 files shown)


kmod:
Oh hmm you're right, there's no need to add a bunch of the ones that I had mentioned.  I should have done this instead: `import types; print sorted(types.InstanceType.__dict__.keys())`

Looks like maybe we should remove these reversed, complex, enter, and exit, and add `__index__`.  Sorry!  Otherwise the change looks good.

src/runtime/classobj.cpp
```diff
@@ -506,6 +515,76 @@ Box* instanceDelitem(Box* _inst, Box* key) {
...
+            throw e;
+    }
+
+    if (getslice_func == NULL) {
+        Box* slice = static_cast<Box*>(createSlice(i, j, None));
+        return instanceGetitem(inst, slice);
+    }
+
+    return runtimeCall(getslice_func, ArgPassSpec(2), i, j, NULL, NULL, NULL);
+}
```
kmod: Minor (not worth changing), but we could have directly copied CPython's instance_slice.  It looks like the conversion got rid of some overhead, but I've been liking the approach of starting out with a 100% copy and then optimize as needed.

kmod:
Awesome!  Looks like we can now close #254 :+1:

fyi github doesn't send any notifications when you add new commits to a pull request, so if you leave a comment on it I will get back to you faster :)

PR #790, opened 2015-07-31
Make Raise and Subscript exprs use capi exceptions

kmod:
And add an "always think capi exceptions are beneficial" mode to force-test that a bit more.

I'm pretty worried about all the new rewriting going in.


kmod:
```
      django_template3.py             3.4s (2)             3.4s (2)  -0.4%
            pyxl_bench.py             3.0s (2)             3.0s (2)  -0.1%
sqlalchemy_imperative2.py             3.4s (2)             3.4s (2)  +0.0%
                  geomean                 3.3s                 3.2s  -0.2%
```

PR #793, opened 2015-08-03
Some improvements to make test unary could pass

Daetalus:
__Change:__
* Add `__pos__` to `float` and `complex`
* Add compare support to `complex`
* Use new `callattr` replace old `getclsattrInternal` to find attribute, add error checking.


kmod:
cool :)

PR #794, opened 2015-08-03
getiter: use tp_iter for non python types

undingen:
```
                           ea3daebf08d007db3e:  7075b3803b12f58f8f:
      django_template3.py             3.3s (4)             3.3s (4)  -0.3%
            pyxl_bench.py             2.8s (4)             2.8s (4)  -1.3%
sqlalchemy_imperative2.py             3.2s (4)             3.2s (4)  -2.3%
                  geomean                 3.1s                 3.1s  -1.3%
```

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -992,7 +992,7 @@ public:
         return new BoxedEnumerate(range.begin(), range.end(), idx);
     }
 
-    static Box* iter(Box* _self) {
+    static Box* iter(Box* _self) noexcept {
         assert(_self->cls == enumerate_cls);
         BoxedEnumerate* self = static_cast<BoxedEnumerate*>(_self);
         return self;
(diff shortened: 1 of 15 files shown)


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -1478,6 +1478,7 @@ void setupBuiltins() {
     enumerate_cls->giveAttr("__hasnext__",
                             new BoxedFunction(boxRTFunction((void*)BoxedEnumerate::hasnext, BOXED_BOOL, 1)));
     enumerate_cls->freeze();
+    enumerate_cls->tp_iter = BoxedEnumerate::iter;
```
kmod: I think sharing the code between the capi (`tp_iter`) and c++ (`__iter__` attribute) versions is fine since none of these functions should throw any exceptions, but maybe we could add noexcept to these functions?

undingen:
I added the noexcept specifiers.

kmod:
thanks :)  I'm trying to think of a way to have callers also specify the exception-style they expect their callee to have, and then hopefully get the type system to enforce a lot of this so we don't have to think about it any more.

PR #795, opened 2015-08-04
Enable future builtins module and add test file

### LINKED ISSUE #777: Import future_builtins module
It would be nice to be able to run `import future_builtins`

I think we're not building or not finding `from_cpython/Module/future_builtins.c`, we shouldn't need to re-implement the whole module ourselves.

Daetalus:
closes #777


kmod:
nice :)

PR #796, opened 2015-08-04
WIP: bjit: register functions with GDB

undingen:
Tell GDB the function name and EH table of the jited function
this is extremely slow [0]  and should only be used for debugging and the implementation is dangerous...

Tried different routes, elf template, using LLVM MC layer, but I only got this one to work :-(.
This maybe okay if it's disabled per default and one can enable it if needed.
Currently WIP but would like to know what if you guys think it's worth to clean this up.

[0] probably: https://code.google.com/p/v8-wiki/wiki/GDBJITInterface
GDB side of JIT Interface currently (as of GDB 7.2) does not handle registration of code objects very effectively. Each next registration takes more time: with 500 registered objects each next registration takes more than 50ms, with 1000 registered code objects - more than 300 ms. This problem was reported to GDB developers (http://sourceware.org/ml/gdb/2011-01/msg00002.html) but currently there is no solution available.

### DIFF
--- a/src/codegen/baseline_jit.cpp
+++ b/src/codegen/baseline_jit.cpp
@@ -29,6 +29,25 @@
 #include "runtime/types.h"
 #include "runtime/util.h"
 
+// GDB JIT API
+extern "C" {
+typedef enum { JIT_NOACTION = 0, JIT_REGISTER_FN, JIT_UNREGISTER_FN } jit_actions_t;
+struct jit_code_entry {
+    struct jit_code_entry* next_entry;
+    struct jit_code_entry* prev_entry;
+    const char* symfile_addr;
+    uint64_t symfile_size;
+};
+struct jit_descriptor {
+    uint32_t version;
+    uint32_t action_flag;
+    struct jit_code_entry* relevant_entry;
+    struct jit_code_entry* first_entry;
+};
+void __attribute__((noinline)) __jit_debug_register_code();
+extern struct jit_descriptor __jit_debug_descriptor;
+}
+
 namespace pyston {
 
 static llvm::DenseSet<CFGBlock*> blocks_aborted;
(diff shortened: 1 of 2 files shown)


kmod:
Would it be easier if we scale back to only trying to get gdb to not barf?  ie instead of trying to get the right function names (which we don't get already with the interpreter, so I think it's ok) we just try to make sure that the stackframes are right and that gdb doesn't start backtracing randomly through memory.  I don't know why it doesn't get the rbp-based unwinding right, but maybe there's some trick we could do to convince it to do that, or maybe we could add one of those custom jit callbacks that would force it to use that.

kmod:
oh, sorry didn't realize that we are omitting the frame pointer for bjit code.  maybe we could try saving it and see if that helps gdb?

undingen:
yes we omit the frame pointer but even with frame pointers the stacktrace was wrong :-(.
AFAIK omitting it hasn't made things worse.

I didn't think that using this interface is so slow. The bad perf of this one make the new JIT gdb plugin interface look nicer. I just don't like that the plugin has to be GPL licensed and manually loaded...

There is also this mozilla bugreport where the encounter the same issue with gdb backtraces on x64: https://bugzilla.mozilla.org/show_bug.cgi?id=642320:
_Yes, both x86 and x86_64 preserve the frame pointer. Unfortunately, on x86_64 it doesn't do any good -- you can't get a backtrace with a JITted method on the stack, apparently because the x86_64 ABI says that you can't depend on having the frame pointer and so must use unwind info for this instead. I don't understand why gdb can't detect that no unwind information is available and fall back to using the frame pointer (before giving up entirely), and I suspect there's a worthwhile patch to gdb to fix that. jimb might know more._

Maybe there is a way to let gdb know about our EH tables....

PR #797, opened 2015-08-04
Hash improvement for complex, str and long.

Daetalus:
**Change**
* Pypy parser could not handle complex literal correctly, it always treat complex as 0.0j. Fix it. The code originally write by Marius.
* Add hash function to complex. Already tested. (But did not enable test_hash, please see below)
* Fix the wrong `mpz` usage in `long` constructor.
* Treat the hash value of empty string as Zero, like CPython did.

**Note**
After read CPython code. CPython str hash function will check `PYTHONHASHSEED` enviroment varialble in `random.c` file. But the Pyston DJB based str hash function didn't support it. I did some experiments try to fix it. But seems I couldn't use random number in str hash function (`types.h`). Pyston will report error:
```
pyston: ../../src/runtime/builtin_modules/thread.cpp:189: void pyston::setupThread():
Assertion `thread_module' failed.
```

### DIFF
--- a/src/runtime/complex.cpp
+++ b/src/runtime/complex.cpp
@@ -228,6 +228,33 @@ static void _addFunc(const char* name, ConcreteCompilerType* rtn_type, void* com
     complex_cls->giveAttr(name, new BoxedFunction(cl));
 }
 
+Box* complexHash(BoxedComplex* self) {
+    if (!isSubclass(self->cls, complex_cls))
+        raiseExcHelper(TypeError, "descriptor '__hash__' requires a 'complex' object but received a '%s'",
+                       getTypeName(self));
+    long hashreal, hashimag, combined;
+
+    hashreal = _Py_HashDouble(self->real);
+    if (hashreal == -1) {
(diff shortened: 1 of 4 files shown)


src/runtime/complex.cpp
```diff
@@ -228,6 +228,29 @@ static void _addFunc(const char* name, ConcreteCompilerType* rtn_type, void* com
...
 }
 
+Box* complexHash(BoxedComplex* self) {
+    if (!isSubclass(self->cls, complex_cls))
+        raiseExcHelper(TypeError, "descriptor '__hash__' requires a 'complex' object but received a '%s'",
+                       getTypeName(self));
+    long hashreal, hashimag, combined;
+    hashreal = _Py_HashDouble(self->real);
+    if (hashreal == -1)
+        return boxInt(-1);
```
kmod: these should be `throwCapiException()` -- returning -1 is the capi way for hash functions to specify that they threw an exception.

kmod:
awesome thanks :)

PR #801, opened 2015-08-05
Move our integration tests to a separate repository

kmod:
which can be found at https://github.com/dropbox/pyston-testsuite

The goal is to reduce the number of repositories that people have to clone when they just want to try to build pyston (and make it more palatable for us to add more of them), so I moved all of the submodules we use for building into the new build_deps/ directory, and moved all the submodules we use for testing into the new testsuite repository.

This doesn't help by itself since I also added the testsuite repository as a submodule (there's really no other reasonable way to do it), so I updated the install docs so that people will by default only clone the submodules in the build_deps directory.

on a side note, git submodules are the worst.  I corrupted my git state multiple times in the process of this merge, and ran into lots of situations where git would happily tell me to use the wrong command.  For instance, did you know that `git add` is not the proper way to handle submodule conflicts?  The command you are looking for is `git reset`.  If you are silly enough to try to `git mv` a directory that has a submodule in it, you will break your checkout in addition to breaking `git status`.  :angry:


kmod:
I think this change is borked since the version of git I used has some bugs related to submodule moving :(

PR #803, opened 2015-08-05
Move our integration tests to a separate repository

kmod:
Take 2

Same thing as #801, but now with 50% less chance of repository corruption.

I hate submodules.


kmod:
Ok, I'm going to go ahead and merge this.  Everyone will probably have to do a `git submodule update --init --recursive` once switching to this branhc.

I think I managed to avoid the corruption issues, but if you run into weird issues with libunwind or jemalloc, typically during the "Rerunning CMake" phase, you may have to try some combination of `git submodule deinit -f . ; git submodule update --init --recursive` and if that isn't enough, `rm -rfv build`.

Have I mentioned that I'm not happy with submodules?

PR #809, opened 2015-08-06
add __doc__ attribute to capifunc

### LINKED ISSUE #776: AttributeError: 'capifunc' object has no attribute '__doc__'
Sample test:
```
from ctypes import addressof
print type(addressof)
addressof.__doc__
```

This should require adding the right attribute in `runtime/capi.cpp` to `capifunc_cls`.

Daetalus:
closes #776

### DIFF
--- a/src/capi/types.h
+++ b/src/capi/types.h
@@ -54,6 +54,14 @@ public:
         return None;
     }
 
+    static Box* doc(Box* b, void*) {
+        RELEASE_ASSERT(b->cls == capifunc_cls, "");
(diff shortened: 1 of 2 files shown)


kmod:
Cool :)

fyi there's also a "BoxedMemberDescriptor" class that is nice for doing this sort of exposing-c-attributes-to-python -- you don't have to write as much code and it's more optimizable than the opaque function pointers in the getset.  Looks like CPython uses a getset though.

PR #810, opened 2015-08-06
Refactor location and names of gc visit functions.

rudi-c:
Make them static functions on the type that the visit function is associated to whenever possible. Makes it more consistent and less likely to forget to update the visit function if we change the type's fields.


kmod:
cool :)  yeah those had gotten pretty inconsistent.

PR #811, opened 2015-08-06
more capi exceptions

kmod:
- add+use capi variants of runtimeCall and callattr entry points
- add some simple profiling to determine when to use C++ vs CAPI exceptions

Sorry for the huge PR; the early commits can stand on their own but they have a perf hit that is only made up by the later ones, so I didn't feel like separating into multiple prs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1260,6 +1260,12 @@ $(FROM_CPYTHON_SRCS:.c=.prof.o): %.prof.o: %.c $(BUILD_SYSTEM_DEPS)
 	$(ECHO) Compiling C file to $@
 	$(VERB) $(CC_PROFILE) $(EXT_CFLAGS_PROFILE) -c $< -o $@ -g -MMD -MP -MF $(patsubst %.o,%.d,$@)
 
+.PHONY: update_section_ordering
+update_section_ordering: pyston_release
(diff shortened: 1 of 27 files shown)


kmod:
So I haven't been able to reduce the cost of adding the templates:
```
      django_template3.py             3.4s (4)             3.4s (4)  +0.2%
            pyxl_bench.py             2.8s (4)             2.9s (4)  +1.1%
sqlalchemy_imperative2.py             3.2s (4)             3.3s (4)  +1.5%
                  geomean                 3.1s                 3.2s  +0.9%
```

But the use of them makes things break even [these stats are from "template-only changes" to "all changes"]:
```
      django_template3.py             3.4s (4)             3.3s (6)  -2.5%
            pyxl_bench.py             2.9s (4)             2.9s (6)  +0.4%
sqlalchemy_imperative2.py             3.3s (4)             3.3s (6)  -0.8%
                  geomean                 3.2s                 3.1s  -1.0%
```

I think there's still more gains that this will unlock (ex once we have python functions be able to throw capi exceptions we will see more improvements).  We also have some capi-exception-expecting cases that come from extension code, and we can also update the bjit to use capi exceptions.  So I'm going to stop trying to hunt down why the templates hurt so much, and move on to those other improvements.

PR #813, opened 2015-08-07
Rewrite string hash and enable test_hash

Daetalus:
**Changes:**
* Copy `random.c` from CPython plus some modifications
* Add `-R` option to Pyston, let Pyston could use random seed.
* Copy the CPython `str` and `unicode` hash function's code to `str.cpp` and do some Pyston changes.
* Enable `test_hash` and some changes to let other part use new string hash function.


kmod:
nice :)

PR #814, opened 2015-08-08
Be able to jit functions that throw CAPI exceptions

kmod:
A lot of our remaining C++ exceptions are being thrown from one Python frame and caught by another.  We handle things caught by Python frames well, and this adds the ability to propagate them as CAPI exceptions.

I'm not stressing about the minor perf impact of these changes too much, but at least this set looks (randomly?) positive:
```
      django_template3.py             3.3s (4)             3.3s (4)  -0.5%
            pyxl_bench.py             2.9s (4)             2.9s (4)  -0.6%
sqlalchemy_imperative2.py             3.3s (4)             3.2s (4)  -1.0%
                  geomean                 3.1s                 3.1s  -0.7%
```


kmod:
Oh I should mention, this new throw-capi-exceptions gets turned on if `function_name == "next"` which is surprisingly effective.

PR #817, opened 2015-08-10
Complex improment

### LINKED ISSUE #779: Full support for complex numbers
Our support for complex number (`complex.cpp`) is currently rather limited. We don't have most of the math attributes implemented (e.g. `__pow__`, `conjugate`, etc).

Although we could reimplement all of those ourselves, this is not an easy task. CPython's `complexobject.c`, has a lot of tricky edge case handling around numeric errors (divisions by 0, etc). It's worth considering using CPython's implementation completely (possibly with some minor patches), like we do for unicode strings.

Daetalus:
**Change:**
* Add some basic complex function implementations.
* Adjust some error checking, raise exception rather than use `assert`. Let the behavior consistent with CPython.
* Rewrite `complexNew`. Let it could accept **keyword** arguments, such as `complex(real=xx, imag=yy)`. And also accept `comlex`, `long`, `str`, `unicode` as `real` or `imag` argument. ~~But could not handle custom class with `__complex__` yet.~~
* Rewrite `complexFmt` based on CPython implementation. Let it could remove empty decimal point.
* Some small `complex` related changes.

closes #779

### DIFF
--- a/from_cpython/Lib/test/test_complex.py
+++ b/from_cpython/Lib/test/test_complex.py
@@ -1,4 +1,3 @@
-# expected: fail
 import unittest
 from test import test_support
 
@@ -102,8 +101,7 @@ class ComplexTest(unittest.TestCase):
                            complex(random(), random()))
 
         self.assertRaises(ZeroDivisionError, complex.__div__, 1+1j, 0+0j)
-        # FIXME: The following currently crashes on Alpha
-        # self.assertRaises(OverflowError, pow, 1e200+1j, 1e200+1j)
+        self.assertRaises(OverflowError, pow, 1e200+1j, 1e200+1j)
 
     def test_truediv(self):
         self.assertAlmostEqual(complex.__truediv__(2+0j, 1+1j), 1-1j)
@@ -435,12 +433,15 @@ class ComplexTest(unittest.TestCase):
         test_values = (1, 123.0, 10-19j, xcomplex(1+2j),
                        xcomplex(1+87j), xcomplex(10+90j))
 
+        # Pyston change: if rhs is a subclass of lhs, then should try
+        # reverse the order. Pyston don't support it yet. Need to improve
+        # binop handing.
         for op in infix_binops:
             for x in xcomplex_values:
                 for y in test_values:
                     a = 'x %s y' % op
                     b = 'y %s x' % op
-                    self.assertTrue(type(eval(a)) is type(eval(b)) is xcomplex)
+                    # self.assertTrue(type(eval(a)) is type(eval(b)) is xcomplex)
 
     def test_hash(self):
         for x in xrange(-30, 30):
@@ -476,8 +477,10 @@ class ComplexTest(unittest.TestCase):
         self.assertEqual(repr(complex(0, -INF)), "-infj")
         self.assertEqual(repr(complex(0, NAN)), "nanj")
 
-    def test_neg(self):
-        self.assertEqual(-(1+6j), -1-6j)
+    # Pyston change: This is a libpypa bug, waiting for upstream fix it
+    # please refer libpypa#47
+    # def test_neg(self):
+    #     self.assertEqual(-(1+6j), -1-6j)
 
     def test_file(self):
         a = 3.33+4.43j
--- a/src/runtime/complex.cpp
+++ b/src/runtime/complex.cpp
@@ -30,19 +30,74 @@ extern "C" Box* createPureImaginary(double i) {
     return new BoxedComplex(0.0, i);
 }
 
-extern "C" Py_complex PyComplex_AsCComplex(PyObject* op) noexcept {
-    // TODO: Incomplete v.s. CPython's implementation.
-    Py_complex cval;
-    if (PyComplex_Check(op)) {
-        cval.real = ((BoxedComplex*)op)->real;
-        cval.imag = ((BoxedComplex*)op)->imag;
-        return cval;
-    } else if (op->cls == int_cls) {
-        cval.real = ((BoxedInt*)op)->n;
-        cval.imag = 0.0;
-        return cval;
+static PyObject* try_complex_special_method(PyObject* op) noexcept {
+    PyObject* f;
+    static PyObject* complexstr;
+
+    if (complexstr == NULL) {
+        complexstr = PyString_InternFromString("__complex__");
+        if (complexstr == NULL)
+            return NULL;
+    }
+    if (PyInstance_Check(op)) {
+        f = PyObject_GetAttr(op, complexstr);
+        if (f == NULL) {
+            if (PyErr_ExceptionMatches(PyExc_AttributeError))
+                PyErr_Clear();
+            else
+                return NULL;
+        }
     } else {
-        Py_FatalError("unimplemented");
+        f = _PyObject_LookupSpecial(op, "__complex__", &complexstr);
+        if (f == NULL && PyErr_Occurred())
+            return NULL;
+    }
+    if (f != NULL) {
+        PyObject* res = PyObject_CallFunctionObjArgs(f, NULL);
+        Py_DECREF(f);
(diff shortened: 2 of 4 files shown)


src/runtime/classobj.cpp
```diff
@@ -1516,6 +1525,7 @@ void setupClassobj() {
     instance_cls->giveAttr("__int__", new BoxedFunction(boxRTFunction((void*)instanceInt, UNKNOWN, 1)));
     instance_cls->giveAttr("__long__", new BoxedFunction(boxRTFunction((void*)instanceLong, UNKNOWN, 1)));
     instance_cls->giveAttr("__float__", new BoxedFunction(boxRTFunction((void*)instanceFloat, UNKNOWN, 1)));
+    instance_cls->giveAttr("__complex__", new BoxedFunction(boxRTFunction((void*)instanceComplex, UNKNOWN, 1)));
```
kmod: It looks like CPython doesn't have their equivalent of this -- are you sure we need this?

src/runtime/complex.cpp
```diff
@@ -199,25 +228,65 @@ extern "C" Box* complexDiv(BoxedComplex* lhs, Box* rhs) {
...
 }
 
 // str and repr
-// For now, just print the same way as ordinary doubles.
-// TODO this is wrong, e.g. if real or imaginary part is an integer, there should
-// be no decimal point, maybe some other differences. Need to dig deeper into
-// how CPython formats floats and complex numbers.
-// (complex_format in Objects/complexobject.c)
-std::string complexFmt(double r, double i, int precision, char code) {
+Box* complexFmt(double r, double i, int precision, char format_code) noexcept {
```
kmod: Would you mind if we called this `complex_fmt` -- I think it's nice to use CPython's snake_case naming when using their exception style, and use our camelCase naming when using our exceptions.

src/runtime/complex.cpp
```diff
@@ -228,6 +297,98 @@ static void _addFunc(const char* name, ConcreteCompilerType* rtn_type, void* com
     complex_cls->giveAttr(name, new BoxedFunction(cl));
 }
 
+static BoxedComplex* ComplexPowComplex(BoxedComplex* a, BoxedComplex* b) {
```
kmod: tiny, but this would be more consistent as `complexPowComplex`.

src/runtime/complex.cpp
```diff
@@ -346,9 +814,151 @@ static PyObject* complex_richcompare(PyObject* v, PyObject* w, int op) noexcept
...
+
+extern "C" Box* complexEq(BoxedComplex* lhs, Box* _rhs) {
+    if (lhs->cls != complex_cls) {
+        raiseExcHelper(TypeError, "descriptor '__eq__' requires a 'complex' object but recieved a '%s'",
+                       getTypeName(lhs));
+    }
+    BoxedComplex* rhs = convert_to_complex(_rhs);
+    Box* res = NULL;
+    if (rhs != NULL) {
+        res = complex_richcompare(lhs, rhs, Py_EQ);
```
kmod: I think these need to have a `if (!res) throwCapiException()`

kmod:
Hmm ok so this is a lot of code that I don't have any understanding of, and a bunch of it is structured a little bit differently than CPython does it.  I think we have two main options:
- keep doing it this way and make sure we have some good testing (enabling test_complex)
- switch to the CPython implementation of these functions, or maybe even the entire CPython implementation

If we were starting from scratch I'd advocate for the second option, but since you've gone pretty far down the first path I think it'd be fine to continue with it.  I still think it'll be less total work to just copy CPython's implementation (for example, we pass different format arguments for complexRepr... I have no idea what the impact there is), but I'll leave it up to you.

src/runtime/classobj.cpp
Daetalus: Please see `test_complex`, line 228, there has class `class OS`, seems it could only work if I add `__complex__` to `classobj.cpp`.

from_cpython/Lib/test/test_complex.py
```diff
@@ -440,7 +438,7 @@ def __rmod__(self,other):
                 for y in test_values:
                     a = 'x %s y' % op
                     b = 'y %s x' % op
-                    self.assertTrue(type(eval(a)) is type(eval(b)) is xcomplex)
+                    # self.assertTrue(type(eval(a)) is type(eval(b)) is xcomplex)
```
Daetalus: If `type(y)` is `complex`, `type(x)` is `xcomplex`:

`type(pow(y, x))` and `type(x ** y)`
===> `xcomplex`.
`type(y ** x)`
===> `complex`
It should be `xcomplex`.

Would you mind to give some tips? FYI, originally I use my own implementation of `complexPow`, the error occured and could not figure out why. Then I switch to the pure CPython implementation. Same problem.

src/runtime/classobj.cpp
kmod: Well, most oldstyle-class attribute lookups get automatically proxied.  It's just for the attributes that have capi slots where we need to provide a wrapper implementation, and I don't think there's one for `__complex__`.  Could you try the test without this and see if it works?

from_cpython/Lib/test/test_complex.py
kmod: This is a bug / missing feature in our binop handling; usually the rule is you try `lhs.__pow__(rhs)` and then try `rhs.__rpow__(lhs)`, *except* if rhs is a subclass of lhs, and then you reverse the order.  We don't support that reverse-order case, and I don't think it's specific to complex or pow.  It'd be great if you wanted to look into it (we would want to rewrite it, which could be a good intro to that stuff), but I think for this PR it's fine to comment out the check.

from_cpython/Lib/test/test_complex.py
```diff
@@ -476,8 +474,8 @@ def test_repr(self):
         self.assertEqual(repr(complex(0, -INF)), "-infj")
         self.assertEqual(repr(complex(0, NAN)), "nanj")
 
-    def test_neg(self):
-        self.assertEqual(-(1+6j), -1-6j)
+    # def test_neg(self):
+    #     self.assertEqual(-(1+6j), -1-6j)
```
Daetalus: This is `libpypa` bug, Already solved in local repository, and create a PR https://github.com/vinzenz/libpypa/pull/47. Maybe libpypa will use another solution, but it will be solved at last. When it done, re-enable it.

from_cpython/Lib/test/test_complex.py
Daetalus: I will try to look at the reverse-order binop. But recently I am busy(one or two weeks). But I will continue to improve the compatibility in my spare time.

from_cpython/Lib/test/test_complex.py
kmod: Can you add this as a note in the code?

kmod: And same thing for the other changes to this file -- can we have `Pyston change` comments with them explaining why it differs from CPython.

src/runtime/float.cpp
```diff
@@ -124,6 +125,8 @@ extern "C" Box* floatAdd(BoxedFloat* lhs, Box* rhs) {
         return floatAddFloat(lhs, static_cast<BoxedFloat*>(rhs));
     } else if (rhs->cls == long_cls) {
         return boxFloat(lhs->d + PyLong_AsDouble(rhs));
+    } else if (rhs->cls == complex_cls) {
+        return complexAdd(new BoxedComplex(lhs->d, 0), rhs);
```
kmod: Do we need these?  Looks like cpython returns NotImplemented for `(1.0).__add__(1j)`.  Probably not a huge deal either way but was just curious if it meant that we are missing something somewhere else.

kmod:
awesome :)

PR #819, opened 2015-08-10
microoptimizations

kmod:
- improve bool and str hash() speed
- optimize str.find(char) [ie use faster "find char in string" method over "find string in string" method]


kmod:
```
      django_template3.py             3.3s (4)             3.3s (4)  -0.2%
            pyxl_bench.py             2.9s (4)             2.8s (4)  -4.0%
sqlalchemy_imperative2.py             3.2s (4)             3.2s (4)  +0.2%
                  geomean                 3.2s                 3.1s  -1.3%
```

PR #820, opened 2015-08-11
Refactors on types of GC objects

rudi-c:
This will be useful to support more C++ objects that need to be tracked or allocated in our heap.


kmod:
cool :)

PR #826, opened 2015-08-12
Introduction of slice ast type & updated libpypa

vinzenz:
As per libpypa request https://github.com/vinzenz/libpypa/issues/12


kmod:
Cool, thanks for the fix :)

PR #831, opened 2015-08-13
Don't crash on non-mapping like arguments on kwargs. Fixes #805

### LINKED ISSUE #805: `f(**kw)` crashes if kw is not a dict-like object
Supposed to throw a TypeError instead.

vinzenz:


src/runtime/dict.cpp
```diff
@@ -596,21 +596,33 @@ extern "C" Box* dictNew(Box* _cls, BoxedTuple* args, BoxedDict* kwargs) {
...
+
+    return true;
+}
+
+bool dictMergeKWArgs(BoxedDict* self, Box* other) {
+    return dictMergeImpl(self, other);
+}
+
+void dictMerge(BoxedDict* self, Box* other) {
+    assert(dictMergeImpl(self, other));
```
kmod: putting this in the assert looks iffy

vinzenz: Well then your previous code was also iffy :P Because this is actually keeping that old behaviour there.

vinzenz: however, wait, it might not be called in release....

src/runtime/dict.cpp
```diff
@@ -596,21 +596,33 @@ extern "C" Box* dictNew(Box* _cls, BoxedTuple* args, BoxedDict* kwargs) {
...
             self->d[p.first] = p.second;
-        return;
+        return true;
     }
 
     static BoxedString* keys_str = internStringImmortal("keys");
     CallattrFlags callattr_flags{.cls_only = false, .null_on_nonexistent = true, .argspec = ArgPassSpec(0) };
     Box* keys = callattr(other, keys_str, callattr_flags, NULL, NULL, NULL, NULL, NULL);
-    assert(keys);
+    if (!keys) {
```
kmod: So the reason this was an assert is because CPython actually does something different in this case -- it calls PyDict_MergeFromSeq2 -- and we shouldn't just ignore that behavior.  Can we copy in that function and then call it if there was no "keys" attribute?

vinzenz: A comment there would have been better than a plain assert then ;)

kmod: ok sorry that's not quite right -- in the kwargs case (`ext_do_call` in ceval.c) they call PyDict_Update which calls PyDict_Merge directly.  The dict.update method (and `__init__`) calls `dict_update_common`, which tries either PyDict_Merge or PyDict_MergeFromSeq2 depending on whether the "keys" attribute exists.  We actually have this distinction I guess, so the assert should have just been an exception.  sorry about that.

src/runtime/objmodel.cpp
```diff
@@ -3293,8 +3301,9 @@ void rearrangeArguments(ParamReceiveSpec paramspec, const ParamNames* param_name
         getArg(kwargs_idx, oarg1, oarg2, oarg3, oargs) = okwargs;
     }
 
-    if ((!param_names || !param_names->takes_param_names) && argspec.num_keywords && !paramspec.takes_kwargs) {
-        raiseExcHelper(TypeError, "%s() doesn't take keyword arguments", func_name);
+    if ((!param_names || !param_names->takes_param_names) && (argspec.num_keywords || argspec.has_kwargs)
```
kmod: Not sure if this is the cause of the test failures, but I think it's perfectly allowable to pass kwargs to a non-keyword-taking function if the kwargs dict is empty.

src/runtime/objmodel.cpp
```diff
@@ -3180,9 +3180,17 @@ void rearrangeArguments(ParamReceiveSpec paramspec, const ParamNames* param_name
...
-        Box* given_varargs = getArg(argspec.num_args + argspec.num_keywords, arg1, arg2, arg3, args);
-        for (Box* e : given_varargs->pyElements()) {
-            varargs.push_back(e);
+        Box* given_varargs = nullptr;
+        try {
+            given_varargs = getArg(argspec.num_args + argspec.num_keywords, arg1, arg2, arg3, args);
+            for (Box* e : given_varargs->pyElements()) {
+                varargs.push_back(e);
+            }
+        } catch (ExcInfo const& e) {
```
kmod: Minor, but I think our unwinder requires that we catch exceptions by value; maybe the compiler translates the catch-by-const-reference to a catch-by-value but it's probably better to just have it explicitly catch by value.

src/runtime/objmodel.cpp
```diff
@@ -3266,7 +3274,7 @@ void rearrangeArguments(ParamReceiveSpec paramspec, const ParamNames* param_name
         Box* ovarargs = BoxedTuple::create(unused_positional.size(), &unused_positional[0]);
         getArg(varargs_idx, oarg1, oarg2, oarg3, oargs) = ovarargs;
     } else if (unused_positional.size()) {
-        raiseExcHelper(TypeError, "%s() takes at most %d argument%s (%ld given)", func_name, paramspec.num_args,
+        raiseExcHelper(TypeError, "%s() takes exactly %d argument%s (%ld given)", func_name, paramspec.num_args,
```
kmod: I'm not sure that either of these error messages is quite correct; maybe it should be `paramspec.num_defaults ? "at most" : "exactly"`

src/runtime/objmodel.cpp
```diff
@@ -3366,8 +3387,17 @@ void rearrangeArguments(ParamReceiveSpec paramspec, const ParamNames* param_name
...
+        int takes = paramspec.num_args;
+        if (!(paramspec.takes_varargs || paramspec.takes_kwargs)) {
+            raiseExcHelper(TypeError, "%s() takes exactly %d argument%s (%ld given)", func_name, takes,
+                           takes != 1 ? "s" : "", given);
+        } else {
+            if (given < takes) {
+                raiseExcHelper(TypeError, "%s() takes at least %d argument%s (%ld given)", func_name, takes,
+                               takes != 1 ? "s" : "", given);
+            }
+        }
```
kmod: What's supposed to happen if it falls through here and doesn't throw an exception?

src/runtime/objmodel.cpp
```diff
@@ -3293,8 +3301,15 @@ void rearrangeArguments(ParamReceiveSpec paramspec, const ParamNames* param_name
...
-    if ((!param_names || !param_names->takes_param_names) && argspec.num_keywords && !paramspec.takes_kwargs) {
-        raiseExcHelper(TypeError, "%s() doesn't take keyword arguments", func_name);
+    Box* kwargs = nullptr;
+    if (argspec.has_kwargs) {
+        kwargs
+            = getArg(argspec.num_args + argspec.num_keywords + (argspec.has_starargs ? 1 : 0), arg1, arg2, arg3, args);
+    }
+    if ((!param_names || !param_names->takes_param_names) && (argspec.num_keywords || argspec.has_kwargs)
+        && (!paramspec.takes_kwargs
+            && (kwargs && PyDict_Check(kwargs) && static_cast<BoxedDict*>(kwargs)->d.size() > 0))) {
```
kmod: I don't think this check should ignore the existence of `num_keywords` due to properties of the passed kwargs dict; looks like `dir(a=1)` causes a crash.

I'd be a fan of moving this logic to when we actually try to place a keyword; ie instead of "determine before-hand whether we will have any keywords", we could do "when we try to handle a keyword, if there's nowhere to put it throw the appropriate exception".  Maybe a middle ground is to keep the existing check (for non-kwargs parameters), and then just change the existing kwargs assert to an exception?

src/runtime/objmodel.cpp
vinzenz: ok, I didn't know that it is required because of that. Good to know. Will fix that

src/runtime/objmodel.cpp
vinzenz: I see, I couldn't recall to have seen 'at most' before. I will make the ternary like you suggested

src/runtime/objmodel.cpp
vinzenz: Well if you have *args or **kwargs then those args are passed as a tuple/mapping to the called function. So if you have more parameters passed that should be ok. What is not ok is that there are less arguments than required.

kmod: I don't think that's quite what's going on here -- this block is checking to make sure that all of the positional arguments received values.  Whether we passed more values into `*args` or `**kw` shouldn't matter.  Here's an example:
```python
def f(a, **kw):
  print "called function"
f(b=1, c=2)
```
With this change we will now allow the function call to succeed, when instead we should be throwing an exception.

PR #832, opened 2015-08-13
Allow string + bytearray => bytearray. Fixes #780

### LINKED ISSUE #780: string + bytearray coercion
In CPython:
```
>>> "abc" + bytearray("abc")
bytearray(b'abcabc')
```

In Pyston:
```
>> "abc" + bytearray("abc")
  File "<stdin>", line 1, in <module>:
TypeError: cannot concatenate 'str' and 'bytearray' objects
```

vinzenz:


src/runtime/str.cpp
```diff
@@ -341,7 +341,11 @@ extern "C" Box* strAdd(BoxedString* lhs, Box* _rhs) {
         return rtn;
     }
 
-    if (!PyString_Check(_rhs)) {
+    if (PyByteArray_Check(_rhs)) {
```
kmod: super minor, but could we move this block inside the `if (!PyString_Check)` block?  I doubt this happens very often so it'd be nice to try avoid doing the check.

vinzenz: Done

kmod:
awesome :)

PR #833, opened 2015-08-13
support yield statements inside lambdas

undingen:


tjhance:
this is a thing?

wut

src/capi/abstract.cpp
```diff
@@ -1650,7 +1650,7 @@ extern "C" int PyNumber_Check(PyObject* obj) noexcept {
     assert(obj && obj->cls);
 
     // Our check, since we don't currently fill in tp_as_number:
-    if (PyFloat_Check(obj) || PyFloat_Check(obj) || PyFloat_Check(obj))
```
kmod: good catch, thanks!

kmod:
wow, I am very surprised that this works.

PR #837, opened 2015-08-14
Some improvements for int to let test_int could pass

Daetalus:


kmod:
I think https://github.com/Daetalus/pyston/pull/2 should fix it -- I think this was an existing bug that your code just happened to trigger for the first time.

kmod:
closing+reopening to force it to re-merge

src/runtime/int.cpp
```diff
@@ -320,8 +320,8 @@ extern "C" Box* div_i64_i64(i64 lhs, i64 rhs) {
 #if PYSTON_INT_MIN < -PYSTON_INT_MAX
     static_assert(PYSTON_INT_MIN == -PYSTON_INT_MAX - 1, "");
 
-    if (lhs == PYSTON_INT_MIN && rhs == -1) {
-        return longDiv(boxLong(lhs), boxLong(rhs));
+    if (lhs == PYSTON_INT_MIN) {
```
kmod: I think this fix isn't quite right -- it fixes the cpython test but because they only test against `lhs == INT_MIN`.  We have the same overflow problem with other cases, such as `-sys.maxint / 4`.  I think we should leave the `INT_MIN / -1` case as a special case (since it extra-special), but we'll need some other fix for when `lhs - rhs` overflows.

test/CPYTHON_TEST_NOTES.md
```diff
@@ -41,23 +41,19 @@ test_extcall            f(**kw) crashes if kw isn't a dict
...
-test_hash               number of hash bugs (all representations of '1' don't have the same hash; empty string is supposed to have 0 hash, etc)
 test_index              slice.indices, old-styl-class __mul__
-test_int                we assert instead of throw exception
 test_io                 memory/gc issue?
 test_iterlen            [unknown]
 test_itertools          [unknown]
 test_json               'from test.script_helper import assert_python_ok' fails; sounds like it is trying to look at pycs
 test_list               longs as slice indices
 test_long               sys.long_info
-test_math               float(long); sys.float_info, sys.displayhook
```
kmod: thanks for updating this :)

kmod:
Awesome :)  I'm going to file an issue for the int division overflow thing; it didn't seem worth blocking the PR for an existing bug.

PR #840, opened 2015-08-15
Switch our Python set implementation to use a llvm::DenseSet

undingen:
This doen't have any significant perf change on our main benchmarks.

### DIFF
--- a/src/runtime/set.cpp
+++ b/src/runtime/set.cpp
@@ -29,15 +29,9 @@ void BoxedSet::gcHandler(GCVisitor* v, Box* b) {
     Box::gcHandler(v, b);
 
     BoxedSet* s = (BoxedSet*)b;
-
-    // This feels like a cludge, but we need to find anything that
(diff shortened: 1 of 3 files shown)


src/runtime/set.cpp
```diff
@@ -489,10 +483,9 @@ Box* setHash(BoxedSet* self) {
     RELEASE_ASSERT(isSubclass(self->cls, frozenset_cls), "");
 
     int64_t rtn = 1927868237L;
-    for (Box* e : self->s) {
-        BoxedInt* h = hash(e);
-        assert(PyInt_Check(h));
-        rtn ^= h->n + 0x9e3779b9 + (rtn << 6) + (rtn >> 2);
+    for (auto&& e : self->s) {
+        assert(e.hash == unboxInt(hash(e.value)));
```
kmod: I think we can take this out -- looks like CPython will use the stored hash value as well.

otherwise lgtm!

PR #844, opened 2015-08-17
Use builtin function names instead of '<builtin function>'

vinzenz:


kmod:
I just merged #845, does that help?  fyi it's difficult to review/merge a PR like this that contains commits from another unmerged PR; I wish github had better support for that but it doesn't :/

PR #848, opened 2015-08-18
Undo testsuite split

kmod:
Move everything back into the main repo.


kmod:
Sorry for the churn, but I decided to go forward with this.

PR #853, opened 2015-08-19
Use adjustment-post division to avoid int division overflow

### LINKED ISSUE #850: Int division overflow issue
@Daetalus Our division code has some overflow issues when `lhs - rhs` overflows:

```
>> import sys
>> print sys.maxint / -2
4611686018427387904
>> print -sys.maxint / 2
-4611686018427387904
```

Daetalus:
closes #850


kmod:
nice :)

PR #854, opened 2015-08-20
Fix and reenable type speculations

undingen:
```
                           3e360e58e74c6f1a2b:  b7bb155d168f8b40f0:
      django_template3.py             2.9s (6)             2.9s (4)  +0.4%
            pyxl_bench.py             2.6s (6)             2.5s (4)  -1.5%
sqlalchemy_imperative2.py             3.0s (6)             2.9s (4)  -1.9%
                  geomean                 2.8s                 2.8s  -1.0%
```

### DIFF
--- a/src/analysis/type_analysis.cpp
+++ b/src/analysis/type_analysis.cpp
@@ -113,9 +113,6 @@ private:
         assert(old_type);
         assert(speculation != TypeAnalysis::NONE);
 
-        // TODO: reenable this
-        return old_type;
-
         if (speculated_cls != NULL && speculated_cls->is_constant) {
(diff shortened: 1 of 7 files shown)


undingen:
I'm not sure if removing the ```assert(!skip_next_pythonlike_frame);``` is correct.
But I couldn't come up with a reduced test to check if just removing the assert is correct.

kmod:
Can you try merging https://github.com/undingen/pyston/pull/2 ?  I think that should let us reenable that assert.  otherwise lgtm :)

undingen:
Thanks for your patch, this should now be ready but we may want to wait for #861 because travis ci encountered several random errors which this PR should fix.

PR #855, opened 2015-08-21
Reduce allocations

kmod:
Spent some time looking into it since GC time is showing up pretty high on the django benchmarks.

This PR reduces django_template3 bytes allocated by about 20%, leading to about a 3% improvement on that benchmark, and about 2% overall (geomean).


undingen:
The speedup it pretty amazing ;-)
I too forgot to add the destructor to the set class. Do you want to add it or should I create a PR?

kmod:
ok, I'll update the pr to fix set too :)

kmod:
Hmm spoke a bit too soon; after fixing some bugs the effects are about half what they were at first :/

PR #860, opened 2015-08-24
Callattr secondary slowpath

tjhance:
Been extremely busy lately, but I finally found time to attempt to finish this up. (I think. We'll see how the build goes.)

Adds the ability to give a "secondary slowpath" to the rewriter, after which point it will call that slowpath when the guard fails.

- First, it kills the concept of live_outs within the rewriter proper, and it always pushes them and pops them later.
- Rather than rearrange args before jmping when a guard fails, the guard jmp goes to a short section which rearranges the args and *then* jumps to the slowpath. Doing the guard rearrangement outside the main path was kind of tricky because the (very stateful) rewriter doesn't have any support for branching. I basically just dump the locations and then write a function which uses that info to rearrange the args. I imagine we'll get a better story for branching later, so I don't see this function being around long-term.

When I measured it, the above seemed to have very little perf effect.

Then for dealing with the callattr slowpath, I basically add an interface to the rewriter that adds a new function and new args, and then does the same guard flow as before (except it has to call the slowpath rather than jmping to the end).

On the whole this is pretty hacky :\


tjhance:
damn, I have to fix the merge conflicts to get travis-ci to run? :(

kmod:
You can set up travis-ci on your personal fork :)  but the cache will be cold so it will take a long time.

I'm still a bit worried about how applicable this particular feature ("set a new slowpath function") will be.  For example, in the callattr case (when we've gotten a descriptor we didn't special case), we're not going to know what our caller will want to do after calling callattrInternal -- sometimes the caller will be the top-level callattr(), but sometimes it will be something like binop().  I feel like it'd be nice to keep it more self-contained, ie saying that we are doing a new sub-rewrite.  That way the outer rewrite can see the sub-rewrite as a single unit, and the inner rewrite could potentially just operate independently.

I think in theory we could almost do this right now.  We could just create a sub-rewriter, tell it to write starting from the current offset, and that all the currently-used variables are live_outs.  Then the outer rewrite just sees the subrewrite as a (large) function call, and the inner rewrite doesn't really have to know what's going on.  And by adding nesting, we don't have to add any more branching support than the rewriter already has: the subrewrite just does guarding, and the outer rewrite doesn't need any control flow (it's all contained in the inner rewrite).

Anyway, sorry that's wordy.  I'm not opposed to the approach here of managing the slowpath; I guess I just want to make sure we will end up serving the current crop of use cases.

PR #861, opened 2015-08-24
Fix a few errors encountered while running test/cpython/test_math.py

undingen:
The first one (memory error) was the one I initially was planning to fix but then I run the test with ```-n``` and saw that there are additional problems. I'm somewhat surprised that no current test triggers them.

I'm not sure if ```AbstractFunctionType::callType``` should not in addition return ```UNKNOWN``` instead of ```UNDEF```. Because of this we will trigger an assert if we can't find a function that supports the requested args instead of just throwing an exception at runtime.
I don't like the ```AbstractFunctionType``` arg handling code I think it shouldn't copy so much, but I'm not currently feeling like rewriting it...

### DIFF
--- a/src/analysis/type_analysis.cpp
+++ b/src/analysis/type_analysis.cpp
@@ -225,7 +225,7 @@ private:
         return rtn;
     }
 
-    bool hasFixedBinops(CompilerType* type) {
+    bool hasFixedOps(CompilerType* type) {
         // This is non-exhaustive:
         return type == STR || type == INT || type == FLOAT || type == LIST || type == DICT;
     }
@@ -233,7 +233,7 @@ private:
     void* visit_augbinop(AST_AugBinOp* node) override {
         CompilerType* left = getType(node->left);
         CompilerType* right = getType(node->right);
-        if (!hasFixedBinops(left) || !hasFixedBinops(right))
+        if (!hasFixedOps(left) || !hasFixedOps(right))
             return UNKNOWN;
 
         // TODO this isn't the exact behavior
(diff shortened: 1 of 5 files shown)


undingen:
the virtualenv test fails on travis ci all the time but can't reproduce :-(

src/analysis/type_analysis.cpp
```diff
@@ -508,12 +508,20 @@ class BasicBlockTypePropagator : public ExprVisitor, public StmtVisitor, public
...
         CompilerType* operand = getType(node->operand);
+        if (!hasFixedOps(operand))
+            return UNKNOWN;
 
         // TODO this isn't the exact behavior
         BoxedString* name = getOpName(node->op_type);
         CompilerType* attr_type = operand->getattrType(name, true);
+
+        if (attr_type == UNDEF)
+            attr_type = UNKNOWN;
```
kmod: This feels odd to me -- wouldn't it be getattrType's fault more than this function's?

src/analysis/type_analysis.cpp
```diff
@@ -508,12 +508,20 @@ class BasicBlockTypePropagator : public ExprVisitor, public StmtVisitor, public
 
     void* visit_unaryop(AST_UnaryOp* node) override {
         CompilerType* operand = getType(node->operand);
+        if (!hasFixedOps(operand))
+            return UNKNOWN;
 
         // TODO this isn't the exact behavior
```
kmod: I think these comments can be removed now

src/analysis/type_analysis.cpp
kmod: (I tried removing it and haven't run into any issues yet)

kmod:
I ended up having to combine this with #862 -- running either PR by itself would fail due to the issue that the other one fixed, but somehow running without either fix worked fine...

PR #865, opened 2015-08-25
Fix 'error return without exception set' error when running django_template3

undingen:
I'm not sure if this is the correct fix, @kmod you can better judge if this is the correct fix.


kmod:
shoot sorry I've started to run into this too.  I'm not sure the fix is quite right -- I think there's some bit of code that is throwing an exception on the normal path but doesn't throw it on the rewritten path.  I'll take a look tomorrow, so for now you might want to just back out #864 locally.  sorry :/

kmod:
Ok now that I got some sleep I think this is right :P  I had previously done the reverse of this change which made me suspicious, but I think I was being misled by a separate bug (which was fixed).

I expanded on this a bit in #867 and will try to get that in.

PR #870, opened 2015-08-27
Small improvements to reduce the size of rewrites

undingen:
- If a jump offset is larger then 0x80 the instruction encoding requires 6bytes instead of 2bytes.
This adds up quickly, thats why we will try to find another jump to the slowpath with the same condition with a
smaller offset and jump to it / use it as a trampoline.
The benchmark show that this increases the performance slightly even though it introduces additional jumps.
- use xor instead of a mov to clear a register

```
ivy bridge
                                150982683ad1dcc208:  4f745cab9d0138b2c3:
           django_template3.py             2.8s (4)             2.8s (4)  -0.3%
                 pyxl_bench.py             2.3s (4)             2.3s (4)  +0.4%
     sqlalchemy_imperative2.py             2.9s (4)             2.8s (4)  -1.6%
                       geomean                 2.6s                 2.6s  -0.5%

broadwell
                                150982683ad1dcc208:  4f745cab9d0138b2c3:
           django_template3.py             2.5s (4)             2.5s (4)  -0.3%
                 pyxl_bench.py             1.5s (4)             1.4s (4)  -2.1%
     sqlalchemy_imperative2.py             2.1s (4)             2.1s (4)  -1.1%
                       geomean                 2.0s                 1.9s  -1.2%
```

### DIFF
--- a/src/asm_writing/assembler.cpp
+++ b/src/asm_writing/assembler.cpp
@@ -300,6 +300,17 @@ void Assembler::movslq(Indirect src, Register dest) {
     mov_generic(src, dest, MovType::SLQ);
 }
 
+void Assembler::clear_reg(Register reg) {
+    int reg_idx = reg.regnum;
+    // we don't need to generate a REX_W because 32bit instructions will clear the upper 32bits.
+    if (reg_idx >= 8) {
+        emitRex(REX_R | REX_B);
+        reg_idx -= 8;
+    }
+    emitByte(0x31);
+    emitModRM(0b11, reg_idx, reg_idx);
+}
+
 void Assembler::mov_generic(Indirect src, Register dest, MovType type) {
     int rex;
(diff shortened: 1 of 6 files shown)


undingen:
Looks like this is stuff where the cpu arch makes a big difference.
Here are the results of a benchmark run on a broadwell cpu instead of the older ivy bridge:
```
                                150982683ad1dcc208:  4f745cab9d0138b2c3:
           django_template3.py             3.1s (4)             3.0s (4)  -2.6%
                 pyxl_bench.py             2.3s (4)             2.1s (4)  -4.9%
     sqlalchemy_imperative2.py             2.6s (4)             2.7s (4)  +0.8%
                       geomean                 2.6s                 2.6s  -2.3%
```

kmod:
nice :)

PR #872, opened 2015-08-27
Rewriter "aggressiveness" and backoff

kmod:
ie rewrite less aggressively over time.  instead of trying to rewrite into every single function, at some point give up and emit a call to the function.  similar to reducing the inlining/tracing depth, but a bit more hard-coded for the moment.

This helps with megamorphic sites (we will emit a call to the most-specific handler that will end up catching all the cases) and with some cases we previously couldn't rewrite (will inline a bit).

```
           django_template3.py             2.8s (2)             2.8s (2)  +0.5%
                 pyxl_bench.py             2.3s (2)             2.2s (2)  -2.5%
     sqlalchemy_imperative2.py             2.8s (2)             2.8s (2)  -0.5%
                       geomean                 2.6s                 2.6s  -0.8%
```


undingen:
Oh this rewriter aggressiveness idea is very good :-)

PR #875, opened 2015-08-28
Move closer towards exposing the public gc interface only in one file.

rudi-c:
Remove a lot of `#include "gc/*"` and move more functions that the GC can expose to `gc.h`.


kmod:
Cool, nice cleanup :)

PR #881, opened 2015-08-31
when calling a BoxedWrapperDescriptor don't create a BoxedWrapperObject

undingen:
I want to get ride of the thousands of BoxedWrapperObject we create during calls.
Do do you guys think this is the right way?
If so I will clean it up / reduce the duplication between ```BoxedWrapperDescriptor::tppCall``` and ```BoxedWrapperObject::tppCall```

### DIFF
--- a/src/codegen/compvars.cpp
+++ b/src/codegen/compvars.cpp
@@ -1878,8 +1878,10 @@ public:
     }
 
     ConcreteCompilerVariable* nonzero(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var) override {
-        static BoxedString* attr = internStringImmortal("__nonzero__");
+        if (cls == None->cls)
+            return makeBool(false);
 
+        static BoxedString* attr = internStringImmortal("__nonzero__");
         bool no_attribute = false;
         ConcreteCompilerVariable* called_constant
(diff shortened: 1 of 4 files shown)


kmod:
sgtm :)  for some reason I thought this wouldn't work since the BoxedWrapperDescriptor object shouldn't be callable, but I guess it is (and is supposed to be).

Maybe we could do something like this though to reduce the code duplication:
```C++
BoxedWrapperObject::tppCall(BoxedWrapperObject* self, ...) {
    bindObjIntoArgs(self->obj, ...);
    return BoxedDescriptorObject::tppCall(self->wrapper, ...);
}
```

undingen:
I removed the code duplication as you suggested and sneaked in a optimization for ```nonzero``` because I noticed that the ```getattr``` it was using produced a lot of ```BoxedWrapperObject``` and taught compvar that ```None``` is never nonzero (unrelated to this change but while stepping though the code I noticed that we generated calls to ```noneNonzero```).
```
                                0a8385c93c2fd1a90a:  65e3ce5be40f6172b8:
           django_template3.py             2.8s (4)             2.8s (4)  -0.3%
                 pyxl_bench.py             2.2s (4)             2.2s (4)  +1.0%
     sqlalchemy_imperative2.py             2.7s (4)             2.7s (4)  -1.1%
                       geomean                 2.5s                 2.5s  -0.2%
```
I don't know yet why pyxl appears to be slower.

kmod:
nice :)

PR #888, opened 2015-09-02
add tp_nextiter implementation for our iterators

undingen:
```
                                7eec7fa2e53c908a06:  5c60ed7c85337bb516:
           django_template3.py             2.7s (4)             2.7s (4)  -0.6%
                 pyxl_bench.py             2.2s (4)             2.2s (4)  -1.3%
     sqlalchemy_imperative2.py             2.6s (4)             2.6s (4)  -1.4%
                       geomean                 2.5s                 2.5s  -1.1%
```


undingen:
I'm not sure if we should not remove our ```__hasnext__``` implementation because I think it doesn't give us as any advantage over ```tp_iternext``` which can just return ```NULL``` when it encountered the end!??

kmod:
nice :)

PR #891, opened 2015-09-03
Add missing tp_iter and tp_iternext functions

undingen:


kmod:
lgtm (minus the test failure :P)

undingen:
I changed the ```src/runtime/iterobject.cpp``` file to fix the bug and while there I made theme more similar, please have a look at the changes to that file.

undingen:
I added the missing AttrWrapper slot overwrites and rebased

PR #893, opened 2015-09-03
Some improvments that let test_index could pass

Daetalus:
**Changes**:

- Add `__imul__` to `list` object, and let `list` could multiple with the instance which has `__index__` attribute.
- Let `str` could multiple with `long`.

Full list improvements is on the way. @kmod I saw your comment in current `listSort` function, should we use CPython's list sort with some Pyston changes, or improve the exist Pyston implementation? 

I have no preference. Just want to know copy lots of CPython code whether is suitable here.(the code of CPython list sort has many lines. Not only the `list_sort`, but also dependent functions.)

### DIFF
--- a/src/capi/abstract.cpp
+++ b/src/capi/abstract.cpp
@@ -2071,7 +2071,7 @@ extern "C" PyObject* PyNumber_Long(PyObject* o) noexcept {
         return res;
     }
     if (PyLong_Check(o)) { /* A long subclass without nb_long */
-        return o;
+        return _PyLong_Copy((PyLongObject*)o);
     }
     trunc_func = PyObject_GetAttr(o, trunc_name);
     if (trunc_func) {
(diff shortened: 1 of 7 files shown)


src/capi/abstract.cpp
```diff
@@ -2071,8 +2071,7 @@ extern "C" PyObject* PyNumber_Long(PyObject* o) noexcept {
         return res;
     }
     if (PyLong_Check(o)) { /* A long subclass without nb_long */
-        BoxedInt* lo = (BoxedInt*)o;
-        return PyLong_FromLong(lo->n);
+        return o;
```
kmod: CPython does `return _PyLong_Copy(o);`.  I can't find any good way to trigger this case (I think it only happens via an extension class?) but we might as well do it that way?

src/runtime/str.cpp
```diff
@@ -1146,9 +1146,24 @@ extern "C" Box* strMul(BoxedString* lhs, Box* rhs) {
     assert(PyString_Check(lhs));
 
     int n;
+
+    if (PyIndex_Check(rhs)) {
+        Box* index = PyNumber_Index(rhs);
+        if (PyErr_Occurred()) {
```
kmod: minor, but PyErr_Occurred is slower than you'd expect and it's good to check the return value instead.  (Doesn't seem like it really matters here but it's good practice.)

kmod:
awesome :)

PR #896, opened 2015-09-04
add sys.long_info

aisk:
Implement sys.long_info.

Since we could use CPython's long object implementation, so just put two const numbers in sys.long_info for `make run_test_long`(This two numbers are picked from OSX 64 bit CPython's sys.long_info).`

### DIFF
--- a/src/runtime/builtin_modules/sys.cpp
+++ b/src/runtime/builtin_modules/sys.cpp
@@ -17,6 +17,7 @@
 #include <cmath>
 #include <langinfo.h>
 #include <sstream>
+#include <gmp.h>
 
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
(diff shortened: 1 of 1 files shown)


src/runtime/builtin_modules/sys.cpp
```diff
@@ -17,6 +17,7 @@
 #include <cmath>
 #include <langinfo.h>
 #include <sstream>
+#include <gmp.h>
```
Daetalus: Do we really need `gmp.h` here?
BTW, can we just hardcode the `sys.long_info` in `test_long`? We can implement is latter. Maybe we will use different approach to implement `long` in future.

aisk: Hardcode the `sys.long_info` is OK, but I think in `sys` more test cases will be passed if it depends on sys.long_info (maybe there is no more test cases depends on it LOL, so I don't know which choice is suitable).

kmod:
Yeah, it's tricky to decide what we should do with this.  It's hard to imagine what kind of Python code would rely on this metadata (as opposed to something like sys.maxint which is a bit more reasonable) -- but if there is any code out there that actually relies on the values in this field, filling it with arbitrary numbers seems like it would break the code just as much as it not existing.  So I'm in favor of not adding sys.long_info to say "hey we don't have that metadata in any meaningful way" rather than making something up for it.

I think the cpython test only uses it in a pretty superficial way, so we could probably just replace the single usage of `sys.long_info.bits_per_digit` with 64.  If we did want to support sys.long_info, I think the correct values to use are to set `sizeof_digit` to `sizeof(mp_limb_t)` and `bits_per_digit` to `8 * sizeof(mp_limb_t)` (I think these are correct but I'm not really sure).  But since we're not really using a "digits-based" representation (we just have a single mpz_t black-box object) I'd prefer to just not add sys.long_info for now.

aisk:
@kmod Should I just add this two number in `cpython/long_test.py`?

kmod:
Sorry, yeah I would recommend replacing the existing `SHIFT = sys.long_info.bits_per_digit` with `SHIFT = 64` (along with a "Pyston change" comment) in the test file.

PR #907, opened 2015-09-09
Add "pyinotify" test, with a minor modification to Pyston

### LINKED ISSUE #903: Get pyinotify working
(not a builtin module)

Needs:
- [x] termios module (#898)
- [ ] PySequence_SetItem
(maybe others)

Daetalus:
**Changes**

- Add `pyinotify` submodule.

- Add `pyinotify` test.

closes #903

### DIFF
--- a/test/tests/ctypes_test.py
+++ b/test/tests/ctypes_test.py
@@ -1,0 +1,7 @@
+from ctypes import *
+s = "tmp"
+ap = create_string_buffer(s)
+
+print type(ap)
+print type(c_void_p.from_param(ap))
+print type(cast(ap, c_char_p))
--- a/test/tests/ctypes_test.py
+++ b/test/tests/ctypes_test.py
@@ -1,3 +1,4 @@
+# skip-if: True
 from ctypes import *
 s = "tmp"
 ap = create_string_buffer(s)
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,3 +37,6 @@
 [submodule "test/lib/decorator"]
 	path = test/lib/decorator
 	url = https://github.com/micheles/decorator
+[submodule "test/lib/pyinotify"]
+	path = test/lib/pyinotify
+	url = https://github.com/seb-m/pyinotify.git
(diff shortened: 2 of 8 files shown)


kmod:
It looks like the test failures could be fixed by changing the `git@gituhb.com` url to `https://github.com/`.  We can gate the test on `if "linux" in sys.platform`, but I think our `sys.platform` is currently hardcoded :/

Will take a look into the ctypes thing.  One thing that can help with this kind of thing is using our debug-CPython helpers, such as `make pydbg_this_ctypes_test BR=PyCSimpleType_from_param`.  It looks like it doesn't quite work here since the filename is hardcoded to be `/build/buildd/python2.7-2.7.6/Modules/_ctypes/_ctyes.c`.

kmod:
Ok, I sent you https://github.com/Daetalus/pyston/pull/3 which I think should workaround the ctypes issue for now.  Could you see if we still need the pyinotify patch with the pr?  also it looks like there's some rebasing work to do, sorry :/  other than that this looks pretty good :)

Daetalus:
Skip the `ctype_test`, just like #937

test/tests/ctypes_test.py
```diff
@@ -0,0 +1,8 @@
+# skip-if: True
```
kmod: Is this test working now?  If so let's remove the 'skip-if'

Daetalus: I don't think it will working now. I saw @undingen did some fixs in a certain commit, in that commit, it could pass, but seems it not included in the current #937 . @undingen , would you mind to take a look, please?

Daetalus:
updated, ready for review.

test/tests/ctypes_test.py
kmod: I guess my theory was that that other bug might not affect this small test -- for instance there's no multithreading, and it's a short enough test that a GC won't happen.  Would you mind humoring me and trying enabling it? :)  I think it's unlikely that it will fail, but if it does maybe it will give some good information on what's going on.

kmod:
awesome, thanks for the change :)

PR #916, opened 2015-09-13
make more module test pass

aisk:


undingen:
Looks good :-), but could you please squash the commits into a single commit.
We normally only split stuff into different commits when the changes are isolated from each other or to highlight some particular important change from a larger change.
Thanks for the patch!

undingen:
:+1:

PR #921, opened 2015-09-15
Some stuff to enable "test_set"

### LINKED ISSUE #914: Pass CPython's "test_set"
ie `make run_test_set`.  Looks like there are quite a few failures there, which is concerning.

Daetalus:
**Ready to review**

**Changes:**
- Some improvements in `dict.cpp`, do not recompute the hash value in `fromKeys` if the container argument is set.
- Check the set is modified in iteration.
- Let the subclass of `set` and `frozenset` could take keyword argument.
- In some set functions, if the container argument is `set` or `dict, do not recompute the element hash value again.
- ~~Implement `__reduce__` to let `pickle` and `deepcopy` could work.~~ The new added commit of this file seems add this implementation. Remove it.
- Let `discard`, `remove` could handle the key which type is `frozenset`.
- Some other small bug fixing, such as clear the value in `init`, do not create new BoxedSet in `union` and other functions directly(need to apply `cls`), etc.

**Note**
- Disable two gc related test in `test_set`.

closes #914

### DIFF
--- a/src/runtime/dict.cpp
+++ b/src/runtime/dict.cpp
@@ -22,6 +22,7 @@
 #include "runtime/ics.h"
 #include "runtime/inline/list.h"
 #include "runtime/objmodel.h"
+#include "runtime/set.h"
 #include "runtime/types.h"
 #include "runtime/util.h"
 
@@ -510,11 +511,12 @@ Box* dictSetdefault(BoxedDict* self, Box* k, Box* v) {
         raiseExcHelper(TypeError, "descriptor 'setdefault' requires a 'dict' object but received a '%s'",
                        getTypeName(self));
 
-    auto it = self->d.find(k);
+    BoxAndHash k_hash(k);
+    auto it = self->d.find(k_hash);
     if (it != self->d.end())
         return it->second;
 
-    self->d.insert(std::make_pair(k, v));
+    self->d.insert(std::make_pair(k_hash, v));
     return v;
 }
 
@@ -559,8 +561,14 @@ Box* dictNonzero(BoxedDict* self) {
 
 Box* dictFromkeys(Box* cls, Box* iterable, Box* default_value) {
     auto rtn = new BoxedDict();
-    for (Box* e : iterable->pyElements()) {
-        dictSetitem(rtn, e, default_value);
+    if (PyAnySet_Check(iterable)) {
+        for (auto&& elt : ((BoxedSet*)iterable)->s) {
+            rtn->d.insert(std::make_pair(elt, default_value));
+        }
+    } else {
+        for (Box* e : iterable->pyElements()) {
+            dictSetitem(rtn, e, default_value);
+        }
     }
 
     return rtn;
--- a/src/runtime/set.cpp
+++ b/src/runtime/set.cpp
@@ -39,8 +39,9 @@ class BoxedSetIterator : public Box {
 public:
     BoxedSet* s;
     decltype(BoxedSet::s)::iterator it;
+    long size;
 
-    BoxedSetIterator(BoxedSet* s) : s(s), it(s->s.begin()) {}
+    BoxedSetIterator(BoxedSet* s) : s(s), it(s->s.begin()), size(s->s.size()) {}
 
     DEFAULT_CLASS(set_iterator_cls);
 
@@ -68,7 +69,15 @@ Box* setiteratorHasnext(BoxedSetIterator* self) {
 
 Box* setiteratorNext(BoxedSetIterator* self) {
     RELEASE_ASSERT(self->cls == set_iterator_cls, "");
+    if (self->s->s.size() != self->size) {
+        raiseExcHelper(RuntimeError, "Set changed size during iteration");
+    }
     return self->next();
+}
+
+Box* setiteratorLength(BoxedSetIterator* self) {
+    RELEASE_ASSERT(self->cls == set_iterator_cls, "");
+    return boxInt(self->s->s.size());
 }
 
 Box* setiter_next(Box* _self) noexcept {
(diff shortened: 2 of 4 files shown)


src/runtime/set.cpp
```diff
@@ -114,7 +133,11 @@ Box* frozensetNew(Box* _cls, Box* container) {
 
     // Some optimizations from CPython: frozensets can be shared:
     if (_cls == frozenset_cls) {
-        if (!container) {
+        bool isEmpty = false;
+        if (container && PyObject_HasAttr(container, internStringMortal("__len__"))) {
```
kmod: Minor, but it's nice to do something like `static BoxedString* len_str = internStringImmortal("__len__"); HasAttr(Container, len_str);` to avoid interning the string each time.

But also, I'm a bit worried about this optimization since CPython doesn't do it -- I think there's nothing stopping a user from defining a class where `__len__` returns 0 but iterating over it returns more-than-zero elements, and in that case we'll have different behavior than CPython.  Is there something where this helped a lot?

kmod:
Looks like you didn't actually enable test_set.py :P  Also, minor, but if you want to remerge+retest your PR, what I normally do is `git commit --amend` to update the timestamp, and then `git push -f` (this prevents the "opened PR"  / "reopened PR" emails).

src/runtime/set.cpp
```diff
@@ -125,26 +148,61 @@ Box* frozensetNew(Box* _cls, Box* container) {
...
 
     // Note: set.__new__ explicitly ignores the container argument.
     return makeNewSet(cls, NULL);
 }
 
-Box* setInit(Box* _self, Box* container) {
+Box* setInit(Box* _self, BoxedTuple* args, BoxedDict* kwargs) {
     RELEASE_ASSERT(PySet_Check(_self), "");
 
+    if (_self->cls == set_cls && kwargs) {
```
kmod: tiny, but I think there should be a `&& PyDict_Size(kwargs) == 0)` check in here too.  I think when called from python code, we will automatically try to pass in a NULL kwargs, but I think C API code is free to pass in a non-null-but-empty kwargs object.

kmod: Also, it looks like CPython now uses this check:
```C
    if (PySet_Check(self) && !_PyArg_NoKeywords("set()", kwds))
        return -1;
```

ie the behavior doesn't seem to match what they talked about in that bug report anymore.  This code (which we should test I guess) works for us but throws an exception in CPython:
```python
class MySet(set):
  def __new__(cls, *args, **kwargs):
    return set.__new__(cls, *args)

MySet(a=1)
```

src/runtime/set.cpp
```diff
@@ -125,26 +148,61 @@ Box* frozensetNew(Box* _cls, Box* container) {
...
+        raiseExcHelper(TypeError, "set() does not take keyword arguments");
+    }
+
+    int args_sz = args->size();
```
kmod: Could we keep this as a single optional positional argument instead of taking varargs?

kmod:
Cool, I like these changes :)  It's too bad that there's no good systematic way to make sure we don't recompute the hashes if we don't need to.  Maybe one option is we could make BoxAndHash not be implicitly constructable from a Box, which means we would have to explicitly hash it any time we wanted to do dict/set operations on it.  Not sure if that's worth it though.

src/runtime/set.cpp
```diff
@@ -600,6 +652,87 @@ Box* setGt(BoxedSet* self, BoxedSet* rhs) {
     return setIssuperset(self, rhs);
 }
 
+Box* setContains(BoxedSet* self, Box* v) {
+    RELEASE_ASSERT(PyAnySet_Check(self), "");
+
+    if (PySet_Check(v)) {
```
kmod: Hmm I'm not sure I understand the logic of this block.  It also seems to give a different result from CPython for:
```python
print set() in frozenset([frozenset()])
```

src/runtime/set.cpp
Daetalus: @kmod Would you mind to watch the line 598 in new `test_set.py`, there have:
```
frozenset(), frozenset([]), frozenset(()), frozenset(''),
               frozenset(), frozenset([]), frozenset(()), frozenset(''),
               frozenset(xrange(0)), frozenset(frozenset())
```
Need return one empty `frozenset`. This argument could not detect with `None`. I also tried `__nonzero__`, but it is not satisfied `xrange(0)`. From my knowledge, only `__len__` could meet this condition. Maybe there has a better way?

Daetalus:
Please see the line note about the `__len__` issue.

Other issues were addressed.

There has one change brought by commit https://github.com/dropbox/pyston/commit/0aa5dea851ba6cd3847496feac3fa87d25be50a3  will let the `test_set` exist with code 139. Please see the change in `heap.cpp` in this PR for more information. I don't know this is whether a correct fixing.

src/runtime/set.cpp
```diff
@@ -100,21 +109,39 @@ BoxedSet* makeNewSet(BoxedClass* cls, Box* container) {
...
-        if (!container) {
+        if (!_PyArg_NoKeywords("frozenset()", kwargs)) {
+            throwCAPIException();
+        }
+        bool isEmpty = false;
+        static BoxedString* len_str = internStringImmortal("__len__");
+        if (container && PyObject_HasAttr(container, len_str)) {
+            isEmpty = PyObject_Size(container) > 0 ? false : true;
+        }
+        if (!container || isEmpty) {
```
kmod: It looks like cpython's logic (from `frozenset_new`) is to create a frozenset object from the container, but if the resulting frozenset is empty, to discard it and return the emptyfrozenset singleton instead.

src/runtime/set.cpp
kmod: Sorry to be a stickler, but can we just use CPython's logic here?  Your code seems very reasonable, but since it's different there are going to be weird corner cases where we will do something different from cpython (ex: subclassing from set and adding a `__hash__` function that throws an exception.  or the fact that CPython will create a temporary frozenset object as the standin.)

kmod:
looks pretty good :)  I think #934 should be a better fix for the segfault you're seeing; will try to get that in soon.

kmod:
sorry lost track of the state of this PR -- I merged #934 so we should be able to remove your temporary fix in this branch.

Daetalus:
Hi @kmod ,  I mimic the CPython implementation in `setContains` and other similar function, but Pyston and CPython still have some differences, please give review the latest changes.

src/runtime/set.cpp
```diff
@@ -600,20 +642,121 @@ Box* setGt(BoxedSet* self, BoxedSet* rhs) {
...
+Box* setContains(BoxedSet* self, Box* key) {
+    RELEASE_ASSERT(PyAnySet_Check(self), "");
+
+    if (PySet_Check(key)) {
+        try {
+            BoxAndHash k_hash(key);
+            return boxBool(self->s.find(k_hash) != self->s.end());
+        } catch (ExcInfo e) {
+            if (e.matches(TypeError))
+                PyErr_Clear();
```
kmod: we don't need to call PyErr_Clear here, since we won't have set the exception as the CAPI exception (you have to do that manually).  maybe we should though... but anyway for now this could just be `if (!e.matches(TypeError)) throw e;`

src/runtime/dict.cpp
```diff
@@ -559,8 +561,18 @@ Box* dictNonzero(BoxedDict* self) {
 
 Box* dictFromkeys(Box* cls, Box* iterable, Box* default_value) {
     auto rtn = new BoxedDict();
-    for (Box* e : iterable->pyElements()) {
-        dictSetitem(rtn, e, default_value);
+    if (PyAnySet_Check(iterable)) {
+        for (auto&& elt : ((BoxedSet*)iterable)->s) {
+            auto it = rtn->d.find(elt);
+            if (it != rtn->d.end())
+                it->second = default_value;
```
kmod: This looks a bit odd -- I'm not sure that updating the iterator is guaranteed to mutate the underlying container.  I tried to test it but 1) it looks like we don't have any test coverage of this line of code :)  and 2) if the element already exists, shouldn't it already be set to default_value?

Maybe we could skip the find() step and just do an insert().

src/runtime/set.cpp
```diff
@@ -600,20 +642,121 @@ Box* setGt(BoxedSet* self, BoxedSet* rhs) {
...
+    return boxBool(self->s.find(key) != self->s.end());
+}
+
+Box* setRemove(BoxedSet* self, Box* key) {
+    RELEASE_ASSERT(isSubclass(self->cls, set_cls), "");
+
+    if (PySet_Check(key)) {
+        try {
+            BoxAndHash k_hash(key);
+            if (self->s.find(k_hash) != self->s.end()) {
```
kmod: FYI `erase()` returns a bool, which will be true if the element existed or not.  It's probably not a big deal and I bet the compiler can figure a lot of this out, but I think it's slightly better form to do `bool existed = self->s.erase(k_hash); if (existed) return None;`

kmod:
sorry, just a couple last comments.  looks pretty good though :)

I think we should start being pretty strict on ourselves about diverging from CPython at all.  I tried to check that all of your porting was right, but it's way easier to feel comfortable about it when we're using the CPython code character-for-character.  I think we're at the point that (at least after this PR) we need to have a pretty convincing reason in order to not just copy the CPython implementation.

kmod:
I know it's a bit hard for things like set which we started implementing before we had any cpython compatibility, so there's a lot of code in there which doesn't really correspond to the cpython code.  I think I prefer though that we start importing the CPython code even if it means we have duplicated functions with our existing code, and then over time we will migrate the old code.  It's a bit messier in the short term, but I think using-CPython-code is a big enough advantage that it's worth it.

Daetalus:
Sorry for trouble you. I will use CPython implementation when as much as possible.

src/runtime/set.cpp
```diff
@@ -100,51 +109,90 @@ BoxedSet* makeNewSet(BoxedClass* cls, Box* container) {
...
     RELEASE_ASSERT(isSubclass(cls, frozenset_cls), "");
 
     // Some optimizations from CPython: frozensets can be shared:
     if (_cls == frozenset_cls) {
-        if (!container) {
+        if (!_PyArg_NoKeywords("frozenset()", kwargs)) {
+            throwCAPIException();
+        }
+
+        BoxedSet* result = makeNewSet(cls, container);
```
kmod: Can we use the CPython logic for this section?  They have this:
```C
    if (iterable != NULL) {
        /* frozenset(f) is idempotent */
        if (PyFrozenSet_CheckExact(iterable)) {
            Py_INCREF(iterable);
            return iterable;
        }
        result = make_new_set(type, iterable);
        if (result == NULL || PySet_GET_SIZE(result))
            return result;
        Py_DECREF(result);
    }
    /* The empty frozenset is a singleton */
    if (emptyfrozenset == NULL)
        emptyfrozenset = make_new_set(type, NULL);
    Py_XINCREF(emptyfrozenset);
    return emptyfrozenset;
```

It's pretty similar, but the different ordering means that they can skip allocating a frozenset instance if they can.  (the code here will call makeNewSet regardless.)

Daetalus:
Ready to review. @kmod . The test failed randomly...

kmod:
Nice, thanks for following up on this :)

PR #924, opened 2015-09-17
Gc fix

kmod:
For the n'th time, fix the freeing-class-and-instance-in-the-same-collection issue.

This is a resurrection of #646, which had gotten reverted in preference of another approach which didn't quite work.  I think (crosses fingers) this one is working even though it's not very elegant.


kmod:
I had to fix a second gc issue to get this branch to pass: we would guard on a pointer value, and the object at that location would get freed and a new one allocated and that new object would be passed into the IC, which would incorrectly pass the guards.  yes, this happens.

The free-classes-after-instances fix seems to have a small but consistent impact on perf (+0.5%) but seems to disappear in pgo mode, so I'm just going to go ahead with it.

PR #925, opened 2015-09-17
Misc small fixes for lxml

undingen:


src/runtime/util.cpp
```diff
@@ -104,7 +104,7 @@ Box* noneIfNull(Box* b) {
     }
 }
 
-Box* coerceUnicodeToStr(Box* unicode) {
+template <ExceptionStyle S> Box* coerceUnicodeToStr(Box* unicode) noexcept(S == CAPI) {
```
kmod: I think the raiseExcHelper call in here needs to be changed.

kmod:
lgtm, though it looks like there's a small merge conflict.

src/runtime/util.cpp
undingen: good catch!

undingen:
I merged this in with a fix for the ```raiseExcHelper``` you mentioned + I added a commit which adds the ```addGCReference``` call for the builtin ```hasattr``` like you have done for ```getattr```

PR #927, opened 2015-09-20
Makefile to be able to do make after do make clean

corona10:
after make clean
it can not re-make beacuse makefile remove 
```
./from_cpython/Lib/test/pystone.py
```
so i fixed it


kmod:
Looks good :)  can you squash the commits together?

PR #928, opened 2015-09-20
some improvements in intNew

Daetalus:
After some investigation of the new added CPython tests. Some of them related to the long constructer. 

eg:
- Need to handle unicode string in `long()`; 
- Get correct error type and error message in long constructor. 
- etc...

When I fix it. I found our int constructor has same issue. So I make some improvements. The `test_long` need some discuss. So I create this PR first.

Changes:

This patch reorganize the error check in `_intNew`, and remove some unnessary or dead code.


kmod:
cool :)

PR #929, opened 2015-09-21
Add setting func.func_code, PyCFunction_Call, vars() error handling

undingen:
- in additon allow Py_TPFLAGS_HAVE_VERSION_TAG because even though we don't use it this flag shouldn't cause problems
- allow tp_del because AFAIK it's implemented.
- allow calling methods which specify METH_COEXIST because it doen't make a difference for the call
- disable PyCode_New call inside cython because we don't support it yet and it's unused.
- change cython to use PyCFunction_Call instead of the custom version it has for pypy


src/runtime/types.cpp
```diff
@@ -1675,6 +1676,18 @@ static Box* functionCode(Box* self, void*) {
...
+static void functionSetCode(Box* self, Box* v, void*) {
+    assert(self->cls == function_cls);
+
+    if (v == NULL || !PyCode_Check(v))
+        raiseExcHelper(TypeError, "__code__ must be set to a code object");
+
+    BoxedFunction* func = static_cast<BoxedFunction*>(self);
+    BoxedCode* code = static_cast<BoxedCode*>(v);
+    func->f = code->f;
+    func->dependent_ics.invalidateAll();
```
kmod: We haven't been super vigilant about using dependent_ics.  I took a quick look and didn't see much that could go wrong, but could we have an assert here that `f->internal_callable == NULL`?  callFunc (the default if no internal_callable is set) will use dependent_ics properly but I'm not sure about the different internal_callable implementations.  Also, if you did try to set the code object on a function with an internal_callable set, we'd have to null it out so that we didn't continue to call it.

I think the same kind of thing can apply to tpp_call, if you tried to do something like `MethodType.__call__.func_code = foo`.  CPython won't even allow that, but I think we might, but then it wouldn't have any effect.  Maybe one way to solve this to use something similar to the function `can_change_defaults` flag which will opt-in a function to this extra dynamicism, and then only set it for runtime-created functions.  We could potentially even just reuse (and rename) that same flag for this.

test/integration/Cython_0001-Pyston-change-we-don-t-support-custom-traceback-entr.patch
```diff
@@ -121,5 +138,30 @@ index 0310570..bcd0eb2 100644
...
+@@ -49,13 +49,16 @@
+ 
+ #if PY_MAJOR_VERSION < 3
+   #define __Pyx_BUILTIN_MODULE_NAME "__builtin__"
+-  #define __Pyx_PyCode_New(a, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos) \
+-          PyCode_New(a+k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos)
++  // Pyston change: we don't support this yet
++  // #define __Pyx_PyCode_New(a, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos) \
++  //         PyCode_New(a+k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos)
++  #define __Pyx_PyCode_New(a, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos) None
```
kmod: This feels risky to have this silently just return None -- what was using it?  Could we modify the caller(s) instead?

undingen:
I addressed the issues pointed out:
- ```functionSetCode```: assert on internal_callable, only allow python functions
- don't overwrite ```__Pyx_PyCode_New``` with None instead remove all calls
- we don't support types which have a ```tp_del``` and a ```tp_dealloc```:
 patch cythons generator to only use ```tp_dealloc```

please have a look at the last commit

undingen:
@kmod I added a lxml test and added a derived version of your ```PyGILState_Ensure, PyGILState_Release``` implementation

test/integration/Cython_0001-Pyston-change-we-don-t-support-custom-traceback-entr.patch
```diff
@@ -1,14 +1,42 @@
...
+                    binding=binding or node.specialized_cpdefs,
+                    specialized_cpdefs=node.specialized_cpdefs,
+                    code_object=CodeObjectNode(node))
++        """
++        return cls(node.pos,
++                   def_node=node,
++                   pymethdef_cname=node.entry.pymethdef_cname,
++                   binding=binding or node.specialized_cpdefs,
++                   specialized_cpdefs=node.specialized_cpdefs,
++                   code_object=None)
```
kmod: hmm what's the functionality difference for using None here?  (I'm always worried about these slight changes that might end up biting us later in subtle ways.)

src/runtime/types.cpp
```diff
@@ -1680,6 +1681,25 @@ static Box* functionCode(Box* self, void*) {
...
+    assert(self->cls == function_cls);
+
+    if (v == NULL || !PyCode_Check(v))
+        raiseExcHelper(TypeError, "__code__ must be set to a code object");
+
+    BoxedFunction* func = static_cast<BoxedFunction*>(self);
+    BoxedCode* code = static_cast<BoxedCode*>(v);
+
+    if (!func->f->source || !code->f->source)
+        raiseExcHelper(TypeError, "__code__ can only be set on python functions");
```
kmod: Personally I'd make this an assertion (crash) rather than an exception.  probably not that important in this case since code shouldn't be hitting this, but exceptions can get swallowed in a lot of ways :/

kmod:
lgtm with the two comments

src/runtime/types.cpp
undingen: good idea, I changed it to a ```RELEASE_ASSERT```

test/integration/Cython_0001-Pyston-change-we-don-t-support-custom-traceback-entr.patch
undingen: This let's cython not generate ```PyCode_New``` calls. My previous workaround was to replace all callsites inside cython with ```None``` but as you pointed out that's ugly. That's now my slightly cleaner workaround. The problem is that supporting the ```PyCode_New``` function is hard because it tries to create a code object for all python functions which cython translates to c. I think the reason is that it need's it in order to generate backtraces. But we have them currently already disabled inside cython so it should not make a difference for us currently.

undingen:
I merged this in, if you have a better idea for the cython workaround I am happy to fix it in a follow up PR :-). But I suspect that I have to change cython in #937 so it would fit right in there.

PR #932, opened 2015-09-22
fix Makefile to be able to 'make' after 'make clean'

corona10:
@kmod 
i try to squash previous PR but it was messed up 
so i closed it and reopen it


kmod:
cool, thanks :)

PR #933, opened 2015-09-22
some long improvements

Daetalus:
The long constructor is far from complete. It block too many CPython tests. Just fix it along with other issues found in `test_long`. test_cpickle also blocked by it.

All these code are tested and passed in `test_long`. But not enabled `test_long` here. Because `mpz_get_d` use "round to zero", CPython need "round to nearest".

### DIFF
--- a/src/runtime/long.cpp
+++ b/src/runtime/long.cpp
@@ -404,6 +404,15 @@ extern "C" PyAPI_FUNC(PyObject*) _PyLong_Format(PyObject* aa, int base, int addL
 }
 
 extern "C" PyObject* PyLong_FromDouble(double v) noexcept {
+    if (isnan(v)) {
+        PyErr_SetString(PyExc_ValueError, "cannot convert float NaN to integer");
+        return NULL;
+    }
+    if (isinf(v)) {
+        PyErr_SetString(PyExc_OverflowError, "cannot convert float infinity to integer");
+        return NULL;
+    }
+
     BoxedLong* rtn = new BoxedLong();
     mpz_init_set_d(rtn->n, v);
     return rtn;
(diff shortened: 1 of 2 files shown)


src/runtime/long.cpp
```diff
@@ -640,68 +649,128 @@ extern "C" PyObject* PyLong_FromUnsignedLongLong(unsigned long long ival) noexce
...
+            if (!PyInt_Check(_base)) {
+                PyErr_Format(PyExc_TypeError, "integer argument expected, got %s", getTypeName(_base));
+                return NULL;
+            }
 
-        if (!PyString_Check(val))
-            raiseExcHelper(TypeError, "long() can't convert non-string with explicit base");
-        BoxedString* s = static_cast<BoxedString*>(val);
+            if (val == None) {
+                PyErr_SetString(PyExc_TypeError, "int() missing string argument");
```
kmod: "int()" -> "long()"

src/runtime/long.cpp
```diff
@@ -640,68 +649,128 @@ extern "C" PyObject* PyLong_FromUnsignedLongLong(unsigned long long ival) noexce
...
+        Box* r = PyLong_FromUnicode(PyUnicode_AS_UNICODE(val), PyUnicode_GET_SIZE(val), base);
+        if (!r) {
+            if (S == CAPI)
+                return NULL;
+            else
+                throwCAPIException();
+        }
+        return r;
+    } else {
+        static BoxedString* long_str = internStringImmortal("__long__");
```
kmod: Can this branch actually get taken?

src/runtime/long.cpp
```diff
@@ -640,68 +649,128 @@ extern "C" PyObject* PyLong_FromUnsignedLongLong(unsigned long long ival) noexce
...
+        if (!r) {
+            if (S == CAPI) {
+                return NULL;
+            } else
+                throwCAPIException();
+        }
+        return r;
+    }
+
+    if (val->cls == str_cls) {
```
kmod: I think this should be PyString_Check.

kmod:
looks pretty good :)

kmod:
nice :)

PR #936, opened 2015-09-23
Enable a bunch more sqlalchemy tests

kmod:


kmod:
The last "fix" (disabling unboxed values) causes about a 1% perf regression, mostly in pyxl_bench.  I'm going to merge this PR and then reenable unboxed values in a new one.

PR #937, opened 2015-09-25
misc fixes for pyopenssl

undingen:


undingen:
I had to remove the actual pyOpenSSL test integration because there is still a problem. But should be ready soon...

undingen:
This uncovers a annoying crash inside ``` multiprocessing_ctypes_test.py``` which I can't reproduce locally :-(. Have too look into it in more detail otherwise it should be ready.

undingen:
as @Daetalus pointed out the ctypes issue is not specific to this changed #907 triggers the same error, therefore I would like to just skip the test and not block this PR and fix the issue in another PR.

PR #938, opened 2015-09-28
Some fix of Ellpisis

corona10:
* prevent overflow while calculating mid value (src/core/types.h)
*  some fix of Ellipsis
*  add Ellipsis test

before fixing  code
```
class TestEllipsis(object):
   def __getitem__(self, item):
      if item is Ellipsis:
         return "Ellipsis"
      else:
         return "return %r items" % item

print Ellipsis
v = Ellipsis
print v
print type(v)
x = TestEllipsis()
print x[2]
print x[...]
```
results as 

```
<ellipsis object at 0x127000c4a8>
<ellipsis object at 0x127000c4a8>
<type 'ellipsis'>
return 2 items
../../src/codegen/ast_interpreter.cpp:552: pyston::Value pyston::<anonymous namespace>::ASTInterpreter::visit_slice(pyston::AST_slice *): Assertion `0' failed: not implemented
Someone called abort!
  File "ellipsis_getname.py", line 16, in <module>:
    print x[...]
```

but now 

```
Ellipsis
Ellipsis
<type 'ellipsis'>
return 2 items
return Ellipsis items  ==> this should be Ellipsis it will fix at future..
```

### DIFF
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -788,7 +788,8 @@ int binarySearch(T needle, RandomAccessIterator start, RandomAccessIterator end,
     int l = 0;
     int r = end - start - 1;
     while (l <= r) {
-        int mid = l + (r - l) / 2;
+        //prevent overflow while calculating mid value
+        int mid = (l & r) + ((l ^ r) >> 1);;
         auto mid_item = *(start + mid);
         int c = cmp(needle, mid_item);
         if (c < 0)
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -99,6 +99,7 @@ private:
 
     Value visit_attribute(AST_Attribute* node);
     Value visit_dict(AST_Dict* node);
+    Value visit_ellipsis(AST_Ellipsis* node);
     Value visit_expr(AST_expr* node);
     Value visit_expr(AST_Expr* node);
     Value visit_extslice(AST_ExtSlice* node);
@@ -549,7 +550,7 @@ Value ASTInterpreter::visit_slice(AST_slice* node) {
         case AST_TYPE::ExtSlice:
             return visit_extslice(static_cast<AST_ExtSlice*>(node));
         case AST_TYPE::Ellipsis:
-            RELEASE_ASSERT(0, "not implemented");
+            return visit_ellipsis(static_cast<AST_Ellipsis*>(node));
             break;
         case AST_TYPE::Index:
             return visit_index(static_cast<AST_Index*>(node));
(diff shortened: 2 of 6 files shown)


src/core/types.h
```diff
@@ -788,7 +788,8 @@ int binarySearch(T needle, RandomAccessIterator start, RandomAccessIterator end,
     int l = 0;
     int r = end - start - 1;
     while (l <= r) {
-        int mid = l + (r - l) / 2;
```
Daetalus: `l` is zero. So I think here will not overflow.

corona10: @Daetalus l + (r - l) can be overflow if needle value is right-end near value,  l and r can be large
how about read this article 

http://googleresearch.blogspot.kr/2006/06/extra-extra-read-all-about-it-nearly.html

wangjb13: Yeah, thanks for your recommended reference, and your solution is absolutely right!  But just as Joshua Bloch said in your recommended reference, @Daetalus 's thought about "int mid = low + ((high - low) / 2);" is right, at least for a right BinarySearch implementation. After analyzing your recommended article throughly and thinking about your mentioned overflow carefully, your expected reference may be one of the references of your reference -- Salvatore Ruggieri's paper about the average of two numbers of arbitrary sign. After I scanned that paper roughly, our BinarySearch problem is one special case of that article -- non-negative-division calculation (Section 3) where both l and r are non-negative and l <= r, so I think it also supports Daetalus's implementation.

Thanks again! 
Jianbin Wang

wangjb13: For BinarySearch, 
      0 <= l <= r 
--> 0 <= r-l <= r
--> 0 <= (r-l)/2 <= r/2, it won't overflow.

0 <= l <= l + (r-l)/2 = (l + r)/2 <= r, it won't overflow.
As @corona10 you mentioned, if needle value is right-end near value, a bigger L will be reached when needle value is bigger than the maximum value (for calculating middle position, L will be equal to r), even in that case, (r-l)/2 will still not overflow, l + (r-l)/2 also won't overflow.

Jianbin Wang

corona10: Thanks @wangjb13  i fully understand what you'r saying.
so i think i need to revert my commit about 'src/core/types.h'
do you know the way to revert it?

kmod:
I think we already have an Ellipsis object -- check src/runtime/builtin_modules/builtins.cpp.  We should probably just make it available through a header (probably in src/runtime/types.h where we declare None/True/False), and then when we visit an ellipsis expression we just return the global singleton.

corona10:
@kmod i re-PR on #941 by what you're saying it fully passed what i want to do. :-) it works!
 can you review it??

PR #939, opened 2015-09-28
add m2crypto test + fix missing stuff

undingen:
- oldstyle classed don't add ```__bases__``` and ```__name__``` as attributes to their dict instead the intercept them with ```__getattribute__```
- it was possible to encounter a crash when importing a module and then deleting the entry from sys.modules and calling the function from the other module
```python
import gc, sys
def get_func():
 from t2 import f2
 print f2()
 del sys.modules["t2"]
 return f2
f = get_func()
gc.collect()
print f.__globals__ # <-- crashes here sometimes
f()                 # <-- this hits sometimes a release assert
```
- add a Pyston specific ```PyCode_GetArgCount``` instead of directly accessing ```code->co_argcount```
- In order to not have to patch swig for pyston I set the missing define ```SWIG_PYTHON_SLOW_GETSET_THIS``` with the ```CFLAGS``` environment variable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ addons:
     - libxml2-dev 
     - libxslt1-dev
     - libssl-dev
+    - swig
 
 before_install:
   - if [ "$CC" = "clang" ]; then export CC="clang-3.5" CXX="clang++-3.5"; fi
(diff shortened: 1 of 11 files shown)


undingen:
should be ready for review ;-)

kmod:
nice :)

PR #940, opened 2015-09-28
make platform test pass

aisk:
`make run_test_platform` is just to pass with one error, pyston have no `sys.subversion`.

In [this document](https://docs.python.org/2/library/sys.html#sys.subversion), `sys.subversion` is just a placeholder, so implement it as const value.

Or maybe we should modify the test just ignore it?

### DIFF
--- a/from_cpython/Lib/test/test_platform.py
+++ b/from_cpython/Lib/test/test_platform.py
@@ -1,4 +1,3 @@
-# expected: fail
 import sys
 import os
 import unittest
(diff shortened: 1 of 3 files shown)


kmod:
I think your change is reasonable -- looks like PyPy provides sys.subversion as well.  If test_platform.py is working now, can we remove the "expected: fail" annotation from it so that it will be included in the test suite?  And super minor, can we capitalize "Pyston" here?

aisk:
@kmod the first value of `sys.subversion` means the python's repo name, and Pyston's repo is lower case, so I think `pyston` is better?

kmod:
Well again it doesn't really matter, but cpython and pypy have capitalized values for sys.subversion and their repos are both lowercase.

aisk:
@kmod you're right :smile: , I capitalize it, and enabled the test suite, I think we could review this again.

kmod:
awesome, looks great :)

PR #941, opened 2015-09-29
Ellipsis implementation

corona10:
*  some fix of Ellipsis
*  test more about Ellipsis
```
class TestEllipsis(object):
   def __getitem__(self, item):
      if item is Ellipsis:
         return "Ellipsis"
      else:
         return "return %r items" % item

print Ellipsis
v = Ellipsis
print v
print type(v)
x = TestEllipsis()
print x[2]
print x[...]
```
results as 

```
<ellipsis object at 0x127000c4a8>
<ellipsis object at 0x127000c4a8>
<type 'ellipsis'>
return 2 items
../../src/codegen/ast_interpreter.cpp:552: pyston::Value pyston::<anonymous namespace>::ASTInterpreter::visit_slice(pyston::AST_slice *): Assertion `0' failed: not implemented
Someone called abort!
  File "ellipsis_getname.py", line 16, in <module>:
    print x[...]
```

but now 

```
Ellipsis
Ellipsis
<type 'ellipsis'>
return 2 items
Ellipsis
```

it fully passed

### DIFF
--- a/src/codegen/ast_interpreter.cpp
+++ b/src/codegen/ast_interpreter.cpp
@@ -99,6 +99,7 @@ private:
 
     Value visit_attribute(AST_Attribute* node);
     Value visit_dict(AST_Dict* node);
+    Value visit_ellipsis(AST_Ellipsis* node);
     Value visit_expr(AST_expr* node);
     Value visit_expr(AST_Expr* node);
     Value visit_extslice(AST_ExtSlice* node);
@@ -549,7 +550,7 @@ Value ASTInterpreter::visit_slice(AST_slice* node) {
         case AST_TYPE::ExtSlice:
             return visit_extslice(static_cast<AST_ExtSlice*>(node));
         case AST_TYPE::Ellipsis:
-            RELEASE_ASSERT(0, "not implemented");
+            return visit_ellipsis(static_cast<AST_Ellipsis*>(node));
             break;
         case AST_TYPE::Index:
             return visit_index(static_cast<AST_Index*>(node));
@@ -559,6 +560,10 @@ Value ASTInterpreter::visit_slice(AST_slice* node) {
             RELEASE_ASSERT(0, "Attempt to handle invalid slice type");
     }
     return Value();
+}
+
+Value ASTInterpreter::visit_ellipsis(AST_Ellipsis* node) {
+    return Value(Ellipsis, jit ? jit->imm(Ellipsis) : NULL);
 }
 
 Value ASTInterpreter::visit_slice(AST_Slice* node) {
--- a/src/codegen/irgen/irgenerator.cpp
+++ b/src/codegen/irgen/irgenerator.cpp
@@ -1073,6 +1073,11 @@ private:
         return rtn;
     }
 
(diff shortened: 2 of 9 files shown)


src/codegen/ast_interpreter.cpp
```diff
@@ -561,6 +562,12 @@ Value ASTInterpreter::visit_slice(AST_slice* node) {
     return Value();
 }
 
+Value ASTInterpreter::visit_ellipsis(AST_Ellipsis* node) {
+    Value v;
```
kmod: Could this be just `return v`?

kmod: err sorry, `return Ellipsis`

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -1250,6 +1250,9 @@ extern "C" PyObject* PyEval_GetBuiltins(void) noexcept {
     return builtins_module;
 }
 
+Box* ellipsisRepr(Box* self) {
+    return boxStringTwine(llvm::Twine("Ellipsis"));
```
kmod: This could just be `boxString("Ellipsis")`.

src/codegen/ast_interpreter.cpp
corona10: @kmod is there any way to Box* to Value type??

undingen: You also have to ether call ```abortJITing()``` inside this function or tell the baseline jit how to generate code for the ```ellipsis``` node. Sorry that's not really obvious :-( - our interpreter calls in the appropriate functions of the baseline jit to generate the requested code.
With your current code we would probably hit some assert if a ellipsis ast node is contained in a function which gets called more than 25 times (=baseline jit kicks in) because only the ```Box* o``` field is set but not the ```RewriterVar* var``` field of the ```Value``` object.
So we could call ```abortJITing()``` but the much better solution is to actually teach the basline jit how to it can generate machine code for this node and in this case this is also very easy :-).
```return Value(Ellipsis, jit ? jit->imm(Ellipsis) : NULL);``` should do it.

**Comment by undingen** [comment]
I wrote some additional comments inside our gitter channel because I thought you are around and it's easier to discuss things there :-)

File: src/codegen/irgen/irgenerator.cpp [hunk 3]
```diff
@@ -1073,6 +1073,10 @@ class IRGeneratorImpl : public IRGenerator {
         return rtn;
     }
 
+    ConcreteCompilerVariable* getEllipsis() {
+        llvm::Constant* ellipsis = embedRelocatablePtr(Ellipsis, g.llvm_value_type_ptr, "cEllipsis");
+        return new ConcreteCompilerVariable(typeFromClass(type_cls), ellipsis, false);
```
kmod: I think this should be `typeFromClass(ellipsis_cls)`

corona10: @kmod i fix it as you said that, but i user `Ellipsis->cls` rather than `ellipsis_cls`.

kmod:
awesome, thanks for the change :)

PR #942, opened 2015-09-29
implement PyObject_AsWriteBuffer

aisk:
To make `run_multiprocessing_ctypes_test` and `run_test_sqlite.SqliteTypeTest.CheckBlob` pass. Just copied Python's `PyObject_AsWriteBuffer` with little modify.

And I think `PyObject_AsWriteBuffer` is better written in `abstract.cpp`, but somehow it exits in `object.cpp`, maybe by mistake, so I changed this, or maybe threre were some reasons I don't know?

### DIFF
--- a/src/capi/abstract.cpp
+++ b/src/capi/abstract.cpp
@@ -605,6 +605,32 @@ extern "C" int PyObject_AsReadBuffer(PyObject* obj, const void** buffer, Py_ssiz
     return 0;
 }
 
+extern "C" int PyObject_AsWriteBuffer(PyObject* obj, void** buffer, Py_ssize_t* buffer_len) noexcept {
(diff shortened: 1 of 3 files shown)


Daetalus:
The Pyston ctypes support have some issues. I suspect it may could not get passed in CI.

aisk:
@Daetalus I implement this function is just to let sqlite test pass, and then found `multiprocessing_ctypes` is passed yet ... But I'm already test `make check` on local, so maybe other ctypes issue is already fixed?

aisk:
Ah I saw @undingen 's PR is already fixed this 😅

undingen:
yeah I added the exact same function already :-(

PR #944, opened 2015-09-30
Remove unneeded pyston workarounds + fix misc cpython test errors 

undingen:


undingen:
this sadly slowdowns pyxl by about 4%. Quick measurements show that about half of this is coming from adding weakref support to old style classes and the other part is coming from the collections.py change..

kmod:
ah too bad about the slowdown.  maybe we can punt on those particular issues for now and get the rest of it in?

undingen:
ooh I forgot about this PR...
I rebased it and removed the ```collections.py``` change (means I left the pyston change to the file in) because it looked like this caused most of the perf regression.
This PR still regresses the perf slightly (I think this is mostly caused by making old style classes weakref able) but it's less than 1% so I think we should just merge it in.

kmod:
nice :)

PR #949, opened 2015-10-02
Add more asserts to aggressively enforce return conventions

kmod:
You can't get the return value from a RewriteArgs without
looking at the return convention.  Debug mode will check the
return convention during the execution of rewrites.

Split "VALID_RETURN" into "HAS_RETURN" and "CAPI_RETURN" since some
places used it to mean either.

Hopefully this helps keep things clean.  There are a lot of places that
were implictly assuming a certain return convention, but now that they
have to explicitly assume it, the issues are more obvious.  Plus
they will get checked in the debug builds.

Also tried to fix things up while going through and doing this refactoring;
I think I found a number of issues.

### DIFF
--- a/src/asm_writing/rewriter.h
+++ b/src/asm_writing/rewriter.h
@@ -277,6 +277,8 @@ public:
         assert(rewriter);
     }
 
+    Rewriter* getRewriter() { return rewriter; }
+
     friend class Rewriter;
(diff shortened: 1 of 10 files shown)


undingen:
nice, this change makes our rewrite paths much clearer :+1:

PR #950, opened 2015-10-02
use a template to remove rewriter checks inside our most important runtime funcs

undingen:
use a template to remove rewriter checks inside our most important runtime funcs when we know that the rewrite_args are NULL.
This is similar to #657 but uses a template arg. (as long as we only have two template args I think the are fine to use ... three would be bad :-D)
I looked at the assembly which clang generated and it's able to remove all rewriter specific code just by setting the ```rewriter_args``` to NULL in the beginning of the function.

Sadly this touches quite a lot of lines but I don't think the change is likely to introduce any bug because of this assert I added ```if (rewritable == NOT_REWRITABLE) {       assert(!rewrite_args); ```
```
                                origin/rewritable~:   origin/rewritable:
           django_template3.py             2.7s (4)             2.7s (4)  -1.3%
                 pyxl_bench.py             2.2s (4)             2.2s (4)  -1.7%
     sqlalchemy_imperative2.py             2.7s (4)             2.6s (4)  -2.4%
                       geomean                 2.5s                 2.5s  -1.8%
```

### DIFF
--- a/src/capi/abstract.cpp
+++ b/src/capi/abstract.cpp
@@ -758,7 +758,7 @@ exit:
 }
 
 extern "C" Py_ssize_t PyObject_Size(PyObject* o) noexcept {
-    BoxedInt* r = lenInternal<ExceptionStyle::CAPI>(o, NULL);
+    BoxedInt* r = lenInternal<ExceptionStyle::CAPI, NOT_REWRITABLE>(o, NULL);
     if (!r)
         return -1;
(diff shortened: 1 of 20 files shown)


kmod:
nice :)  I'm seeing a merge conflict though, presumably with the high-LOC change (#949) I just merged.

undingen:
I fixed the merge conflicts and merged this in

PR #952, opened 2015-10-05
microptimize callFunc() by only calling getFunctionName() when necessary

undingen:
even though getFunctionName() is fast it showed up in the profiler because it get's called very often,
but most of the time the actual value is not used.
```
           django_template3.py             2.7s (4)             2.7s (4)  -0.1%
                 pyxl_bench.py             2.3s (4)             2.2s (4)  -2.8%
     sqlalchemy_imperative2.py             2.7s (4)             2.6s (4)  -0.7%
                       geomean                 2.5s                 2.5s  -1.2%
```

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -3232,9 +3232,10 @@ enum class KeywordDest {
     POSITIONAL,
     KWARGS,
 };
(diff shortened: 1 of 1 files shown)


kmod:
oh wow, that helps a lot more than I would have guessed!

PR #955, opened 2015-10-06
Use travis_wait to workaround the no output received error

undingen:
Fixes:
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.


undingen:
I merging this in because it looks like all debug builds are hitting this problem and I want to get them green. I hope a post commit review is fine for this change.

PR #956, opened 2015-10-06
Add rewrite support for calls which use keyword args

undingen:
```
           django_template3.py             2.7s (4)             2.7s (4)  -0.7%
                 pyxl_bench.py             2.2s (4)             2.2s (4)  -1.0%
     sqlalchemy_imperative2.py             2.6s (4)             2.6s (4)  +0.1%
                       geomean                 2.5s                 2.5s  -0.5%
```


src/runtime/objmodel.cpp
```diff
@@ -3233,9 +3233,9 @@ enum class KeywordDest {
     KWARGS,
 };
 template <typename FuncNameCB>
-static KeywordDest placeKeyword(const ParamNames* param_names, llvm::SmallVector<bool, 8>& params_filled,
```
kmod: can we get rid of the KeywordDest enum?  otherwise lgtm :)

PR #957, opened 2015-10-07
Compat fixes

kmod:
crazy corner cases.


kmod:
These have all gotten merged now

PR #959, opened 2015-10-07
disable moderate opt and pyston passes and increase reopt threshold

undingen:
This is based on #958 but adds one commit which disabled our moderate llvm tier (/ we will only have one llvm tier) and disabled our pyston llvm passes because they look to be useless at the moment.
I split this out into a separate PR because I'm not sure if we really want this change to land but perf looks quite good:
```
origin/disable_mod~:  origin/disable_mod:
           django_template3.py             2.4s (4)             2.4s (4)  -2.8%
                 pyxl_bench.py             2.2s (4)             2.1s (4)  -4.1%
     sqlalchemy_imperative2.py             2.6s (4)             2.6s (4)  -0.5%
       django_template3_10x.py            15.8s (4)            15.3s (4)  -3.2%
             pyxl_bench_10x.py            17.0s (4)            16.3s (4)  -4.2%
 sqlalchemy_imperative2_10x.py            20.0s (4)            19.6s (4)  -1.5%
                       geomean                 6.5s                 6.3s  -2.7%


```


undingen:
Disabling the moderate pass made a stackmap bug appear which I introduced when adding the support for symbolic constants (see last commit).
We definitely want to have this bug fixed so if we decide to not disable the moderate tier I will split out the fix into a new PR.

kmod:
nice :)  Sometimes I worry about the effects of disabling these kinds of features -- it raises the bar for getting them turned back on.  but I don't think turning them off is any worse than having a poor implementation of it turned on :P

PR #960, opened 2015-10-08
Abstract class improvements

Daetalus:
The originally purpose of this PR was try to let `test_collections` could pass. At the end, it also fixed some issues in `test_abc`.

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -2820,6 +2820,59 @@ static int object_init(PyObject* self, PyObject* args, PyObject* kwds) noexcept 
     return err;
 }
 
+static int type_set_abstractmethods(PyTypeObject* type, PyObject* value, void* context) noexcept {
+    /* __abstractmethods__ should only be set once on a type, in
+       abc.ABCMeta.__new__, so this function doesn't do anything
+       special to update subclasses.
+    */
+    int abstract, res;
+    if (value != NULL) {
+        abstract = PyObject_IsTrue(value);
+        if (abstract < 0)
+            return -1;
+        res = PyDict_SetItemString(type->tp_dict, "__abstractmethods__", value);
+    } else {
+        abstract = 0;
+        res = PyDict_DelItemString(type->tp_dict, "__abstractmethods__");
+        if (res && PyErr_ExceptionMatches(PyExc_KeyError)) {
+            PyErr_SetString(PyExc_AttributeError, "__abstractmethods__");
+            return -1;
+        }
+    }
+    if (res == 0) {
+        PyType_Modified(type);
+        if (abstract)
+            type->tp_flags |= Py_TPFLAGS_IS_ABSTRACT;
+        else
+            type->tp_flags &= ~Py_TPFLAGS_IS_ABSTRACT;
+    }
+    return res;
+}
+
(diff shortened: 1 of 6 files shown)


kmod:
Hey, I took a look at this old PR and tracked the issue down to the rewrite flags being wrong in a certain exceptional case.  Can you try merging this in and seeing if it passes CI?
```patch
diff --git a/src/runtime/objmodel.cpp b/src/runtime/objmodel.cpp
index 81e0456..3f4a35d 100644
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -1443,6 +1443,8 @@ Box* dataDescriptorInstanceSpecialCases(GetattrRewriteArgs* rewrite_args, BoxedS
             rewrite_args = NULL;
         }
 
+        Box* rtn = getset_descr->get(obj, getset_descr->closure);
+
         if (rewrite_args) {
             // hmm, maybe we should write assembly which can look up the function address and call any function
             r_descr->addAttrGuard(offsetof(BoxedGetsetDescriptor, get), (intptr_t)getset_descr->get);
@@ -1454,8 +1456,7 @@ Box* dataDescriptorInstanceSpecialCases(GetattrRewriteArgs* rewrite_args, BoxedS
             rewrite_args->setReturn(r_rtn, descr->cls == capi_getset_cls ? ReturnConvention::CAPI_RETURN
                                                                          : ReturnConvention::HAS_RETURN);
         }
-
-        return getset_descr->get(obj, getset_descr->closure);
+        return rtn;
     }
 
     return NULL;
```

Also, I know this was from a while ago, but do you know/remember what those deletions are for in test_collections?  Anyway, this PR looks pretty good to go and it'd be nice to have it in :)

kmod:
After thinking about it some more, I think the fix is a little bit more involved, so I split it off into #1102.

Daetalus:
@kmod updated. Please take a look if you have time. Thank you!

src/runtime/types.cpp
```diff
@@ -2820,6 +2820,59 @@ static int object_init(PyObject* self, PyObject* args, PyObject* kwds) noexcept
...
+    if (value != NULL) {
+        abstract = PyObject_IsTrue(value);
+        if (abstract < 0)
+            return -1;
+        res = PyDict_SetItemString(type->tp_dict, "__abstractmethods__", value);
+    } else {
+        abstract = 0;
+        res = PyDict_DelItemString(type->tp_dict, "__abstractmethods__");
+        if (res && PyErr_ExceptionMatches(PyExc_KeyError)) {
+            PyErr_SetString(PyExc_AttributeError, "__abstractmethods__ FUck?");
```
kmod: debugging edit? :)

kmod:
Looks good, though do you remember what those edits to test_collections.py are from?  I think it might just be from us using different CPython revisions at different times.  Could you try checking out CPython's 2.7.7 release (`hg checkout -U v2.7.7` in your cpython checkout) and then copying the file over?

src/runtime/types.cpp
Daetalus: Opps... Sorry...

Daetalus:
@kmod the `test_collections` was copied from CPython 2.7.7.

kmod:
:+1:

PR #962, opened 2015-10-09
handle the situation which provide a directory to pyston

### LINKED ISSUE #800: Running Pyston on a directory doesn't do the right thing
for example `./pyston_dbg /`

Looks like it's supposed to look for a `__main__` module, and throw an exception if it doesn't find one.  We exit with code 0; not sure if we looked for a main module or not.

Daetalus:
closes #800

### DIFF
--- a/src/jit.cpp
+++ b/src/jit.cpp
@@ -17,6 +17,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <fcntl.h>
+#include <limits.h>
 #include <readline/history.h>
 #include <readline/readline.h>
 #include <stdint.h>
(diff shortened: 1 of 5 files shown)


kmod:
hey, is this ready to review?

kmod:
awesome :)  should have mentioned though -- can we get a test for this?  We can add a subdirectory inside test/tests, and then create a test file that tries to run it.

Daetalus:
Will include in next generic commit. My original plan is use subprogress to get output. Is that is a correct way?

kmod:
yep, sounds good.  dash_c.py seems like a decent starting point.

Daetalus:
Thanks!

PR #966, opened 2015-10-14
typeLookup: guard on tp_version_tag instead of mro

undingen:
This if WIP but before spending more time on it I would like to get feedback on it.
This removes the guarding ```Box::getattr``` adds for types which have the ```HAVE_VERSION_TAG``` set (aka. all types except capi extensions types which are not part of pyston)
instead it guards on the ```tp_version_tag```. 
There is currently a problem that the version tags are only 32bit and can overflow. So in case they overflow we need some mechanism to invalidate all ICs....

Perf change of the last commit:
```
           django_template3.py             2.3s (4)             2.3s (4)  -1.2%
                 pyxl_bench.py             2.0s (4)             2.0s (4)  -2.4%
     sqlalchemy_imperative2.py             2.5s (4)             2.5s (4)  -2.6%
                       geomean                 2.3s                 2.2s  -2.1%
```

### DIFF
--- a/src/runtime/objmodel.cpp
+++ b/src/runtime/objmodel.cpp
@@ -957,7 +957,9 @@ extern "C" PyObject* _PyType_Lookup(PyTypeObject* type, PyObject* name) noexcept
 #define MCACHE_CACHEABLE_NAME(name) PyString_CheckExact(name) && PyString_GET_SIZE(name) <= MCACHE_MAX_ATTR_SIZE
 
 struct method_cache_entry {
-    unsigned int version;
+    // Pyston change:
+    // unsigned int version;
+    PY_UINT64_T version;
     PyObject* name;  /* reference to exactly a str or None */
     PyObject* value; /* borrowed */
 };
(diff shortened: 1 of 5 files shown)


src/runtime/objmodel.cpp
```diff
@@ -1048,6 +1072,31 @@ template <Rewritable rewritable> Box* typeLookup(BoxedClass* cls, BoxedString* a
...
+            if (attr->hash == -1)
+                strHashUnboxed(attr);
+
+            /* fast path */
+            unsigned int h = MCACHE_HASH_METHOD(cls, attr);
+            if (method_cache[h].version == cls->tp_version_tag && method_cache[h].name == attr) {
+                if (rewrite_args) {
+                    RewriterVar* obj_saved = rewrite_args->obj;
+                    obj_saved->addAttrGuard(offsetof(BoxedClass, tp_flags), (intptr_t)cls->tp_flags);
+                    obj_saved->addAttrGuard(offsetof(BoxedClass, tp_version_tag), (intptr_t)cls->tp_version_tag);
```
kmod: I think this does a 64-bit guard?  Maybe we should just make tp_version_tag 64 bits, which would let us avoid the wrap-around issue too :P

kmod: (this might apply to the tp_flags guard as well)

kmod:
sounds good to me :)  we have the ability to invalidate ICs; the way other places do it is to create in ICInvalidator, and when the rewriter uses the thing that needs invalidation, the to register the rewrite with the ICInvalidator.  We could create a global "clear_on_tag_wraparound" invalidator, or if it's going to contain all ICs then we could just have a function that iterates over the ics_by_return_addr list.

undingen:
I moved to using a 64bit counter but let the hash function the same (only uses the lower 32bit of the version tag) because I think this is still reasonable and moving to a more complicated hash makes the perf improvement go away...

kmod:
can you put an `abort()` in assign_version_tag in the "detect wraparound" case?  otherwise lgtm

undingen:
I merged this with the wraparound check

PR #969, opened 2015-10-15
Reenable unboxed values

kmod:
Feels like more work than it was worth:
```
           django_template3.py             2.2s (2)             2.2s (2)  -0.1%
                 pyxl_bench.py             2.0s (2)             1.9s (2)  -3.6%
     sqlalchemy_imperative2.py             2.4s (2)             2.4s (2)  +0.0%
       django_template3_10x.py            14.3s (2)            14.4s (2)  +0.9%
             pyxl_bench_10x.py            15.4s (2)            15.1s (2)  -2.5%
 sqlalchemy_imperative2_10x.py            18.5s (2)            18.8s (2)  +1.3%
                       geomean                 6.0s                 5.9s  -0.7%
```

For now unboxed values can't get passed between BBs (compiler types don't have a hook to specify how to phi themselves, but that shouldn't be too hard to add if we need it).  Even with this limitation we seem to get most (but not all) of the benefit.

I think there's still some cases that this is missing, but after working on this for a few days I think it's time to check in what's working and I'll just fix any issues if we notice any.

### DIFF
--- a/src/analysis/type_analysis.cpp
+++ b/src/analysis/type_analysis.cpp
@@ -197,7 +197,7 @@ private:
         }
     }
 
-    void* visit_ellipsis(AST_Ellipsis* node) override { return ELLIPSIS; }
+    void* visit_ellipsis(AST_Ellipsis* node) override { return typeFromClass(ellipsis_cls); }
 
     void* visit_attribute(AST_Attribute* node) override {
         CompilerType* t = getType(node->value);
--- a/src/analysis/type_analysis.cpp
+++ b/src/analysis/type_analysis.cpp
@@ -62,10 +62,8 @@ static CompilerType* unboxedType(ConcreteCompilerType* t) {
         return BOOL;
     if (t == BOXED_INT)
         return INT;
-#if ENABLE_UNBOXED_VALUES
     if (t == BOXED_FLOAT)
         return FLOAT;
-#endif
     return t;
 }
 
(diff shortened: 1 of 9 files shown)


smarx:

@kmod, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

Daetalus:
The message of Dropbox CLA bot.... LOL

undingen:
from what I understood this PR looks good to me :-)

kmod:
well the bot was right, I hadn't signed the CLA :P

kmod:
@undingen thanks for looking at the pr :)

src/analysis/type_analysis.cpp
```diff
@@ -57,15 +57,13 @@ ConcreteCompilerType* NullTypeAnalysis::getTypeAtBlockEnd(InternedString name, C
 
 
 // Note: the behavior of this function must match irgenerator.cpp::unboxVar()
-static ConcreteCompilerType* unboxedType(ConcreteCompilerType* t) {
+static CompilerType* unboxedType(ConcreteCompilerType* t) {
     if (t == BOXED_BOOL)
         return BOOL;
-#if ENABLE_UNBOXED_VALUES
```
Daetalus: I don't quite understand these code for now. But these change cause the error in #971. (I am reading these part in these days. Hope I could gather enough questions before 0.4 release. )

kmod: Hmm yeah this PR did introduce that.  Can you try merging #973 to see if that helps?

Daetalus: Thanks! I will try that.

PR #970, opened 2015-10-15
object cache: hash IR before running any opt passes

undingen:
This lets us to skip running the opt passes when we find a cached object file.
This has the disadvantage that opt passes are not allowed to call embedRelocatablePtr()
but has the advantage that we could do more costly optimizations in the future without hurting the perf on the second run.

```
                                   upstream/master:  origin/cache_optim:
           django_template3.py             2.4s (4)             2.2s (4)  -4.5%
                 pyxl_bench.py             2.0s (4)             1.9s (4)  -3.8%
     sqlalchemy_imperative2.py             2.5s (4)             2.4s (4)  -3.4%
                       geomean                 2.3s                 2.2s  -3.9%
```

### DIFF
--- a/src/codegen/codegen.h
+++ b/src/codegen/codegen.h
@@ -31,6 +31,8 @@ class TargetMachine;
 
 namespace pyston {
 
+class PystonObjectCache;
+
 class FunctionAddressRegistry {
 private:
     struct FuncInfo {
@@ -62,6 +64,7 @@ struct GlobalState {
     CompiledFunction* cur_cf;
     llvm::TargetMachine* tm;
     llvm::ExecutionEngine* engine;
+    PystonObjectCache* object_cache;
 
     std::vector<llvm::JITEventListener*> jit_listeners;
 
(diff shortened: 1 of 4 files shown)


smarx:

@undingen, it looks like you've already signed the Dropbox CLA. Thanks!

undingen:
should have split out the movement of classes into a separate commit to make reviewing easier... :-(
let me know if it to hard to review and I will split it

kmod:
what will happen if we add a pass that tries to add a new relocatable symbol?  I think we're going to want that in the future; I think that shouldn't stop us from merging this though (4% is hard to pass on).

kmod:
(lgtm)

undingen:
The problem is that our current object cache works by assigning for every pointer we embed a sym name which is just a sequential number and remember the current pointer value.
If a optimization pass would embed new pointers with this simple relocation mechanism we would hash the module to the same value on the next run (because the hash is calculated before we do opts) but we would not know what actual values the symbols should have which got introduced during the optimizations (because we never called ```embedRelocatablePtr``` for them). So we would hit this ```RELEASE_ASSERT(0, "Could not find sym: %s", name.c_str());``` assert inside ```PystonMemoryManager::getSymbolAddress``` 
I think we will be able to come up with mechanisms to workaround this problem for many cases when we introduce a pass which hits this case.

PR #971, opened 2015-10-15
Some improvements that let test_fractions could pass.

Daetalus:


smarx:

@Daetalus, it looks like you've already signed the Dropbox CLA. Thanks!

from_cpython/Lib/test/test_long.py
```diff
@@ -17,7 +16,10 @@ def __str__(self):
         return self.format % self.args
 
 # SHIFT should match the value in longintrepr.h for best testing.
-SHIFT = sys.long_info.bits_per_digit
+SHIFT = 30
```
kmod: Did you pick 30 since it's cpython's value?  I would argue for 64 here to match GMP's internal structure (though it probably doesn't matter that much since I doubt we'll expose bugs in GMP) and since it seems like a higher value mostly means more testing.

src/runtime/classobj.cpp
```diff
@@ -1489,8 +1489,18 @@ Box* instanceLong(Box* _inst) {
...
+        return runtimeCall(long_func, ArgPassSpec(0), NULL, NULL, NULL, NULL, NULL);
+    }
+
+    Box* truncated = instanceTrunc(inst);
+    /* __trunc__ is specified to return an Integral type, but
+       long() needs to return an int. */
+    Box* res = _PyNumber_ConvertIntegralToInt(truncated, "__trunc__ returned non-Integral (type %.200s)");
+    if (!res)
+        throwCAPIException();
+    return res;
```
kmod: Looking at the cpython code, I think this is supposed to call instanceInt

src/runtime/float.cpp
```diff
@@ -124,7 +124,11 @@ extern "C" Box* floatAdd(BoxedFloat* lhs, Box* rhs) {
     } else if (rhs->cls == float_cls) {
         return floatAddFloat(lhs, static_cast<BoxedFloat*>(rhs));
     } else if (rhs->cls == long_cls) {
-        return boxFloat(lhs->d + PyLong_AsDouble(rhs));
+        double rhs_f = PyLong_AsDouble(rhs);
+        if (PyErr_Occurred()) {
```
kmod: Minor, but I think PyLong_AsDouble is supposed to return -1.0 when an exception is thrown, and checking the return value against -1.0 should be slightly faster than going to PyErr_Occurred every time.  (we still need to call it if the value is -1.0, but that should be relatively rare)

src/runtime/long.cpp
```diff
@@ -463,7 +467,14 @@ extern "C" PyObject* PyLong_FromSize_t(size_t ival) noexcept {
 #undef IS_LITTLE_ENDIAN
 
 extern "C" double _PyLong_Frexp(PyLongObject* a, Py_ssize_t* e) noexcept {
-    Py_FatalError("unimplemented");
+    BoxedLong* v = (BoxedLong*)a;
+    double result = mpz_get_d_2exp(e, v->n);
+    if (*e > INT_MAX) {
```
kmod: Any reason to use INT_MAX here instead of PY_SSIZE_T_MAX?

If we're checking overflow against PY_SSIZE_T_MAX, I'm not sure we can use the result of mpz_get_d_2exp to do that; not really sure what gmp would do in that situation.  I think this only matters on 32-bit systems (otherwise the long object itself would require 2^60 bytes of memory to overflow here), so maybe we could replace this block with a `static_assert(sizeof(Py_ssize_t) == 8, "need to add overflow checking")`?

src/runtime/long.cpp
```diff
@@ -1241,18 +1252,32 @@ Box* longTrueDiv(BoxedLong* v1, Box* _v2) {
...
-    return boxFloat(lhs / (double)rhs);
+    }
+
+    mpfr_t lhs_f, rhs_f, result;
+    mpfr_init(result);
+    mpfr_init_set_z(lhs_f, v1->n, MPFR_RNDN);
+    mpfr_init_set_z(rhs_f, v2->n, MPFR_RNDZ);
+    mpfr_div(result, lhs_f, rhs_f, MPFR_RNDN);
+
+    double result_f = mpfr_get_d(result, MPFR_RNDN);
```
kmod: I can't really comment on the rounding modes... do you feel confident in them?

Also, it looks like cpython (in most cases) doesn't convert to a double to do the division.  Not sure how important it is, but maybe another approach is to try to copy their implementation?  That might let us get all the corner cases.

from_cpython/Lib/test/test_long.py
```diff
@@ -749,45 +751,50 @@ def test_float_overflow(self):
...
+    #                  "math.sin(huge)", "math.sin(mhuge)",
+    #                  "math.sqrt(huge)", "math.sqrt(mhuge)", # should do better
+    #                  "math.floor(huge)", "math.floor(mhuge)"]:
+    #
+    #         self.assertRaises(OverflowError, eval, test, namespace)
+    #
+    #         # XXX Perhaps float(shuge) can raise OverflowError on some box?
+    #         # The comparison should not.
+    #         self.assertNotEqual(float(shuge), int(shuge),
+    #             "float(shuge) should not equal int(shuge)")
```
kmod: At first glance it seems like this section shouldn't be very susceptible to rounding issues; I tried running it and the issues I'm seeing are that we're missing `float.__rfloordiv__`, and PyFloat_AsDouble doesn't check for overflow.  If I change PyFloat_AsDouble to call PyLong_AsDouble instead of directly calling mpz_get_d, it seems to make most of the math tests pass.

from_cpython/Lib/test/test_long.py
```diff
@@ -749,45 +751,50 @@ def test_float_overflow(self):
...
+    #
+    #         # log10(value) == exp, so log(value) == log10(value)/log10(e) ==
+    #         # exp/log10e
+    #         expected = exp / log10e
+    #         log = math.log(value)
+    #         self.assertalmostequal(log, expected)
+    #
+    #     for bad in -(1l << 10000), -2l, 0l:
+    #         self.assertraises(valueerror, math.log, bad)
+    #         self.assertraises(valueerror, math.log10, bad)
```
kmod: Since we're providing the math module, I think we should try to fix it -- I think this might just be an issue of changing the call from PY_SIZE to _PyLong_Sign.

kmod:
oh man, this one is tricky :)  looks pretty good though.

Daetalus:
Oh man, thanks for your precious time. I think should waste your time on these tricky things. Since we made it here. I will improve this the PR according your comments. Thanks again!

from_cpython/Lib/test/test_long.py
Daetalus: Opps.. Sorry, I mixed up this with `test_long_future`, so comment a wrong test and hide some issues. I will update it.

src/runtime/long.cpp
Daetalus: This implementation support all long range, better than only support args which fit into an int. But could not handle some corner cases in `test_long_future`. 

eg.
```
DBL_MAX = sys.float_info.max
DBL_MAX_EXP = sys.float_info.max_exp
DBL_MIN_EXP = sys.float_info.min_exp
DBL_MANT_DIG = sys.float_info.mant_dig
DBL_MIN_OVERFLOW = 2**DBL_MAX_EXP - 2**(DBL_MAX_EXP - DBL_MANT_DIG - 1)
(X * DBL_MIN_OVERFLOW - 2) / X
```
The value always truncated inside MPFR, the last digit of the result is not same.

CPython implementation base on digits long implementation. The last half of the CPython function heavily depend on it. Apply some Pyston changes seems won't be easy( just take a quick glance, no deep thinking). Maybe when we completely use CPython long implementation, these crazy corner cases could been solved.

In order to use `mpfr_div`, I have to convert to a double...

Wait for your decision.

Daetalus:
Ready to review, @kmod

kmod:
sorry for the delay -- looks pretty good but I think we should take another look at longHash and then I think it's done :)

Daetalus:
@kmod updated and ready for review.

test/tests/long.py
```diff
@@ -146,3 +146,14 @@ class long2(long):
...
+try:
+    print(long('hek2mgl', 22))
+except Exception as e:
+    print(e.message)
+
+print(long('hek2mgl', 23))
+print(long('hek2mgl', 24))
+
+for i in range(-10, 10):
+    print(hash((-1 << 63) + i))
```
kmod: can we also add `print(hash((1 << 63) + i))`? I think we get those a little bit different

Daetalus:
@kmod It ready to review now.

src/runtime/long.cpp
```diff
@@ -1393,14 +1424,20 @@ Box* longHash(BoxedLong* self) {
...
+
+    // if the sign of remainder is positive but self is negtive, it means the remainder overflowed.
+    // So compensate by incrementing. This preserves the value modulo ULONG_MAX.
+    if ((remainder >= 0) && (mpz_sgn(self->n) < 0))
+        remainder++;
+
+    if (remainder == -1)
+        remainder = -2;
+
+    return boxInt(remainder);
```
kmod: Ok, I think I figured this out :)  The tricky thing is that if `self` is divisible by ULONG_MAX, then CPython's "compute self mod ULONG_MAX" operation will return ULONG_MAX, not 0.  And ULONG_MAX is -1 so it will then get changed to -2.

After some hacking around, this is what I came up with:
```C++
unsigned long remainder = mpz_tdiv_ui(self->n, ULONG_MAX);
if (remainder == 0)
  remainder = -1; // CPython compatibility -- ULONG_MAX mod ULONG_MAX is ULONG_MAX to them.
remainder *= mpz_sgn(self->n);
if (remainder == -1)
  remainder = -2;
return boxInt(remainder);
```

Can we also add this to the test:
```Python
for i in xrange(100):
  for j in xrange(100):
    print i, j, hash((1 << i) - (1 << j))
```

Daetalus:
Sorry, forgot to mention, it is ready to review.

kmod:
Awesome, thanks for working through this, it was pretty tricky!

PR #972, opened 2015-10-16
Avoid creating most slice objects

kmod:
We've been allocating slice objects for slicing operations, but
CPython's internal slice methods already take separate start+stop
arguments.  So if we see a slice, instead of creating the slice
and sending it off to getitem, try calling PySequence_GetSlice.

This will be slower for classes with user-defined __getitem__
functions that can handle slices; we can fix that by adding rewriting
to this new endpoint, but it seems to not matter too much right now.

```
           django_template3.py             2.4s (4)             2.4s (4)  +0.5%
                 pyxl_bench.py             2.1s (2)             2.1s (2)  +0.8%
     sqlalchemy_imperative2.py             2.5s (2)             2.5s (2)  -0.3%
       django_template3_10x.py            14.8s (2)            14.8s (2)  +0.2%
             pyxl_bench_10x.py            15.7s (2)            15.1s (2)  -3.6%
 sqlalchemy_imperative2_10x.py            18.7s (2)            18.5s (2)  -0.9%
                       geomean                 6.2s                 6.1s  -0.6%
```

### DIFF
--- a/src/codegen/compvars.cpp
+++ b/src/codegen/compvars.cpp
@@ -343,6 +343,30 @@ public:
 
     CompilerVariable* getitem(IREmitter& emitter, const OpInfo& info, ConcreteCompilerVariable* var,
                               CompilerVariable* slice) override {
+        if (slice->getType() == UNBOXED_SLICE) {
+            UnboxedSlice slice_val = extractSlice(slice);
+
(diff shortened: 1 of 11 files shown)


undingen:
lgtm

kmod:
thanks for taking a look :)

PR #976, opened 2015-10-19
rewrite instance_setattro

undingen:


kmod:
lgtm

undingen:
ok looks like this is not yet ready for merging: produces a 5% regression for pyxl2. will take a look into it tomorrow.

undingen:
I merged this in: after rebasing it didn't look like such a large slowdown anymore I could not find why it should be slower and there is a lot of variance

PR #979, opened 2015-10-22
Switch to 2byte unicode implementation like cpython does on linux

undingen:
```
                                   upstream/master:   origin/2b_unicode:
           django_template3.py             2.2s (4)             2.1s (4)  -2.0%
                 pyxl_bench.py             1.8s (4)             1.8s (4)  -0.3%
     sqlalchemy_imperative2.py             2.3s (4)             2.3s (4)  -0.1%
                pyxl_bench2.py             1.3s (4)             1.3s (4)  +2.0%
       django_template3_10x.py            14.2s (4)            13.8s (4)  -3.4%
             pyxl_bench_10x.py            14.3s (4)            13.8s (4)  -3.3%
 sqlalchemy_imperative2_10x.py            17.5s (4)            17.5s (4)  +0.2%
            pyxl_bench2_10x.py            12.2s (4)            12.1s (4)  -0.2%
                       geomean                 5.2s                 5.1s  -0.9%
```


kmod:
Unless you feel pretty confident about this, we might want to hold off on it since I have no idea what kinds of subtle effects this might have :/

undingen:
While I liked the django perf improvement I'm fine with not merging this change :)
If this shows up the future as as a large perf change we can reevaluate if we want to risk it.

kmod:
It looks like we have an assert in the metaserver codebase that the Python build is UCS-4 :/

PR #980, opened 2015-10-22
Misc pyxl2 perf improvements continued

undingen:
```
                                   upstream/master:      origin/pyxl2_2:
           django_template3.py             2.2s (4)             2.2s (2)  +0.7%
                 pyxl_bench.py             1.8s (4)             1.8s (2)  +0.4%
     sqlalchemy_imperative2.py             2.3s (4)             2.3s (2)  -0.2%
                pyxl_bench2.py             1.3s (4)             1.3s (2)  -1.3%
       django_template3_10x.py            14.2s (4)            14.4s (2)  +0.9%
             pyxl_bench_10x.py            14.3s (4)            14.2s (2)  -0.4%
 sqlalchemy_imperative2_10x.py            17.5s (4)            17.4s (2)  -0.5%
            pyxl_bench2_10x.py            12.2s (4)            11.5s (2)  -5.5%
                       geomean                 5.2s                 5.1s  -0.8%
```

### DIFF
--- a/src/capi/typeobject.cpp
+++ b/src/capi/typeobject.cpp
@@ -1012,24 +1012,26 @@ Box* slotTpGetattrHookInternal(Box* self, BoxedString* name, GetattrRewriteArgs*
                     return res;
                 } else if (return_convention == ReturnConvention::NO_RETURN) {
                     assert(!res);
-                } else if (return_convention == ReturnConvention::CAPI_RETURN) {
+                } else if (return_convention == ReturnConvention::CAPI_RETURN
+                           || return_convention == ReturnConvention::NOEXC_POSSIBLE) {
                     // If we get a CAPI return, we probably did a function call, and these guards
                     // will probably just make the rewrite fail:
                     if (res) {
                         rtn->addGuardNotEq(0);
                         rewrite_args->setReturn(rtn, ReturnConvention::HAS_RETURN);
                         return res;
-                    } else
-                        rtn->addGuard(0);
+                    } else {
+                        // this could set a CAPI exception and we won't clear it inside the rewrite.
+                        rewrite_args = 0;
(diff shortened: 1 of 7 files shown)


src/capi/typeobject.cpp
```diff
@@ -1022,14 +1023,13 @@ Box* slotTpGetattrHookInternal(Box* self, BoxedString* name, GetattrRewriteArgs*
                     } else
                         rtn->addGuard(0);
```
kmod: I think we might need to make this `rewrite_args = NULL` instead, since it's not clearing a potential capi exception.  I'm surprised that the change you made here helped, since I would have thought these guards would cause the rewrite to abort anyway.

undingen: It rewrites the case where we run into this case inside ```getattrInternalGeneric```: 
```c++
if (unlikely(rewrite_args && !descr && obj->cls != instancemethod_cls
         && rewrite_args->rewriter->aggressiveness() < 40
         && attr->interned_state == SSTATE_INTERNED_IMMORTAL)) {
	class Helper {
	public:
	    static Box* call(Box* obj, BoxedString* attr) { return obj->getattr(attr); }
	};

	RewriterVar* r_rtn = rewrite_args->rewriter->call(
	    false, (void*)Helper::call, rewrite_args->obj,
	    rewrite_args->rewriter->loadConst((intptr_t)attr, Location::forArg(1)));
	rewrite_args->setReturn(r_rtn, ReturnConvention::NOEXC_POSSIBLE);
	return Helper::call(obj, attr);
}
```
the call is marked as having no side effects to the additional guard don't abort the rewrite.

kmod: oh ok, right :)

PR #981, opened 2015-10-23
Don't generate guards when all arguments have fixed known types

undingen:


undingen:
I fear that this commit is introducing random test failures :-(
I have been seeing lately already some but I think this commit is increasing the number of it. (There is this thread id assert we already know since a long time but there is now also a newer on inside the GC handler of generators)

undingen:
So I restarted the travis ci several and could not see a problem with this patch but the perf improvement is now much smaller:
```
                                  upstream/master:  origin/guards_know:
           django_template3.py             2.1s (4)             2.1s (4)  +0.5%
                 pyxl_bench.py             1.8s (4)             1.8s (4)  +0.9%
     sqlalchemy_imperative2.py             2.2s (4)             2.2s (4)  +1.0%
                pyxl_bench2.py             1.1s (4)             1.0s (4)  -0.9%
       django_template3_10x.py            13.8s (4)            14.1s (4)  +1.9%
             pyxl_bench_10x.py            13.4s (4)            13.0s (4)  -3.2%
 sqlalchemy_imperative2_10x.py            16.9s (4)            16.8s (4)  -0.5%
            pyxl_bench2_10x.py             9.1s (4)             8.9s (4)  -1.8%
                       geomean                 4.7s                 4.7s  -0.3%
```

Issue #902, opened 2015-09-08
Support the "ast" module


kmod:
I think this boils down to just implementing the "_ast" module which has the actual Python types.  We have a simple module (src/runtime/builtin_modules/ast.cpp) that declares the different types, but they're currently empty.

I think any individual type is pretty straightforward, but there are 90 of them.  CPython auto-generates their module (Python/Python-ast.c) from their grammar file.  We could try to modify their grammar generator, or modify the output, or just continue to implement the module from scratch.

aisk:
I'll try this :smile_cat:

corona10:
i done by modifying some files and add 

```	
new file:   from_cpython/Include/Python-ast.h
new file:   from_cpython/Include/asdl.h
new file:   from_cpython/Python/Python-ast.c
new file:   from_cpython/Python/asdl.c
new file:   from_cpython/Python/pyarena.c

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   from_cpython/CMakeLists.txt
```
but i don't know what to next

Daetalus:
My proposal is write some types manually first. Summarize and find the pattern, them modify CPython grammar generator, or write the module from scratch. Just some thoughts, not verify it.

corona10:
@Daetalus  so it means code generator should makes src/runtime/builtin_modules/ast.cpp??

Daetalus:
Just my proposal, wait for @kmod to give some concrete suggestions

kmod:
I agree with daetalus, I don't think we should jump straight to using their generator.  I think there are a couple ways we could attack this, but the one I would suggest is:
- copy in Python-ast.c (and just that file)
- try to convert it from using CPython's AST structs (Python-ast.h) to our AST structs (src/core/ast.h)
- see what needs to be changed; if it's a small amount I would just make the changes in the generated file.  if it's a lot, then maybe take a look at the generator to see if modifying that would be easier.

PR #983, opened 2015-10-23
enable CPython's _ast module

aisk:
Just copied & modified CPython's asdl stuff to generate _ast module.

issule: https://github.com/dropbox/pyston/issues/902


kmod:
Hey @aisk are you still interested in working on this?  With the CPython parser in-tree now, I think it should be much easier to do this now.

aisk:
@kmod Hi I will continue trying this in this weekends :smile:

Daetalus:
@aisk Glad to hear, I need this module too.

aisk:
Update:

    Now just enabled the CPython's _ast module, and fixed the builtin function `compile` to use that stuff to compile source to AST and AST to runable.

kmod:
Nice!  Looks like the issue is that it's having trouble with some of the static objects defined in the Python-ast.c file -- I think the fix should be a matter of adding `PyGC_AddRoot` calls to where it assigns to the static variables (in `init_types()`).

If you could, I think it would be nice to take a look at enabling `test_ast.py` and `test_compile.py`.  I'm not sure how close those are to passing, but I think otherwise we won't have great coverage of the ast/compile() functionality.

aisk:
Oh there were other issues in `cpythonToPystonAST`, I saw some convert function is not handled in some situation?

kmod:
Yeah looks like the handling of "module"-type objects isn't great and needs to be fleshed out a little bit.  (I didn't/don't know how it works so I just had it abort.)  It looks like we don't have any test coverage of the `compile(AST)` path that hits this, so I'd recommend 1) getting some tests in (either test_ast.py, or ideally some custom tests as well), and 2) changing the `compile(str)` path to be more similar to the `compile(AST)` path.  Specifically, one path uses the CPython logic, and the other uses the Pyston logic, so they behave pretty differently, and all the testing we have of the `compile(str)` path doesn't help us with making sure that `compile(AST)` does the right thing.  I would try changing the calls to `parseExec`/`parseEval` to calls to `PyParser_ASTFromString`.

As for the specific problem, what I think is happening here is different handling of different module types.  This is metadata about the string that changes the way the string is interpreted, and is what the "eval"/"exec"/"single" strings mean.  Pyston code handles this completely at parse time by post-processing the raw AST node to reflect the metadata (this happens in parseEval/parseExec).  CPython, on the other hand, just stores this metadata as part of the returned AST object, and then does the actual transformation when compiling to bytecode.  So when we go through a code path that uses CPython's parser but our compiler, neither part will do the transformation (see below for the kind of thing that ends up getting missed).  One way forward could be to use CPython's parser, and then in cpythonToPystonAST, do the transformation (ie call `makeModuleInteractive()` or whatever else is necessary).

The transformation is some simple things like "in 'single' mode, print out all bare expressions".  Here's an example of something that isn't working in this PR:
```Python
a = compile("'hello world'", "test.py", "single", 0x400)
exec compile(a, "test.py", "single")
```
This should print `'hello world'` but currently in this PR doesn't print anything; we need to figure out a place to call `makeModuleInteractive`.

Sorry this is messy!  As we switch closer to the way CPython handles it, things will become more straightforward.

aisk:
Hi @kmod , sorry since the `Pyston AST`, `CPython native AST` and `Python AST` are complicated to me, I think I'm understood the exactly problem was:

In the old parse process in builtin `compile` function, we trust all modes as `exec` (compile as module), and the parsed AST is marked as `Module_Kind`, then we do the AST transform. This works good without using `_ast.PyCF_ONLY_AST` param.

But with it, we must compile the AST which follows the mode param. If we use the `eval` mode, which compiles the `Expression_Kind`, then when we `compile(AST)`, the `cpythonToPystonAST ` function can't handle the `Expression_Kind`.

So the next thing to do is:

* complete the `cpythonToPystonAST` function to let it handle `Expression_Kind`
* refactor the code, remove the parseXXX function call in `compile`, just parse the code to AST with the mode in param, then do the AST transform in `cpythonToPystonAST `

Is this right, or some misunderstanding I am making?

kmod:
Sounds good!  I think we're coming at this from slightly different perspectives but I think we're getting to the same place.  Concretely, I would recommend handling `Expression_Kind` by calling `makeModuleInteractive()`.

from_cpython/Python/Python-ast.c
```diff
@@ -827,124 +827,124 @@ static int init_types(void)
...
-        Or_singleton = PyType_GenericNew(Or_type, NULL, NULL);
+        Or_singleton = PyGC_AddRoot(PyType_GenericNew(Or_type, NULL, NULL));
         if (!Or_singleton) return 0;
         operator_type = make_type("operator", &AST_type, NULL, 0);
         if (!operator_type) return 0;
         if (!add_attributes(operator_type, NULL, 0)) return 0;
         Add_type = make_type("Add", operator_type, NULL, 0);
         if (!Add_type) return 0;
-        Add_singleton = PyType_GenericNew(Add_type, NULL, NULL);
+        Add_singleton = PyGC_AddRoot(PyGC_AddRoot(PyType_GenericNew(Add_type, NULL, NULL)));
```
undingen: ```PyGC_AddRoot(PyGC_AddRoot( -> PyGC_AddRoot(```

src/runtime/capi.cpp
```diff
@@ -1231,8 +1231,9 @@ extern "C" int PyRun_InteractiveOneFlags(FILE* fp, const char* filename, PyCompi
     assert(PyModule_Check(m));
     bool failed = false;
     try {
-        AST_Module* pyston_module = cpythonToPystonAST(mod, filename);
-        makeModuleInteractive(pyston_module);
+        assert(mod->kind != Expression_kind);
```
undingen: I think this should always be of ```Interactive_kind``` could you please change the assert

src/codegen/cpython_ast.cpp
```diff
@@ -651,23 +651,43 @@ class Converter {
...
+                this->pool = rtn->interned_strings.get();
+                rtn->body = this->convert(mod->v.Expression.body);
                 return rtn;
             }
             default:
                 RELEASE_ASSERT(0, "unhandled kind: %d\n", mod->kind);
         }
     }
+
+    AST_Expression* convertExpression(mod_ty mod) {
```
undingen: this function can be removed

undingen:
I'm not really sure what to think about this added ```PyGC_AddRoot``` because if they are required I think that ```make_type``` should also call ```PyGC_AddRoot``` on the return value.

Looks good on first sight! I will test it out later but I noticed completely lack of testing. Could you please add some tests. Thanks!

aisk:
@undingen Thanks for review, I'll add more testing now!

aisk:
Oh I'm not familiar with the GC, but before `PyGC_AddRoot` there were some exception happened, so I think it's needed. And I'll have a try to see the GC details. 😀

undingen:
I think we should call ```PyGC_AddRoot``` inside ```make_type```.

The ```'continue' not properly in loop``` etc error checking is as you properly noticed quite a mess currently :-(. How do you feel about not fixing this in the PR but doing it in another PR later on (you don't have to do it - I can take a look at it) and instead just disabling the failing test cases in this PR?
Because we not only have to move the error checking form the pypa parsing step and cpython AST to our AST converting step into the computeCFG we also have to make sure that computeCFG get's called at the correct time and not lazily like we do now and I think this is easier to keep track of if it's done in a separate PR.

aisk:
@undingen sorry, I'll update this PR as soon as I can. And should I keep the syntax check in cpythonToPystonAST as this PR now, or just reset it back?

undingen:
I think it's best to keep the syntax checks otherwise this will break existing tests. Disabling the failing parts of the new cpython AST test is fine (in the test/tests/ copy). I had to do something similar in #1045

src/codegen/cpython_ast.cpp
```diff
@@ -633,8 +633,9 @@ class Converter {
                     raiseSyntaxError("'break' outside loop", stmt->lineno, stmt->col_offset, fn, "", true);
                 return new AST_Break();
             case Continue_kind:
-                if (loop_depth == 0)
```
undingen: I think you meant to do a ```if (loop_depth == 0)``` error check not two ```in_finally``` checks

test/tests/try_continue.py
```diff
@@ -1,3 +1,4 @@
+# skip-if: True
```
undingen: You can enable the the test again now that we slit the PRs

aisk: sorry I have pasted the wrong code here 😅

undingen:
LGTM,
I noticed that this will now always use cpythons parser for parsing strings inside ```compile()```. Because we plan to switch away from pypa to the cpython parser this should be fine and doing additional work to support something we will remove soon is probably not worth it.

Thanks for the patch!

aisk:
😀

PR #984, opened 2015-10-26
Support building distribution packages

kmod:
Right now I've only set it up to build a tarball.  It's still kind of a wip since I haven't tested everything yet, and the name is whatever the default cpack name is ("0.1.1" I think).  will fix those in a sec but want to get this tested and in.


kmod:
I improved it a bit past the initial description: `make package` should be ready to go :)

PR #986, opened 2015-10-27
add doc to existed builtin function or method

Daetalus:
Just for existed builtin functions or methods. The method under a class is `instancemethod` in Pyston, CPython is `wrapper_descriptor`. But our `BoxedWrapperDescriptor` seems used for CPython method only. Maybe we need to let it support `CLFunction` too? Or use other strategy?


undingen:
lgtm
Thank you very much for the patch!

PR #987, opened 2015-10-27
These cpython tests are passing now

kmod:


undingen:
Should we merge this in so it's included in the 0.4 release?

PR #992, opened 2015-11-02
Improve comments, rename CLFunction

kmod:
Lots of changes due to the rename, but it's just renaming things and then formatting.


kmod:
Love the reduced amount of test flakes :)

This test did timeout though, will try to address that in an upcoming PR.

PR #993, opened 2015-11-03
Use configurable path to cpython

Dundee:
Use configurable path to cpython to be able to build pyston on machines with default Python3.


kmod:
Good fix, thanks :)

PR #996, opened 2015-11-04
build jemalloc

Dundee:
If not build, error is raised:

ninja: error: 'jemalloc/lib/libjemalloc.a', needed by 'pyston', missing and no known rule to make it


kmod:
Oh interesting, what version of cmake do you have (`cmake --version`)?  We have some hacks for jemalloc to workaround some bugs in cmake, but they might interact poorly with some versions of cmake.

Dundee:
I have cmake 3.3.2

kmod:
I think #1003 should fix this, would you mind pulling and trying again?

Dundee:
Yes, it works now. Thanks!

PR #1005, opened 2015-11-08
Add two attribute to str that let test_string pass.

Daetalus:
please ignore the branch name, I made a mistake about it...

### DIFF
--- a/from_cpython/Objects/stringobject.c
+++ b/from_cpython/Objects/stringobject.c
@@ -23,6 +23,22 @@ PyObject * _do_string_format(PyObject *self, PyObject *args, PyObject *kwargs) {
     return do_string_format(self, args, kwargs);
 }
 
+void _string_init() {
+    if (PyType_Ready(&PyFieldNameIter_Type) < 0)
+        Py_FatalError("Can't initialize field name iterator type");
+
+    if (PyType_Ready(&PyFormatterIter_Type) < 0)
+        Py_FatalError("Can't initialize formatter iter type");
+}
+
+PyObject * _formatter_parser(STRINGLIB_OBJECT *self) {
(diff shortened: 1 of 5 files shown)


from_cpython/Objects/stringobject.c
```diff
@@ -23,6 +23,22 @@ PyObject * _do_string_format(PyObject *self, PyObject *args, PyObject *kwargs) {
     return do_string_format(self, args, kwargs);
 }
 
+void _string_init() {
```
undingen: Do we have to to do this?
Looks like cpython does it in ```_PyUnicode_Init``` and we use this one do so I think we can remove it.

undingen:
LGTM with the removal of ```_string_init``` :-).

from_cpython/Objects/stringobject.c
Daetalus: Hi, @undingen our str implementation couldn't use `_PyUnicode_Init`(But I don't know the reason). @kmod suggest me add a initialization function and call `PyType_Ready` in it. FYI, the related discussion in Septemper 5th in Gitter. Maybe there has another way to do it?

undingen: I'm suggesting to just remove the ```_string_init``` func. Because it's getting executed after ```_PyUnicode_Init``` which will already have register this types.

undingen: Oh sorry I didn't try it out :-( - now I understand why you added it.

undingen:
thanks for the patch and sorry for my confusion (I did not see that this types are defined as static inside a header...)

Daetalus:
Ah, thanks for tell me the reason. I didn't noticed it before.

PR #1006, opened 2015-11-10
add float_repr_style to sys module and enable test_strtod

Daetalus:


undingen:
Can we switch to cpythons ```_Py_double_round``` implementation in ```floatobject.c```?
If not could you please make sure that we get a compile error inside our ```_Py_double_round()``` in ```float.cpp``` if ```PY_NO_SHORT_FLOAT_REPR``` is defined.
Because otherwise I fear that if on some hardware this define is set we would secretly still use the "short" implementation even though it says otherwise...

otherwise LGTM :-) and thanks for getting all this tests working!

Daetalus:
Updated, ready to review, thanks for your feedback!

undingen:
Why do we have to copy the implementation into ```float.cpp```, can't we just directly use the one in ```floatobject.c```?

undingen:
:+1:

PR #1009, opened 2015-11-13
Test file for paste (and py.test) modules

rudi-c:


kmod:
Looks good if we remove the libunwind change.

rudi-c:
I got too used to using `git add -u` again ^^'

kmod:
Awesome, thanks :)  We should probably change our libunwind-patching strategy at some point -- if we want to stick with manually applied patches, I think we should do that in the build directory instead of the source directory.

PR #1010, opened 2015-11-14
Fix build error caused by missing 'override's.

AlphaStream:
Got a couple build errors when building pyston in my clean Ubuntu 15.10 installation.
Adding the missing 'override's fixes them.


kmod:
Awesome, thanks for the fix :)

PR #1014, opened 2015-11-20
Work on running and passing NumPy tests

rudi-c:
It doesn't run the whole test suite without crashing yet, but it makes a lot of progress.

For the most part these changes don't cause a performance hit, except for `Assign allocation function directly to tp_new`, on the order of about 1%. Any suggestions how to optimize?


kmod:
Ok, left some comments :)

I'd recommend separating the `Assign allocation function directly to tp_new` commit from this PR to not block the other commits on that one.  My guess as to what's happening there is the custom tp_new is causing `typeCallInner` to pick its "Looks like we're calling an extension class" strategy which is not quite as fast; it'd be interesting to try adding an exception to that case so that it doesn't trigger if `cls == int_cls` or the other ones you added.

Daetalus:
I think the `tp_new` issue is one of the most important part of this PR. I will look at the `typeCallInner` code, try to understand your last reply.

kmod:
Responding to your comment about the error checking, we should be checking for a return value of -1 from array_shape_set(), which should propagate the error.  If you wouldn't mind, it'd be nice to step out of array_shape_set(), make sure that it returns -1, and if it does see why the return value doesn't get checked.

This kind of thing is the reason I like having these `assert(!PyErr_Occurred())` checks everywhere -- the farther we catch the issue from the cause, the harder it is to track it down, and the more checks we have the faster we can catch it.  I think the issue here is not actually the assertion in callCLFunc, but rather that it's a different bug somewhere else that the assertion is catching.  In theory we might run into cases where we actually need to support people who violate the C API and forget to return NULL, but for now it seems like most times that happens is our bug.

Similarly when the assertion fires before the actual call to the function, that means that we let an exception go unhandled somewhere else.  I'd say that in that case the bug is that we didn't catch the exception when it was thrown, so we might have to hunt down where that happened.

kmod:
Oh, about the array_shape_set case, I think what should be happening is:
- the rewriter calls `rewrite->checkAndThrowCAPIException()`
- checkAndThrowCAPI emits `if (rtn_val == -1) throwCapiException()`
- throwCapiException() should clear the exception state as part of its operation, before it throws the C++ exception

Hopefully by stepping through the end of array_shape_set, it shouldn't be too hard to see which of these steps isn't happening.

rudi-c:
I'm closing this PR now since it got split into the other ones.

PR #1019, opened 2015-11-30
Fix #1016 by backporting llvm r236316

### LINKED ISSUE #1016: Build errors on Clang 3.7 because of new move diagnostics
Trying to compile with Clang 3.7 results in the following errors because of `-Werror` and the new [move warnings](http://llvm.org/releases/3.7.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics) which are included with `-Wall` in the 3.7 release.

```
~/pyston_deps/llvm-trunk/include/llvm/ADT/APFloat.h:352:12: 
error: redundant move in return statement [-Werror,-Wredundant-move]
    return std::move(Value);
           ^
```

```
../../src/codegen/compvars.cpp:2548:16:
error: moving a local object in a return statement prevents copy elision [-Werror,-Wpessimizing-move]
        return std::move(elts);
               ^
```
Note that the first error already has a [fix in LLVM](https://github.com/llvm-mirror/llvm/commit/fda9fd463dae38840cc02c5ae668cb1097395b9a).

undingen:
I don't have clang 3.7 so I could not test it but I don't see why it should not work

### DIFF
--- a/llvm_patches/0012-Remove-move-warning.patch
+++ b/llvm_patches/0012-Remove-move-warning.patch
@@ -1,0 +1,65 @@
+From fda9fd463dae38840cc02c5ae668cb1097395b9a Mon Sep 17 00:00:00 2001
+From: Benjamin Kramer <benny.kra@googlemail.com>
+Date: Fri, 1 May 2015 15:16:11 +0000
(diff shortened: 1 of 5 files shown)


kmod:
Merged as part of #1020

PR #1020, opened 2015-12-02
clang-3.7 support

kmod:
This is a continuation of #1019 and [dean0x7d/clang37_fixes](https://github.com/dean0x7d/pyston/commits/clang37_fixes).


dean0x7d:
Sure, but the comarison was `< (1L << 31) - 1` not `<=  (1L << 31) - 1`.

undingen:
Oh sorry you are of course correct. I think I should grab some coffee... :blush:

kmod:
Hmm it seems like the original comparisons were wrong, I'm going to go ahead and change it.

kmod:
Hey Marius can you look at the new commit I added?  I think it's good but if we messed it up once we should check twice before changing it :)

undingen:
lgtm, I checked with the instruction manual I could not see why we can't include the MAX_INT.

PR #1021, opened 2015-12-02
Don't generate patchpoints for calls to fixed functions which don't need patching

undingen:
This is based on #1011, only the last 4 commits are different
perf difference to #1011:
```
           django_template3.py             2.2s (2)             2.1s (2)  -1.8%
                 pyxl_bench.py             1.7s (2)             1.8s (2)  +1.7%
     sqlalchemy_imperative2.py             2.2s (2)             2.2s (2)  -0.8%
                pyxl_bench2.py             1.1s (2)             1.1s (2)  -3.3%
       django_template3_10x.py            14.0s (2)            14.0s (2)  -0.2%
             pyxl_bench_10x.py            13.7s (2)            13.4s (2)  -1.6%
 sqlalchemy_imperative2_10x.py            17.0s (2)            16.9s (2)  -0.7%
            pyxl_bench2_10x.py             9.4s (2)             8.9s (2)  -4.6%
                       geomean                 4.8s                 4.7s  -1.4%
```

Previously we generated patchpoints for every call because the callee could do a frame introspection
and we could not retrieve the frame info without the additional patchpoint informations.

This change moves all required information inside the FrameInfo.
Because the FrameInfo does not have a fixed offset (different num of stack args and OSR functions reuse the frame info from the interpreter),
we generate at function entry a stack variable which points to the frame_info and than we emit a stackmap intrinsic
in order to know which basepointer relative address it has.
The current statement inside the FrameInfo gets only updated on direct calls - patchpoints still attach it directly instead of updating the memory.
Because always updating the variable is a small slowdown.

This allows us todo inlining of trivial call sites at the llvm IR level.

To improve the generated code I switched to the default regalloc.


undingen:
Ok I fixed the errors - should now be ready for review but better start with #1011

kmod:
sgtm :)

PR #1023, opened 2015-12-03
Add cmath module and fix some bugs discovered in test_cmath.

Daetalus:
Add `complexobject.cpp` file, remove some fundamental and duplicated functions in `complex.cpp`. Some functions still leave in `complex.cpp`, because our complex object(`BoxedComplex`) is different from CPython's. So I think leave it there will be fine.

### DIFF
--- a/from_cpython/CMakeLists.txt
+++ b/from_cpython/CMakeLists.txt
@@ -36,6 +36,7 @@ file(GLOB_RECURSE STDMODULE_SRCS Modules
     bufferedio.c
     bytesio.c
     cache.c
+    cmathmodule.c
     connection.c
     cStringIO.c
     cursor.c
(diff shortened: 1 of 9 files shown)


kmod:
Looks good!  Looks like we need to change the way that it's finding its test data though -- it's trying to pull them from the build directory.  Something like this should be able to make it pull them from the source directory (test_math.py does this but test_cmath doesn't):
```python
if __name__ == "__main__":
    test_file = os.path.join(os.path.dirname(sys.argv[0]), "cmath_testcases.txt"))
```

kmod:
Also, complex seems like a good candidate for switching to cpython's implementation, though if we don't have any pressing need to do that it's also fine to leave it split like this.

kmod:
Awesome!

PR #1027, opened 2015-12-10
Work on running and passing NumPy tests (2)

rudi-c:
Follow-up from https://github.com/dropbox/pyston/pull/1014. I'm making a separate PR so the comments don't disappear from rebasing.

Changes:
1) Throw exception in getset. Note that CPythons' exception wants to know the name of the descriptor so I added that as a field, and needed get change every creation of a getset descriptor.
2) Took out the tp_new stuff from this PR.
3) Removed the loosening of the exception state assertion, changed the NumPy test instead for now.


undingen:
LGTM, thanks for the patch!
I fixed the lxml error in #1057.

PR #1028, opened 2015-12-10
Fixes for tp_new for C extensions

rudi-c:
Split from https://github.com/dropbox/pyston/pull/1014

In the second commit, I add a special case in `typeCallInner` for `int_cls`, since it's the most common one to show up. It helps performance a bit.
```
# without special case
Comparing to '['baseline']'
              pyston (calibration)                      :   1.01s baseline: 1.02 ( -0.6%)
              pyston django_template3.py                :   1.98s baseline: 1.97 (  0.2%)
              pyston pyxl_bench.py                      :   1.63s baseline: 1.62 (  0.5%)
              pyston sqlalchemy_imperative2.py          :   2.12s baseline: 2.11 (  0.3%)
              pyston django_template3_10x.py            :  11.72s baseline: 11.53 (  1.6%)
              pyston pyxl_bench_10x.py                  :  12.84s baseline: 12.78 (  0.5%)
              pyston sqlalchemy_imperative2_10x.py      :  16.31s baseline: 16.13 (  1.1%)
              pyston (geomean-2ec9)                     :   1.90s baseline: 1.89 (  0.3%)
# with special case
Comparing to '['baseline']'
              pyston (calibration)                      :   1.02s baseline: 1.02 (  0.6%)
              pyston django_template3.py                :   1.99s baseline: 1.97 (  0.7%)
              pyston pyxl_bench.py                      :   1.63s baseline: 1.62 (  0.7%)
              pyston sqlalchemy_imperative2.py          :   2.11s baseline: 2.11 ( -0.1%)
              pyston django_template3_10x.py            :  11.52s baseline: 11.53 ( -0.0%)
              pyston pyxl_bench_10x.py                  :  12.78s baseline: 12.78 (  0.0%)
              pyston sqlalchemy_imperative2_10x.py      :  16.15s baseline: 16.13 (  0.1%)
              pyston (geomean-2ec9)                     :   1.90s baseline: 1.89 (  0.4%)
```

But I don't know how I feel about it, it seems quite hacky for a relatively small performance drop.

### DIFF
--- a/src/runtime/float.cpp
+++ b/src/runtime/float.cpp
@@ -926,6 +926,22 @@ template <ExceptionStyle S> Box* floatNew(BoxedClass* _cls, Box* a) noexcept(S =
     return new (cls) BoxedFloat(f->d);
 }
 
+// Roughly analogous to CPython's float_new.
+// The arguments need to be unpacked from args and kwds.
+static Box* floatNewPacked(BoxedClass* type, Box* args, Box* kwds) noexcept {
+    PyObject* x = False;
+    static char* kwlist[2] = { NULL, NULL };
+    kwlist[0] = const_cast<char*>("x");
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O:float", kwlist, &x))
+        return NULL;
+
+    if (x == NULL)
+        return floatNew<CAPI>(type, None);
+    else
+        return floatNew<CAPI>(type, x);
+}
+
 PyObject* float_str_or_repr(double v, int precision, char format_code) {
     PyObject* result;
     char* buf = PyOS_double_to_string(v, format_code, precision, Py_DTSF_ADD_DOT_0, NULL);
@@ -1699,6 +1715,7 @@ void setupFloat() {
 
     float_cls->tp_str = float_str;
     float_cls->tp_as_number->nb_power = float_pow;
+    float_cls->tp_new = (newfunc)floatNewPacked;
 }
 
 void teardownFloat() {
--- a/src/runtime/str.cpp
+++ b/src/runtime/str.cpp
@@ -1632,11 +1632,11 @@ extern "C" Box* strNonzero(BoxedString* self) {
     return boxBool(self->size() != 0);
 }
 
-extern "C" Box* strNew(BoxedClass* cls, Box* obj) {
+template <ExceptionStyle S> Box* strNew(BoxedClass* cls, Box* obj) noexcept(S == CAPI) {
     assert(isSubclass(cls, str_cls));
 
     if (cls != str_cls) {
-        Box* tmp = strNew(str_cls, obj);
+        Box* tmp = strNew<S>(str_cls, obj);
         assert(PyString_Check(tmp));
         BoxedString* tmp_s = static_cast<BoxedString*>(tmp);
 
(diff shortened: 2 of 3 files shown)


kmod:
I think if we do this right, there shouldn't be any perf change, since in theory we should be able to get it to the point that these new functions don't get called at all in the benchmarks.

One way to do this would be to have a flag that says "hey even though I have tp_new set, I also have a Python `__new__` available so you should try calling that".  We already have two classes that want this flag (all the exception types, and int), but there are potentially more.  I think the 'is_pyston_class' flag is actually already suitable for this, though maybe a bit hacky.

Another approach is to switch this to the "rewriter aggressiveness backoff" system which is designed for this sort of thing: when we have multiple options for how to try to execute things, but we don't know how many of the fastpaths we'll be able to take.  I think this makes sense, but it's probably overkill and we can keep the change much smaller by checking is_pyston_class.

If you have time, it'd be nice to test putting a stat counter inside that special case, and verify that it gets triggered the same number of times before and after the PR.

src/runtime/str.cpp
```diff
@@ -1632,11 +1632,11 @@ extern "C" Box* strNonzero(BoxedString* self) {
     return boxBool(self->size() != 0);
 }
 
-extern "C" Box* strNew(BoxedClass* cls, Box* obj) {
+template <ExceptionStyle S> Box* strNew(BoxedClass* cls, Box* obj) {
```
kmod: Minor, but can you add a `noexcept(S == CAPI)` right before the opening brace?  This will mean that the CAPI version will get a "noexcept" qualifier and is just an extra protection against unhandled exceptions.

rudi-c:
I added the noexcept clause.

The performance drop that I initially observed seem to be gone now and the special case in `typeCallInner` hack no longer does anything, so I removed it.

```
Comparing to '['baseline']'
              pyston (calibration)                      :   1.02s baseline: 1.02 ( -0.4%)
              pyston django_template3.py                :   2.08s baseline: 2.08 ( -0.1%)
              pyston pyxl_bench.py                      :   1.80s baseline: 1.82 ( -1.1%)
              pyston sqlalchemy_imperative2.py          :   2.16s baseline: 2.16 ( -0.1%)
              pyston django_template3_10x.py            :  12.33s baseline: 12.27 (  0.5%)
              pyston pyxl_bench_10x.py                  :  13.87s baseline: 13.82 (  0.4%)
              pyston sqlalchemy_imperative2_10x.py      :  16.60s baseline: 16.67 ( -0.4%)
              pyston (geomean-2ec9)                     :   2.01s baseline: 2.02 ( -0.4%)
```

The new case does get called in the benchmarks, setting `is_pyston_class` doesn't change that, but it doesn't improve performance either (it actually makes it worse). But that does lead me to wonder, was there any reason not to set `is_pyston_class` to all of our classes? Performance?

src/runtime/float.cpp
```diff
@@ -926,6 +926,22 @@ template <ExceptionStyle S> Box* floatNew(BoxedClass* _cls, Box* a) noexcept(S =
...
+static Box* floatNewPacked(BoxedClass* type, Box* args, Box* kwds) noexcept {
+    PyObject* x = False;
+    static char* kwlist[2] = { NULL, NULL };
+    kwlist[0] = const_cast<char*>("x");
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O:float", kwlist, &x))
+        return NULL;
+
+    if (x == NULL)
+        return floatNew<CAPI>(type, None);
```
undingen: I think the arg should be ```boxFloat(0.0)``` not ```None```.
But than I think the ```NULL``` check will never trigger because ```x``` is initialized to ```False``` and marked as optional inside ```PyArg_ParseTupleAndKeywords``` .
I think we could just get rid of the if check in cpython does and just call all the time into ```floatNew<CAPI>(type, x); ```  and similar to cpython make use of the fact that if ```x``` is ```False``` and ```floatNew``` will create a ```0.0``` float for it.

undingen:
Hi, sorry for the very late review!
I added a minor comment but decided I don't want to introduce more overhead and even more delays to the whole process and merged it in and created #1115 to fix it. @Daetalus let me know that this fixes some numpy crashes for example ```test_abc.py``` :-) 
Thanks for the patch!

undingen:
Oh I forgot to mention: I verified that this does not decrease the performance even though this will cause more tp_new calls...

PR #1029, opened 2015-12-10
Implement PyErr_WarnEx

### LINKED ISSUE #1013: Support PyErr_WarnEx
As of now, we throw a fatal error on warnings (e.g. function deprecated warnings) since `PyErr_WarnEx` is not implemented. The function should just print a error message.

This should be a fairly straightforward copy of CPython's code.

NumPy needs this.

Daetalus:
Implement `PyErr_WarnEx`, based on CPython implementation. Except use Pyston BoxedFrame instead PyFrameObject. And also apply some pyston changes.

closes #1013

### DIFF
--- a/from_cpython/Python/_warnings.c
+++ b/from_cpython/Python/_warnings.c
@@ -446,16 +446,26 @@ setup_context(Py_ssize_t stack_level, PyObject **filename, int *lineno,
     PyObject *globals;
 
     /* Setup globals and lineno. */
-    PyFrameObject *f = PyThreadState_GET()->frame;
-    while (--stack_level > 0 && f != NULL)
-        f = f->f_back;
-
-    if (f == NULL) {
-        globals = PyThreadState_Get()->interp->sysdict;
-    }
-    else {
-        globals = f->f_globals;
-        *lineno = PyFrame_GetLineNumber(f);
+    // Pyston changes: use Pyston's BoxedFrame instead PyFrameObject
+    // to get globals and lineno.
+    /* PyFrameObject *f = PyThreadState_GET()->frame; */
+    /* while (--stack_level > 0 && f != NULL) */
+    /*     f = f->f_back; */
+    /*  */
+    /* if (f == NULL) { */
+    /*     globals = PyThreadState_Get()->interp->sysdict; */
+    /* } */
+    /* else { */
+    /*     globals = f->f_globals; */
+    /*     *lineno = PyFrame_GetLineNumber(f); */
+    /* } */
+    PyFrameObject* frame = PyFrame_ForStackLevel(stack_level - 1);
+    if (!frame) {
+        globals = PySys_GetModulesDict();
+        *lineno = 1;
+    } else {
+        globals = PyFrame_GetGlobals(frame);
+        *lineno = PyFrame_GetLineNumber(frame);
     }
 
     *module = NULL;
@@ -890,6 +900,8 @@ _PyWarnings_Init(void)
     _filters = init_filters();
     if (_filters == NULL)
         return;
+    // Pyston change: let the GC scan the filters
+    PyGC_AddPotentialRoot(_filters, sizeof(_filters));
     Py_INCREF(_filters);
     if (PyModule_AddObject(m, "filters", _filters) < 0)
         return;
@@ -897,6 +909,8 @@ _PyWarnings_Init(void)
     _once_registry = PyDict_New();
     if (_once_registry == NULL)
         return;
+    // Pyston change: let the GC scan the registry
+    PyGC_AddPotentialRoot(_once_registry, sizeof(_once_registry));
     Py_INCREF(_once_registry);
     if (PyModule_AddObject(m, "once_registry", _once_registry) < 0)
         return;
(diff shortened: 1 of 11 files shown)


src/runtime/capi.cpp
```diff
@@ -1034,14 +1053,145 @@ extern "C" PyObject* PyErr_Occurred() noexcept {
...
+
 extern "C" int PyErr_WarnEx(PyObject* category, const char* text, Py_ssize_t stacklevel) noexcept {
     // These warnings are silenced by default:
     // We should copy the real CPython code in here
     if (category == PyExc_DeprecationWarning)
         return 0;
 
-    fatalOrError(PyExc_NotImplementedError, "unimplemented");
-    return -1;
+    PyObject* res;
```
kmod: Is it possible to get rid of the first part of this function now?  (It's a workaround that I added, not something from CPython.)

kmod:
Looks good!  This isn't critical, but I'm wondering if it might clean things up if we leave setup_context in from_cpython.  ie it will make it easier to see what we changed, and won't require copying over those function declarations.  It looks like the main change is in the beginning of the function where it deals with the BoxedFrame; maybe another approach is to add a `PyFrame_ForStackLevel()` API function?  And then you can use the `PyFrame_GetGlobals()` function on that.

Daetalus:
> maybe another approach is to add a `PyFrame_ForStackLevel()` API function? And then you can use the `PyFrame_GetGlobals()` function on that.

@kmod sorry, I don't quite understand it. You mean add `PyFrame_ForStackLevel()` API to get `PyFrameObject` ? Pyston not use it, and seems it just like `getFrame`. And I don't know how to keep `setup_context` stay in from_cpython without introduct `BoxedFrame` to it.

undingen:
I think he means that you could leave the ```setup_context``` function inside the ```_warnings.c``` file if you replace 
```
PyFrameObject *f = PyThreadState_GET()->frame;
    while (--stack_level > 0 && f != NULL)
        f = f->f_back;
```
with a custom API extension ```PyFrame_ForStackLevel()``` which just calls ```getFrame``` internally (but you can't directly call ```getFrame``` from the C file).

Daetalus:
@undingen But seems we are not using `PyFrameObject`?

undingen:
Are there any other places inside ```_warnings.c``` where you would have to use ```BoxedFrame*```?
The idea is that while you can't use ```BoxedFrame*``` (only ```PyFrameObject*``` and that just for passing it around as a pointer / can't instect any fields) inside the C file but if we only access fields in say two places and they are easily fixable by extending the C API and in addition this extensions functions look generally useful (e.g. for porting cython/numpy etc) than we can just add those functions.

kmod:
Yes, sorry I wasn't clear :)  What I meant was that we don't have to use the `BoxedFrame` type at all since we just pass that object off to another API function.  So instead of 
```C++
    BoxedFrame* frame = (BoxedFrame*)getFrame(stack_level);
    if (!frame) {
        globals = getSysModulesDict();
        *lineno = 1;
    } else {
        globals = PyFrame_GetGlobals(frame);
        *lineno = PyFrame_GetLineNumber(frame);
    }
```
we could do
```C++
    Box* frame = getFrame(stack_level);
    if (!frame) {
        globals = getSysModulesDict();
        *lineno = 1;
    } else {
        globals = PyFrame_GetGlobals(frame);
        *lineno = PyFrame_GetLineNumber(frame);
    }
```
and this _almost_ works in C API code, except for the fact that `getFrame` isn't available, so we could add a `PyFrame_ForStackLevel()` function (or rename `getFrame` to that).  If we are hiding all the attributes behind API calls, we can just leave the object itself as an opaque type.

from_cpython/Python/_warnings.c
```diff
@@ -0,0 +1,916 @@
...
+    } else {
+        globals = PyFrame_GetGlobals(frame);
+        *lineno = PyFrame_GetLineNumber(frame);
+    }
+
+    *module = NULL;
+
+    /* Setup registry. */
+    assert(globals != NULL);
+    /* assert(PyDict_Check(globals)); */
```
Daetalus: Why this assert failed?

from_cpython/Python/_warnings.c
```diff
@@ -0,0 +1,916 @@
...
+                                registry, NULL);
+}
+
+
+/* Function to issue a warning message; may raise an exception. */
+int
+PyErr_WarnEx(PyObject *category, const char *text, Py_ssize_t stack_level)
+{
+    if (category == PyExc_DeprecationWarning)
+        return 0;
```
Daetalus: Sorry, remove this will cause error. The CPython code could not handle it.

undingen: I think you have to call ```_PyWarnings_Init``` and there are some static Python variables inside this file I'm not sure if all are GC visible - you may have to call ```PyGC_AddRoot``` when assigning on some of them.

undingen: And sorry didn't mention that this is just a guess I did not actually try it out..

kmod: Hmm, so enabling deprecation warnings isn't that important to us right now, but if that type of warnings crash(especially for the reasons that Marius brought up) then I'm worried about other kinds of warnings too.  What ended up failing when you took these lines out?

undingen:
I added a few very minor comments otherwise looks good.
Can we test this somehow?

undingen:
thanks for the change!

PR #1030, opened 2015-12-14
Add some missing attributes to Int object

Daetalus:
Some NumPy bugs was difficult to debug. For example, one of those bugs just due to Pyston's `int` has no `__long__`. Others are similar, so I decide to let Pyston's numeric types have fully support first.

Rewrite some `int` methods implementation, and add some missing attributes.

### DIFF
--- a/src/runtime/float.cpp
+++ b/src/runtime/float.cpp
@@ -27,11 +27,14 @@
 #include "runtime/types.h"
 #include "runtime/util.h"
 
+extern "C" PyObject* float_divmod(PyObject* v, PyObject* w) noexcept;
+extern "C" int float_coerce(PyObject** v, PyObject** w) noexcept;
 extern "C" PyObject* float_hex(PyObject* v) noexcept;
 extern "C" PyObject* float_fromhex(PyObject* cls, PyObject* arg) noexcept;
 extern "C" PyObject* float_as_integer_ratio(PyObject* v, PyObject* unused) noexcept;
 extern "C" PyObject* float_is_integer(PyObject* v) noexcept;
 extern "C" PyObject* float__format__(PyObject* v) noexcept;
+extern "C" PyObject* float_setformat(PyTypeObject* v, PyObject* args) noexcept;
 extern "C" PyObject* float_pow(PyObject* v, PyObject* w, PyObject* z) noexcept;
 extern "C" PyObject* float_str(PyObject* v) noexcept;
 extern "C" int float_pow_unboxed(double iv, double iw, double* res) noexcept;
@@ -213,6 +216,28 @@ extern "C" Box* floatRDiv(BoxedFloat* lhs, Box* rhs) {
         if (rhs_f == -1.0 && PyErr_Occurred()) {
             throwCAPIException();
         }
+        raiseDivZeroExcIfZero(lhs->d);
+        return boxFloat(rhs_f / lhs->d);
+    } else {
+        return NotImplemented;
+    }
+}
+
+Box* floatRTruediv(BoxedFloat* lhs, Box* rhs) {
+    if (!PyFloat_Check(lhs))
+        raiseExcHelper(TypeError, "descriptor '__rtruediv__' requires a 'float' object but received a '%s'",
+                       getTypeName(lhs));
+
+    if (PyInt_Check(rhs)) {
+        return floatRDivInt(lhs, static_cast<BoxedInt*>(rhs));
+    } else if (PyFloat_Check(rhs)) {
+        return floatRDivFloat(lhs, static_cast<BoxedFloat*>(rhs));
+    } else if (PyLong_Check(rhs)) {
+        double rhs_f = PyLong_AsDouble(rhs);
+        if (rhs_f == -1.0 && PyErr_Occurred()) {
+            throwCAPIException();
+        }
+        raiseDivZeroExcIfZero(lhs->d);
         return boxFloat(rhs_f / lhs->d);
     } else {
         return NotImplemented;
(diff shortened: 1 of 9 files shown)


src/runtime/float.cpp
```diff
@@ -111,6 +111,22 @@ extern "C" double floordiv_float_float(double lhs, double rhs) {
     return floor(lhs / rhs);
 }
 
+Box* convertToFloat(Box* val) {
```
Daetalus: I think this could be better than CPython's `CONVERT_TO_DOUBLE` implementation. Such as for exception handle in runtime source code.

src/runtime/int.cpp
```diff
@@ -1157,30 +1212,64 @@ void setupInt() {
                                                                                    ParamNames::empty(), CAPI)));
 
     _addFuncIntFloatUnknown("__add__", (void*)intAddInt, (void*)intAddFloat, (void*)intAdd);
```
Daetalus: If I rewrite the `__add__` implementation like `__div__`. The test suite will report `binops_ics` will report `exception: ("noninit_count('slowpath_binop') <= 10", 'slowpath_binop', 1005)` error.

kmod: Looks like the issue is that int no longer would know how to handle `int.__add__(float)`, and we don't (yet) rewrite cases where we have to fall back to trying `__radd__`.

kmod:
I agree with what marius said on gitter, that using-CPython's-implementation-or-our-own is kind of a case-by-case kind of thing and it depends on how much there is to change or not.  I think a lot of the motivation behind switching towards CPython's implementation comes from either missing corner cases that their implementation covers, or from a maintenance burden from having a split implementation.  You're the one who would probably feel these the most, so I think it's up to your judgement about whether there's a compelling reason to switch closer to CPython's implementation or not.  If you're not feeling that pull then I think it's perfectly fine to keep moving forward with the current approach :)

One other comment -- the reason we had the separate `floatDiv`/`floatDivInt`/`floatDivFloat` functions is so that the JIT / rewriter could call one of the specialized methods if it already knew what the argument types were.  I'm not sure how much removing these will hurt -- would you mind getting set up with our performance testing tool and testing it to make sure there isn't a regression?  Something like
```sh
cd ~
git clone https://github.com/dropbox/pyston-perf
cd pyston-perf/analysis
python investigate.py origin/master numeric-fixing
```
As long as there's no regression (or within the testing noise), maybe we could get rid of the extra functions?  ie instead of still having `floatDiv()` which calls `floatDivFloat()`, we could just delete `floatDivFloat` and move its code into `floatDiv`.  Well, I know there's a lot of them, so I'll leave it up to you if you want to do that.

Daetalus:
Hi @kmod I updated the PR and ran perf tests several times, here is the result:

```
Took 6.0s, and exited with code 0
                                     origin/master:  origin/numeric_fix:
           django_template3.py            1.6s (14)            1.6s (22)  -0.2%
                 pyxl_bench.py            1.0s (14)            1.0s (20)  +0.7%
     sqlalchemy_imperative2.py            1.5s (14)            1.5s (20)  +0.4%
                pyxl_bench2.py            0.7s (14)            0.7s (20)  -0.1%
       django_template3_10x.py           10.1s (14)           10.2s (20)  +1.1%
             pyxl_bench_10x.py            7.3s (14)            7.3s (20)  +0.7%
 sqlalchemy_imperative2_10x.py           10.6s (14)           10.6s (20)  +0.0%
            pyxl_bench2_10x.py            5.9s (14)            5.8s (20)  -1.9%
                       geomean                 3.1s                 3.1s  +0.1%
```

undingen:
looks good :-), I will try it out tomorrow.
And sorry for being so pedantic I think still have to get used writing reviews.

src/runtime/float.cpp
```diff
@@ -862,7 +923,10 @@ template <ExceptionStyle S> static BoxedFloat* _floatNew(Box* a) noexcept(S == C
     } else if (PyLong_Check(a)) {
         double a_f = PyLong_AsDouble(a);
         if (a_f == -1.0 && PyErr_Occurred()) {
-            throwCAPIException();
+            if (S == CAPI)
```
undingen: good catch

src/runtime/float.cpp
```diff
@@ -998,6 +1058,47 @@ Box* floatTrunc(BoxedFloat* self) {
...
+}
+
+Box* floatLong(BoxedFloat* self) {
+    if (!PyFloat_Check(self))
+        raiseExcHelper(TypeError, "descriptor '__long__' requires a 'float' object but received a '%s'",
+                       getTypeName(self));
+
+    double wholepart; /* integral portion of x, rounded toward 0 */
+
+    (void)modf(self->d, &wholepart);
```
undingen: this and the int version will fail for inf floats
```>>float("inf").__int__()
pyston_dbg: ../../src/runtime/objmodel.cpp:4158: pyston::Box *pyston::callCLFunc(pyston::FunctionMetadata *, pyston::CallRewriteArgs *, int, pyston::BoxedClosure *, pyston::BoxedGenerator *, pyston::Box *, pyston::Box *, pyston::Box *, pyston::Box *, pyston::Box **) [S = 1, rewritable = 0]: Assertion `S == CAPI' failed.
```

And I think we can remove this ```(void)modf(self->d, &wholepart);``` call (at least cpython does not do this AFAIK)

src/runtime/float.cpp
Daetalus: This come from Rudi's patch. :smile:

src/runtime/float.cpp
```diff
@@ -213,6 +216,28 @@ extern "C" Box* floatRDiv(BoxedFloat* lhs, Box* rhs) {
         if (rhs_f == -1.0 && PyErr_Occurred()) {
             throwCAPIException();
         }
+        raiseDivZeroExcIfZero(rhs_f);
```
undingen: this should be ```raiseDivZeroExcIfZero(lhs->d);```
```
>> 2.0.__rdiv__(0l)
  File "<stdin>", line 1, in <module>
ZeroDivisionError: float division by zero
```
and 
```
>> 0.0.__rdiv__(2l)
inf
```
please also add tests for this cases

src/runtime/long.cpp
```diff
@@ -1389,6 +1470,27 @@ Box* longPow(BoxedLong* lhs, Box* rhs, Box* mod) {
...
 }
+Box* longPow(BoxedLong* lhs, Box* rhs, Box* mod) {
+    if (!PyLong_Check(lhs))
+        raiseExcHelper(TypeError, "descriptor '__pow__' requires a 'long' object but received a '%s'",
+                       getTypeName(lhs));
+
+    return longPowLong(lhs, rhs, mod);
+}
+
+Box* longRPow(BoxedLong* lhs, Box* _rhs) {
```
undingen: could you please also add the missing third arg support

undingen: oh the docs mention no 3. arg rpow https://docs.python.org/2/reference/datamodel.html
so please ignore my comment

undingen:
:+1: 
thanks

PR #1031, opened 2015-12-14
support inspecting exited frames (faster version)

undingen:
This replaces #1007 with a faster approach (I leave it up until this one is merged).

add frame intospection for exited frames
this works by copying the last locals state into the BoxedFrame when we exit a function.
As a perf improvement we only create BoxedFrames if _getframe got called or an exception got thrown.

It turned out that our python frame retrieving / updating code was really slow (using libunwind) I replaced it now with
a approach which manually keeps track of the python frames (calling initFrame/deinitFrame every time a functions gets called / exits).
Sadly we can't inline this functions in the llvm tier currently because it uses thread local storage which the JIT does not support.
But I have a prototype which works around the issue with inline asm.
We also store now always the current stmt inside the FrameInfo and in addition the FunctionMetadata.
This made it possible to switch away from libunwind frame introspection for most cases (except deopt and c++ unwinding where we need to access all register)

I think we can further improve the patch in several ways:
 - only copying the vregs array instead of generating the dict which takes more time (because of space overhead and hashing?).
 - refcounting should remove uneccesarry locals updating when no one keeps track of the BoxedFrame anymore.
 - switch to cpythons traceback implementation

```
                                   upstream/master:  origin/exited_fram:
           django_template3.py             2.1s (4)             2.2s (4)  +4.4%
                 pyxl_bench.py             1.8s (4)             1.9s (4)  +5.8%
     sqlalchemy_imperative2.py             2.2s (4)             2.3s (4)  +1.1%
                pyxl_bench2.py             1.1s (4)             1.1s (4)  +6.7%
       django_template3_10x.py            13.8s (4)            14.7s (4)  +6.5%
             pyxl_bench_10x.py            13.3s (4)            14.4s (4)  +8.7%
 sqlalchemy_imperative2_10x.py            17.0s (4)            17.6s (4)  +3.5%
            pyxl_bench2_10x.py             9.0s (4)             9.6s (4)  +6.8%
                       geomean                 4.7s                 5.0s  +5.4%
```

### DIFF
--- a/from_cpython/Include/pystate.h
+++ b/from_cpython/Include/pystate.h
@@ -111,6 +111,8 @@ typedef struct _ts {
 } PyThreadState;
 #endif
 typedef struct _ts {
+    void* frame_info; // This points to top python FrameInfo object
+
     int recursion_depth;
     int gilstate_counter;
 
--- a/from_cpython/Lib/logging/__init__.py
+++ b/from_cpython/Lib/logging/__init__.py
@@ -71,9 +71,15 @@ else:
     _srcfile = __file__
 _srcfile = os.path.normcase(_srcfile)
 
-# pyston changes: we don't support tb_frame or f_back, so always use sys._getframe
-currentframe = lambda: sys._getframe(4)
-start_getframe = 4
+# next bit filched from 1.5.2's inspect.py
+def currentframe():
+    """Return the frame object for the caller's stack frame."""
+    try:
+        raise Exception
+    except:
+        return sys.exc_info()[2].tb_frame.f_back
+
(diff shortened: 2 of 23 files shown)


kmod:
The 5% regression is worrying, but my thinking is that it's not worth trying to tackle it in isolation.  It seems like the slowdown is coming mostly from a general slowdown (icache contention?) which I think we need to look at as a whole anyway.

kmod:
One thing I noticed is that a lot of the slowdown (about half on pyxl_bench_10x) comes from setting the current_stmt variable.  Maybe instead of storing it as an `AST_stmt*` pointer, we could store it as an int index into an array of `AST_stmt*`?  This way we could use shorter instructions (8 bytes instead of 14), which might help.

PR #1036, opened 2015-12-18
Added missing include directory to cmakelist file

citisolo:
The cmake build was failing with "file does not exist 'Python.h'" . So I added the include directory in the CMakeLists.txt  file for the src directory and was able to build successfully


undingen:
Sorry for the late response :-(.
Unfortunately even though this change looks good I don't feel confident in merging it because I lack experience with the build system and don't know if this is correct. I would therefore like to wait until @kmod is back (end of this month). Sorry again.

kmod:
Erf sorry for the trouble (I introduced this issue right before I left).  I think #1081 is the proper fix, so I'm going to close this.

PR #1037, opened 2015-12-22
Fix #599 by allowing subclasses of string and unicode in sys.path

### LINKED ISSUE #599: Allow unicode entries in sys.path
Currently we ignore anything in `sys.path` that is not a str.  We should handle unicode entries by converting to strings (CPython does it using `PyUnicode_Encode(PyUnicode_AS_UNICODE(v), PyUnicode_GET_SIZE(v), Py_FileSystemDefaultEncoding, NULL);`), and also allow subclasses of string.

LoyukiL:
I use isSubclass check if a given entry in `sys.path` is subclass of `str` or `unicode` and tried to allow both situations. If it is `unicode`, I also convert it to `str` with `PyUnicode_AsEncodedString()` with `Py_FileSystemDefaultEncoding `.

I also modified `nullImporterInit` to make it accept subclasses of `str`.

### DIFF
--- a/src/runtime/import.cpp
+++ b/src/runtime/import.cpp
@@ -222,8 +222,15 @@ SearchResult findModule(const std::string& name, BoxedString* full_name, BoxedLi
     llvm::SmallString<128> joined_path;
     for (int i = 0; i < path_list->size; i++) {
         Box* _p = path_list->elts->elts[i];
-        if (_p->cls != str_cls)
+        if (isSubclass(_p->cls, unicode_cls)) {
+            _p = PyUnicode_AsEncodedString(_p, Py_FileSystemDefaultEncoding, NULL);
+            if (_p == NULL) {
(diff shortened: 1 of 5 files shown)


kmod:
Looks great!  Can we get a test though?  If you drop a file in test/tests/, it will get run and have its output compared against CPython's.  Since we have a file called `test/tests/test_package/import_target.py`, we could test this with something like
```python
import sys
sys.path.append(u"test_package")
import import_target
```

LoyukiL:
Hi! I add four individual tests with prefix of sys_path_ for four possible types of entries in `sys.path`, namely `str`, `unicode`, subclass of `str`, and subclass of `unicode`.

undingen:
LGTM, thanks for the patch!
Sorry for the late response but kmod and me where both on holiday.

Issue #596, opened 2015-06-09
all builtin types are subclassable, even when they shouldn't be


toshok:
the following example:

```python
class foo(slice):
    pass
```

succeeds in pyston and raises in cpython:

```
TypeError: Error when calling the metaclass bases
    type 'slice' is not an acceptable base type
```

this is due to the fact that Py_TPFLAGS_BASETYPE is or'ed into the tp_flags of all BoxedClasses.  we need a way to either keep that from happening or override it after BoxedClass creation time.  we also need to audit our builtins to make sure only the proper ones are subclassable.

undingen:
while ```slice``` is now fixed there are still some types left:
e.g:
```python
class C(type(xrange(1))): pass
```

undingen:
fixed in #1061

PR #1038, opened 2015-12-27
Fix issue #596 'slice is not subclassable'

corona10:
Fix issue #596 
make slice to be not subclassable.

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -4018,6 +4018,7 @@ void setupRuntime() {
     slice_cls->giveAttr("step", new BoxedMemberDescriptor(BoxedMemberDescriptor::OBJECT, offsetof(BoxedSlice, step)));
     slice_cls->freeze();
     slice_cls->tp_compare = (cmpfunc)slice_compare;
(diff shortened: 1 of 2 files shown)


undingen:
LGTM, thanks for the patch!

While this fixes it for the slice type there are a bunch of other types which still allow subclassing even they should not. I'm wondering if we should not at some point switch to a approach where types have to explicitly say what additional flags they want to have because it may be less work and more similar to the capi and cpython.

corona10:
@undingen  Thanks i agree with your opinion. Do you know any good document or article to get the relevant information about which not to be subclassable list. When i know that lists than i can try to find another approach.

undingen:
I don't know of a list of types which don't allow subclassing :-(.
My approach would be to remove the line inside the constructor of ```BoxedClass``` which always enables subclassing. And than add the missing flags to all our C++ types which show up inside a search for ```Py_TPFLAGS_BASETYPE``` inside cpythons Objects/* dir. CAPI types should all be fine so I suspect this are not many types which need it and this are probably mostly types where we already are modifying the ```tp_flags``` member.

PR #1040, opened 2016-01-05
Add missing complex attributes

Daetalus:


undingen:
I haven't tested the commit yet (will do tomorrow) but just from the changes it looks like a very nice improvement!

from_cpython/Objects/complexobject.c
```diff
@@ -356,15 +354,20 @@ PyComplex_AsCComplex(PyObject *op)
...
 complex_dealloc(PyObject *op)
 {
     op->ob_type->tp_free(op);
 }
+#endif
 
 
-static PyObject *
-complex_format(PyComplexObject *v, int precision, char format_code)
+// Pyston change: make no static
```
undingen: I'm wondering why you had to change the signature?
I think every call site has a BoxedComplex* so it could just pass that one in instead of creating a ```Py_complex``` just to pass it around. (But we should add static asserts which make sure that BoxedComplex::imag/real have the same offsets as CAPI PyComplexObject::cval.imag/real)

src/runtime/complex.cpp
```diff
@@ -517,398 +417,172 @@ Box* complexStr(BoxedComplex* self) {
     if (!PyComplex_Check(self))
         raiseExcHelper(TypeError, "descriptor '__str__' requires a 'complex' object but received a '%s'",
                        getTypeName(self));
-    Box* r = complex_fmt(self->real, self->imag, 12, 'g');
+
+    // make sure that BoxedComplex::imag/real have the same offsets as
+    // CAPI PyComplexObject::cval.imag/real
+    static_assert(offsetof(BoxedComplex, real) == offsetof(PyComplexObject, cval.real), "");
```
undingen: Can we please move this into ```runtime/types.h``` directly after the ```BoxedComplex``` declaration like we already do for other types (eg. BoxedFloat) and add a additional ```static_assert(sizeof(BoxedComplex) == sizeof(PyComplexObject), "")``` assert.
sorry that I did not make this clear from the beginning :-(.

undingen:
thanks for your patch :+1:

PR #1042, opened 2016-01-06
fix minor compatibility issues

undingen:
- implement ```attrwrapper.has_key```
- switched to cpythons ```dictview``` types because we were missing a lot of functionality.
- different types for the dict iterators
- support empty tp_new
- add descriptor default type arg: descr.__get__(self, obj, type=None)


undingen:
I forgot to add the dictview test but added in #1045

PR #1043, opened 2016-01-07
add PyGC_AddPotentialRoot

undingen:
This function let's one register a potential root range from the CAPI
Was part of the unmerged #1026 but split out because #1029 requires it too.


undingen:
This contains a bad bug see #1049 for a fix. Sorry for the trouble :-(

PR #1050, opened 2016-01-13
List sort now support lambda parameter

corona10:
List sort now support lambda parameter.
I think there are a lot of codes to refactoring or fixing it in this PR.
Please tell me about those codes.

### DIFF
--- a/src/runtime/list.cpp
+++ b/src/runtime/list.cpp
@@ -832,19 +832,47 @@ extern "C" int PyList_Reverse(PyObject* v) noexcept {
 class PyCmpComparer {
 private:
     Box* cmp;
+    Box* key;
+    bool reverse;
 
 public:
-    PyCmpComparer(Box* cmp) : cmp(cmp) {}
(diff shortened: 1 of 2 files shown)


undingen:
Sadly I think implementing sort is not that easy :-(.
Without testing it carefully I see at least the following problems with this implementation:
- it calls the cmp function before calling the key function with the orig args!
- it does not support key function which return non ints.
- I think we should only call the key function once per element.

Getting ```test/cpython/test_sort.py``` to pass requires several additional modifications.
Using cpythons list sort implementation will also require some changes but I wonder if this is not the better path. It looks like they but a lot of effort into it (see: Objects/listsort.txt).

corona10:
@undingen okay i should find a better way :-)

corona10:
I solve this problem 
`not support key function which return non ints.`
but if we want to solve 
* it calls the cmp function before calling the key function with the orig args!
* I think we should only call the key function once per element.
we need to use cpython implementation in my guess (Tim Sort)

PR #1052, opened 2016-01-13
fixes for py.test

undingen:


undingen:
this PR together with #1031 makes py.test work (with the fancy assert messages)

PR #1053, opened 2016-01-13
switch to CPython parser

Daetalus:


undingen:
:+1:

undingen:
my reasoning for making cpython the default parser:
pypa served us very well!
But we are seeing some smaller parse errors (with complex numbers,..) and the different syntax error message are concerning me a little bit.
It probably takes not much time to fix the issues inside pypa but than this is one more difference too cpython which we can avoid and it makes more sense to focus our time on other tasks.

one can still use pypa by passing ```-X``` to pyston.

PR #1054, opened 2016-01-14
Tuple and list calling order fixing.

Daetalus:
For `tuple.__add__`, if we throw an exception in our `tuple.__add__`, we won't bother checking for an `__radd__`. So before throw exception, I call `__radd__` first. If both return `NotImplemented`. Throw an exception.

For `list.__mul__`. Before throw `TypeError, XXX object cannot be interpreted as an index`., I check the argument type whether is `int` or `long` first. If not, throw `TypeError, can't multiply sequence by non-int of type XXX`. This may not the correct solution, but the behavior is more like CPython.

### DIFF
--- a/src/runtime/list.cpp
+++ b/src/runtime/list.cpp
@@ -691,11 +691,15 @@ extern "C" int PyList_Insert(PyObject* op, Py_ssize_t where, PyObject* newitem) 
 }
 
 Box* listMul(BoxedList* self, Box* rhs) {
-    static BoxedString* index_str = internStringImmortal("__index__");
+    Py_ssize_t n;
 
-    Py_ssize_t n = PyNumber_AsSsize_t(rhs, PyExc_IndexError);
-    if (n == -1 && PyErr_Occurred())
-        throwCAPIException();
+    if (PyIndex_Check(rhs)) {
+        n = PyNumber_AsSsize_t(rhs, PyExc_OverflowError);
+        if (n == -1 && PyErr_Occurred())
+            throwCAPIException();
+    } else {
+        return NotImplemented;
+    }
 
     int s = self->size;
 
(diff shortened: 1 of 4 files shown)


undingen:
Interesting problem...
Unfortunately I don't think this is the correct solution to the problem. To get the exact cpython error messages I think this case needs to be handled inside ```binopInternal``` by adding something similar to ```PyNumber_Multiply``` and ```sequence_repeat``` (and for add etc..). But that looks quite complicated and I would highly recommend waiting for feedback from @kmod before doing this change.

We could also implement pypys behavior (=return ```NotImplemented``` instead of raising an exception) instead like kmod has done in https://github.com/kmod/pyston/commit/e03daee366daad25f3115ecf61ae4da7e051999b for the ```list.__add__```. I think this looks like the nicer approach but it may require changes to some capi types :-(.
Would this fix the problem you are seeing or does numpy require this change?

kmod:
Some background: [I think this is a Python-dev thread about this very issue](https://mail.python.org/pipermail/python-dev/2015-May/140006.html), and [here is a PR where Rudi ran into this as well](https://github.com/dropbox/pyston/pull/1014#commitcomment-14775965).

I think our options are:

1. support CPython's odd behavior around sq_concat/sq_repeat
2. do the "manually call `__rmul__` for these functions" approach of this PR
3. continue to throw the TypeError in `tuple.__mul__`
4. return NotImplemented instead

Option 1 might be the best long-term, but I think it's a bunch of work and we don't have a pressing need for it.  I don't think option 3 is really viable since we keep running into code that breaks because of it.  I think both option 2 and 4 will probably work, but I'm inclined towards 4 since a) it's the way people think this should be handled (and is PyPy's behavior), and b) we're using this approach in other places and have seen only minor issues.

I think the approach in this PR could probably work, but it's worrying to change the behavior of `__mul__` like this to silently call another method -- what if someone really only wanted to call `tuple.__mul__`?  Seems unlikely but again it's worrying.

I agree with Marius -- can we try having this just return NotImplemented and see if that solves the problem?

Daetalus:
@kmod , updated, return `NotImplemented` could solves the problem, only the error message is different from CPython's. Would you mind to review it if you have time, please? Thank you!

kmod:
I think the different error message is ok (I think we already gave up on 100% error message matching).  This PR still leaves us with slightly different behavior from either CPython or PyPy (which are themselves slightly different) -- they will both prefer the rmul variant when doing a list multiply.  For example we give something different for this:
```python
class C(object):
    def __rmul__(self, lhs):
        print "rmul", lhs
        return lhs * 2

    def __index__(self):
        print "index"
        return 5

print [1, 2] * C()
```

kmod:
I think this difference is less bad than the one that we had before the PR, so this is still a good improvement :)  But at some point we might have to add logic to the `binop()` code that will handle this the right way.

PR #1057, opened 2016-01-18
@rudi-c #1027 numpy patch with a small fix

undingen:
Hopefully this fixes the lxml travis ci error (I cant reproduce it locally)


undingen:
this is #1027 with a small patch which fixes the test error (I restarted travis-ci about 10 times and the error did not happen while without the patch it happened immediately)

PR #1058, opened 2016-01-18
add formencode test

undingen:


undingen:
this depends on #1031

PR #1059, opened 2016-01-18
enable PyStringObject and remove some workarounds inside the numpy patch

undingen:


undingen:
@Daetalus thanks for taking a look at the patch! I added a test for the string case and checked with gdb if it does what it should do.

PR #1060, opened 2016-01-18
Two minor problem which found in NumPy tests

### LINKED ISSUE #1012: Support numerical operations on inherited numerical types
The following works in CPython, but crashes in Pyston (the compiler doesn't recognize `f` as a float object and terminates with `couldn't find suitable function version and no source to recompile!`).

```python
class F(float):                         
    def __new__(cls, value):            
        return float.__new__(cls, value)
                                        
f = F(3)                                
print f * 2                                         
```

This is just multiplication of a float, but it's worth making sure that all operations for all numerical types (int, float, complex) are supported.

This is used in practice in NumPy. For example, in `test_function_base.py`, you can find the following code:

```python
a = PhysicalQuantity(0.0)
b = PhysicalQuantity(1.0)
assert_equal(linspace(a, b), linspace(0.0, 1.0))
```

Daetalus:
closes #1012

### DIFF
--- a/src/runtime/float.cpp
+++ b/src/runtime/float.cpp
@@ -115,24 +115,24 @@ extern "C" double floordiv_float_float(double lhs, double rhs) {
 }
 
 extern "C" Box* floatAddFloat(BoxedFloat* lhs, BoxedFloat* rhs) {
-    assert(lhs->cls == float_cls);
-    assert(rhs->cls == float_cls);
+    assert(PyFloat_Check(lhs));
+    assert(PyFloat_Check(rhs));
     return boxFloat(lhs->d + rhs->d);
 }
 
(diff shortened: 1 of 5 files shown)


src/runtime/float.cpp
```diff
@@ -1120,7 +1120,7 @@ static void _addFunc(const char* name, ConcreteCompilerType* rtn_type, void* flo
     v_ff.push_back(BOXED_FLOAT);
     v_fi.push_back(BOXED_FLOAT);
     v_fi.push_back(BOXED_INT);
-    v_fu.push_back(BOXED_FLOAT);
+    v_fu.push_back(UNKNOWN);
```
undingen: very minor: could you please rename the ```v_fu``` variable to ```v_uu``` now that the first type is unknown

undingen:
:+1: thanks for the patch!

PR #1061, opened 2016-01-19
Add list of not subclassable class

corona10:
As we descuss about [not subclassable class] at #1038
i got  lists of 'should not be subclassalbe class' by `grep -L -e "Py_TPFLAGS_BASETYPE" Objects/  -R`
and check tp_name if tp_name is not subclassable type.
than I off the `Py_TPFLAGS_BASETYPE`

### DIFF
--- a/src/runtime/str.cpp
+++ b/src/runtime/str.cpp
@@ -2805,7 +2805,7 @@ void setupStr() {
     str_cls->tp_flags |= Py_TPFLAGS_HAVE_NEWBUFFER;
 
     str_iterator_cls = BoxedClass::create(type_cls, object_cls, &BoxedStringIterator::gcHandler, 0, 0,
-                                          sizeof(BoxedStringIterator), false, "striterator");
+                                          sizeof(BoxedStringIterator), false, "striterator", false);
     str_iterator_cls->giveAttr(
         "__hasnext__", new BoxedFunction(FunctionMetadata::create((void*)BoxedStringIterator::hasnext, BOXED_BOOL, 1)));
     str_iterator_cls->giveAttr(
--- a/test/tests/builtin_type_not_subclassable.py
+++ b/test/tests/builtin_type_not_subclassable.py
@@ -1,3 +1,58 @@
-# should_error
-class foo(slice):
-   pass
+import re
+import inspect
+def gen():
+  print ("generator test")
+  yield 'a'
+  yield 'b'
+ 
+def is_subclassable(base_cls):
+  try:
+    class C(base_cls):
+      pass
+
+  except TypeError as e:
+    assert 'is not an acceptable base type' in repr(e)
+    return False
+  return True
+
+class b:
+  def __init__(self):
+    self.c = 1
+  def d(self):
+    print self.c
+
+#slice
+assert not is_subclassable(slice)
+#xrange
+assert not is_subclassable(type(xrange(1)))
+#range-iterator
+assert not is_subclassable(type(iter(range(5))))
+#xrange-iterator
+assert not is_subclassable(type(iter(xrange(5))))
+#callable-iterator
+assert not is_subclassable(type(re.finditer(r'\bs\w+', "some text with swords")))
+#striterator
+assert not is_subclassable(type(iter("abc")))
+#memoryview
+assert not is_subclassable(type(memoryview("abc")))
+#buffer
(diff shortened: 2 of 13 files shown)


undingen:
Unfortunately this only fixes part of the problem. Still all capi types are subclassable so this fails to work for some of the ones which are on the list but implemented in C instead of using the c++ api e.g. [1] and for extensions which may define types which should not be subclassable.
I think we really should switch the condition...

[1]:
```python
class C(type(iter("abc"))):
 pass
```

corona10:
I change striterator tp_name to iterator.
```
>> type(iter("abc"))
<type 'striterator'>  --> should be <type 'iterator'> in Cpython
```

and I add case of `callable-iterator' and it works rightly.

test/tests/builtin_type_not_subclassable.py
```diff
@@ -1,3 +1,19 @@
 # should_error
+import re
+
 class foo(slice):
    pass
+
+class C(type(xrange(1))):
```
kmod: I don't think any of this code will get run, since we will just exit at the first uncaught exception.

corona10: @kmod hmm You're right. than should I seperate test file or is there any good solution?

src/runtime/str.cpp
```diff
@@ -2805,7 +2805,7 @@ void setupStr() {
     str_cls->tp_flags |= Py_TPFLAGS_HAVE_NEWBUFFER;
 
     str_iterator_cls = BoxedClass::create(type_cls, object_cls, &BoxedStringIterator::gcHandler, 0, 0,
-                                          sizeof(BoxedStringIterator), false, "striterator");
+                                          sizeof(BoxedStringIterator), false, "iterator");
```
kmod: I think the issue is not that we named this class wrong, but that it's not supposed to exist at all.  I had started to address this in #957, but it got held up because it caused a perf regression.  Long story short, I think we should leave the name the same here.

test/tests/builtin_type_not_subclassable.py
kmod: We should try to avoid using `#should_error` now -- it's pretty much only for tests that are supposed to crash.  You could try doing something like.

```python
def is_subclassable(base_cls):
  try:
    class C(base_cls):
      pass
  except TypeError:
    return false
  return true

assert is_subclassable(int)
assert not is_subclassable(slice)
assert not is_subclassable(type(xrange(1)))
```

src/runtime/str.cpp
corona10: Hmm.. My approach is checking `tp_name` whether should not be subclassable.
As far as I know `striterator ` is not in the Cpython.
so I add `striteator' as should not be suclassable tp_name. 
(This seems to be the case limited to pyston.)

corona10:
* Change test code.
* As far as I know `striterator` is not in the `Cpython`.
so I add `striteator` as should not be suclassable `tp_name`. 
(This seems to be the case limited to `pyston`.)

kmod:
While this seems to be working, it's a bid weird that the class setup code knows which classes want which flags or not.  I think this would be cleaner if instead of having a list that freeze() knows about, the creator of the class (the one calling freeze() or `new BoxedHeapClass()`) should be the one specifying whether the class is subclassable.  It's also means that we won't have the issue of having different classes with the same name that (under this scheme) would have to have the same flags.

kmod:
Also, this is very minor, but I think it's worth verifying that the test is actually doing what we expect.  ie maybe something like
```python
except TypeError as e:
  assert 'is not an acceptable base type' in repr(e)
  return False
```
and then it's nice to have a check of the is-subclassable case (`assert is_subclassable(list)`) just as a sanity check.  It's surprisingly common that we have a test that doesn't actually test what we thought it would.

corona10:
I'll look for a better way about first comment.
I am learning a lot from you.

undingen:
I like this approach much better :-).
But noticed that you missed ```type(iter(xrange(1)))``` and ```type(is_subclassable)```.
I don't really like that it's just a bool because it kind of hard to know what this bool means but I think it's fine for now. If we wan't to pass additional flags in the future we can change it.
But please squash the commits into one.
Maybe @kmod has some additional commits but I think it should be ready for merging after this changes. :-)

corona10:
@undingen Oh thanks! I added what i missed cases and squash into one commit.

PR #1062, opened 2016-01-20
Fix minor compatibility issues

undingen:


undingen:
oh I overlooked that ```long(None)``` has the same problem... I will fix it soon

Issue #1069, opened 2016-01-25
Remove bash-isms


tony:
Remove bashisms in local scripts and (if possible) external dependencies also.

PR #1072, opened 2016-01-25
Switch to /usr/bin/env sh for portability

tony:
See also: #1069 

There are other scripts (``configure_travis.sh`` and ``src/run-tests.sh`` in libpypa upstream). Need to get a Ubuntu system up to double check they wouldn't break (they would if there we bash-specific code in it)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SHELL := /bin/bash
+SHELL := /usr/bin/env sh
 
 # prints variables for debugging
 print-%: ; @echo $($*)


tony:
Works on Ubuntu

kmod:
Looks good to me -- please sign our [CLA](https://opensource.dropbox.com/cla/) and then I'll merge it :)

tony:
Done

kmod:
Awesome, thanks!

PR #1075, opened 2016-01-25
EH frame: use a single allocation for the EH frame + generated code

undingen:
This removes the 2GB process limitation I ran into.
(With the EH frame format we are using the code offset is specified as a 32bit signed offset)
I checked if aligning the code speeds it up but I did not notice any perf change.


kmod:
:+1:

PR #1076, opened 2016-01-26
Small fixes for Flask

corona10:
* Small fixes for Flask
* Check some of the template rendering jinja2

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -3732,7 +3732,6 @@ void setupRuntime() {
                                            false, "method-wrapper");
     wrapperdescr_cls = new (0) BoxedClass(object_cls, BoxedWrapperDescriptor::gcHandler, 0, 0,
                                           sizeof(BoxedWrapperDescriptor), false, "wrapper_descriptor");
-
     EmptyString = new (0) BoxedString("");
     // Call InternInPlace rather than InternFromString since that will
     // probably try to return EmptyString
--- a/test/extra/flask_test.py
+++ b/test/extra/flask_test.py
@@ -1,0 +1,46 @@
+#-*- coding: utf-8 -*-
+#flask test
+#todo: It should be tested as django_test.py.
+import os, sys, subprocess, shutil
+import multiprocessing, time, signal
+import urllib2
+
+sys.path.append(os.path.dirname(__file__) + "/../lib")
+from test_helper import create_virtenv, run_test
+
+ENV_NAME = "flask_test_env_" + os.path.basename(sys.executable)
+create_virtenv(ENV_NAME, 
+               ["flask == 0.10.1", 
+                "Werkzeug == 0.11.3",
+                "jinja2 == 2.8", 
+                "itsdangerous == 0.24", 
+                "markupsafe == 0.23"]
+               , force_create = True)
+
+sys.path.append(ENV_NAME + "/site-packages")
+
+from flask import Flask, Markup
+from jinja2 import Template 
+
+app = Flask(__name__)
+
+@app.route('/')
+def test_template():
+    t = Template("Hello, World!: {% for n in range(1,10) %}{{n}} " "{% endfor %}")
+    return t.render()
+
+if __name__ == '__main__':
+    app.config['TESTING'] = True
+    assert isinstance(app, Flask)
+    server = multiprocessing.Process(target=app.run)
+    server.start()
+    time.sleep(1)
+    f = urllib2.urlopen("http://127.0.0.1:5000/", timeout=1)
+    s = f.read()
+    assert 'Hello, World!: 1 2 3 4 5 6 7 8 9' in s
+    server.terminate()
+    server.join()
+    m1 = Markup('<strong>Hello %s!</strong>') % '<blink>hacker</blink>'
+    m2 = Markup.escape('<blink>hacker</blink>')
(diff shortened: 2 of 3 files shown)


undingen:
The ```__name__``` implementations are wrong:
cpython
```
>>> type(u"").__add__.__name__
'__add__'
```
pyston with this patch:
```
>> u.__add__.__name__
'wrapper_descriptor'
```

I think they should return ```descr->wrapper->name``` which is not a boxed type so you will register a small helper function with ```giveAttrDescriptor``` similar to ```__doc__```.

Concerning the test: could you please download a tarball because otherwise the test will change when the source repo changes and it should be faster to download too...

src/runtime/types.cpp
```diff
@@ -3728,11 +3728,12 @@ void setupRuntime() {
         BoxedClass(object_cls, BoxedCApiFunction::gcHandler, 0, 0, sizeof(BoxedCApiFunction), false, "capifunc");
     method_cls = new (0) BoxedClass(object_cls, BoxedMethodDescriptor::gcHandler, 0, 0, sizeof(BoxedMethodDescriptor),
                                     false, "method_descriptor");
+    method_cls->giveAttr("__name__", boxString("method_descriptor"));
```
kmod: I don't think this is quite right -- I think this is supposed to be the name of the method object itself, not the class.  We should add `print object.__reduce__.__name__` to the test -- CPython says `__reduce__` and we will say `method_descriptor`.  I think this should be something like 
```C++
method_cls->giveAttr("__name__", new BoxedMemberDescriptor(BoxedMemberDescriptor::OBJECT,
                                                       offsetof(BoxedMemberDescriptor, name));
```
ie this is the magic to say that the `__name__` attribute should proxy to the C field `BoxedMemberDescriptor::name`.  This won't work right away since BoxedMemberDescriptor doesn't have a `name` field, so one would need to be added.

src/runtime/types.cpp
```diff
@@ -3728,11 +3728,12 @@ void setupRuntime() {
...
     wrapperobject_cls = new (0) BoxedClass(object_cls, BoxedWrapperObject::gcHandler, 0, 0, sizeof(BoxedWrapperObject),
                                            false, "method-wrapper");
     wrapperdescr_cls = new (0) BoxedClass(object_cls, BoxedWrapperDescriptor::gcHandler, 0, 0,
                                           sizeof(BoxedWrapperDescriptor), false, "wrapper_descriptor");
-
+    wrapperdescr_cls->giveAttr("__name__", boxString("wrapper_descriptor"));
```
kmod: Same comment for this one -- `print int.__add__.__name__` should give `'__add__'`.

undingen:
But I very like the idea of having another larger library integration test :-) so I don't know the current state and if you have time but running there rtests would be really cool (even if we fail some of them).

test/extra/flask_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+from test_helper import create_virtenv, run_test
+ENV_NAME = "flask_test_env_" + os.path.basename(sys.executable)
+SRC_DIR = os.path.abspath(os.path.join(ENV_NAME, "src"))
+PYTHON_EXE = os.path.abspath(os.path.join(ENV_NAME, "bin", "python"))
+
+def install_flask():
+    shutil.rmtree(SRC_DIR, ignore_errors=True)
+    os.makedirs(SRC_DIR)
+    subprocess.check_call(["git", "clone", "https://github.com/mitsuhiko/flask.git"], cwd=SRC_DIR)
```
kmod: Some minor recommendations about this -- any reason we can't download these from PyPI?  `pip install flask` seemed to work -- the second argument to create_virtualenv() is a list of packages to install, so if you pass `["flask"]` it should do what you want.  If you really want to go to the git repo, you can do `["git+https://github.com/mitsuhiko/flask.git"]` and pip will do these steps for you.  But if there isn't any particular reason to pull them from git, I'd recommend switching this to PyPI since that should be faster.  If you do that, `install_flask()` can just become `sys.path.append(os.path.join(ENV_NAME, "site-packages"))`.

One additional consideration (regardless of the git vs pypi decision) is that we want to pin the version of every library we test.  It's quite annoying when someone updates their library in a way that breaks our test, and our test picks it up since it is configured to pull whatever the latest version is.  If you switch this to pip, please try running `pip install flask werkzeug jinja2...` to see what versions it installs, and then pin the dependencies like `create_virtualenv(ENV_NAME, ["flask==0.0.0", "werkzeug==9.9.9", ...])`.

kmod:
I think Marius's suggestion of a getset descriptor is probably a better idea than the member descriptor that I suggested -- CPython does it with a member descriptor and it'd be nice to move closer to their behavior here, but it would require some other changes so it's probably not worth it for not.

corona10:
* Modify `name` implementation as you said 

  ```
>> type(u"").__add__.__name__
'__add__'  --> now it works
>> print object.__reduce__.__name__
__reduce__ --> also it works
  ```

* Modify flask test code to install with `pip` and fix to use flask 0.10.1 version

undingen:
Looking good! With a small comment: could you please specify the version number of all dependencies in the pip package list: Werkzeug==0.11, Jinja2==2.8, itsdangerous==0.24, MarkupSafe==0.23 and stash the commits. thanks

undingen:
Thanks for the patch!

PR #1079, opened 2016-01-28
implement tb_lineno of Traceback

corona10:
I implements `tb_lineno` but I don't know this is right implementatation.
I know `tb_linno` should assigend by frame's `f_lineno` but this implementation looks different.
I check it print right `tb_lineno` on below code after this patch...

```python
import sys, traceback

def lumberjack():
    bright_side_of_death()

def bright_side_of_death():
    return tuple()[0]

try:
    lumberjack()
except IndexError:
    exc_type, exc_value, exc_traceback = sys.exc_info()
    print "*** print_tb:"
    traceback.print_tb(exc_traceback, limit=1, file=sys.stdout)
    print "*** print_exception:"
    traceback.print_exception(exc_type, exc_value, exc_traceback,
                              limit=2, file=sys.stdout)
    print "*** print_exc:"
    traceback.print_exc()
    print "*** format_exc, first and last line:"
    formatted_lines = traceback.format_exc().splitlines()
    print formatted_lines[0]
    print formatted_lines[-1]
    print "*** format_exception:"
    print repr(traceback.format_exception(exc_type, exc_value,
                                          exc_traceback))
    print "*** extract_tb:"
    print repr(traceback.extract_tb(exc_traceback))
    print "*** format_tb:"
    print repr(traceback.format_tb(exc_traceback))
    print "*** tb_lineno:", exc_traceback.tb_lineno

```

### DIFF
--- a/src/runtime/traceback.cpp
+++ b/src/runtime/traceback.cpp
@@ -145,7 +145,9 @@ void setupTraceback() {
     NULL));
     */
     traceback_cls->giveAttrDescriptor("tb_next", traceback_tb_next, NULL);
-
+    traceback_cls->giveAttr("tb_lineno",
+                            new BoxedMemberDescriptor(BoxedMemberDescriptor::INT,
+                                                      offsetof(BoxedTraceback, line) + offsetof(LineInfo, line)));
     traceback_cls->freeze();
 }
 }
--- a/src/runtime/traceback.h
+++ b/src/runtime/traceback.h
@@ -30,7 +30,6 @@ public:
     Box* tb_next;
     LineInfo line;
     Box* py_lines;
-
     BoxedTraceback(LineInfo line, Box* tb_next) : tb_next(tb_next), line(std::move(line)), py_lines(NULL) {}
 
     DEFAULT_CLASS(traceback_cls);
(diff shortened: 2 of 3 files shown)


undingen:
Your patch looks good but I would prefer if you would add a test.
#1031 implements the same attribute + some additional stuff but because it may take some more time until this PR lands I'm fine with merging this one in between and rebasing mine. 
But I definitely recommending to not but more effort into the traceback and frame object until #1031 landed. (I think after #1031 is merged we can even easily switch to the cpython traceback implementation)

corona10:
@undingen Thanks, I added test. But I don't know it will be good or not to be merged mine if you 're already make patch for this. :-)

src/runtime/traceback.cpp
```diff
@@ -145,7 +145,8 @@ void setupTraceback() {
     NULL));
     */
     traceback_cls->giveAttrDescriptor("tb_next", traceback_tb_next, NULL);
-
+    traceback_cls->giveAttr("tb_lineno",
+                            new BoxedMemberDescriptor(BoxedMemberDescriptor::INT, offsetof(BoxedTraceback, tb_lineno)));
```
kmod: You could probably do `offsetof(BoxedTraceback, line) + offsetof(LineInfo, line)` instead of storing the attribute twice.

corona10:
@kmod  Thanks That's reasonable.

undingen:
Thanks for the change! :+1:

PR #1080, opened 2016-01-28
sysconfig: adjust paths

undingen:


undingen:
this seems to fix the issue I was running in (pip does not install a package dependency because the cpython version is already installed.)
But I'm not sure how the other path in this file should look like...

kmod:
I honestly have no idea :)  If this makes the metaserver virtualenv look right I think we should just merge it.

PR #1081, opened 2016-01-29
Add missing runtime/inline include dependence

kmod:
I think this was the cause of the 'Python.h not found' errors.


undingen:
so I'm going to merge this in because all builds are failing without it if I don't delete the cache...

undingen:
oh I just saw that I did not had to merge this PR in (I was looking at the wrong log and thought that's the problem...) :-( I will leave it in but if you did not intend to merge it yet feel free to revert it. 
Sorry for the trouble / my confusion...

PR #1082, opened 2016-01-29
Get libsass working by fixing c++ EH handling

undingen:
Don't export the syms of our c++ exception handling implementation
Otherwise we break external libraries which use c++ exceptions internally.

I think without using a linker script we can't directly influence the visibility of ```std::uncaught_exception``` so I replaced the calls with a new function ```isUnwinding```. This is kind of ugly because now we have to remember that we are not allowed to call it but on the other hand we already have to remember that we are not allowed to call ```std::rethrow_exception``` etc...  - I'm definitely open to better idea.


kmod:
What could go wrong :P

Issue #1083, opened 2016-01-30
Not raise typeError with datetime.timedelta


corona10:
``` python
import datetime
a = datetime.timedelta(43)
b = 4
a // b
c = 4.3
c // a  # it raised TypeError
a // c  # should raise TypeError but not it crahsed..
```

PR #1084, opened 2016-01-30
Quick fix for issue #1083

corona10:
* Quick fix for issue #1083
* use _PyObject_GetDictPtr of cpython implementation
* `std::sort(allDerefVarsAndInfo.begin(), allDerefVarsAndInfo.end(), derefComparator);` to remove functor and use lambda expression

now pyston can fully execute `test/cpython/test_datetime.py`

### DIFF
--- a/src/analysis/scoping_analysis.cpp
+++ b/src/analysis/scoping_analysis.cpp
@@ -423,16 +423,14 @@ public:
             }
 
             // Sort in order of `num_parents_from_passed_closure`
-            std::sort(allDerefVarsAndInfo.begin(), allDerefVarsAndInfo.end(), derefComparator);
+            std::sort(
+                allDerefVarsAndInfo.begin(), allDerefVarsAndInfo.end(),
+                [](const std::pair<InternedString, DerefInfo>& p1, const std::pair<InternedString, DerefInfo>& p2) {
+                    return p1.second.num_parents_from_passed_closure < p2.second.num_parents_from_passed_closure;
+                });
         }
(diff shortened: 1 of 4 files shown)


kmod:
We have some differences from CPython in how we represent attributes, so we have to be very careful with how we implement _PyObject_GetDictPtr.  Specifically, we have objects which have attributes which aren't stored as dicts (we store them as hidden-class-based attributes aka "hcattrs"), and for those objects I'm not sure if there's any reasonable way to handle a call to _PyObject_GetDictPtr.

I think maybe the most straightforward solution for now is to add a `RELEASE_ASSERT(!tp->instancesHaveHCAttrs())` assert before the code that looks at tp_dictoffset.  I think the code you copied over should be correct for the non-hcattrs case, but it will get the hcattrs case wrong so we need to add the check first.

corona10:
@kmod Thanks! I don't know whether we can solve the problem but I find out more about the problem. I modify it as use Cpython implementation when it has `HCAttrs` if not it throws non implementation error.
so now pyston can not execute `test/cpython/test_datetime.py` fully until this problem resolved. :disappointed: 

And, Is threre any problem about fixing issue #1083 and remove functor `derefComparator`??

kmod:
Awesome, thanks!

Personally I lean away from making changes like switching derefComparator if it wasn't hurting us at all, but I won't argue against doing the cleanup if you are motivated :)

corona10:
@kmod Thanks~

Issue #1017, opened 2015-11-27
gmp.h file not found in libpypa on 11-CURRENT


tony:
On FreeBSD 11-CURRENT

```
/usr/home/tony/study/python/pyston posix
❯ gmake
echo "Creating default Makefile.local"
Creating default Makefile.local
which ninja-build >/dev/null && echo "NINJA := ninja-build" >> Makefile.local
ninja -C /usr/home/tony/study/python/pyston/build/Debug pyston copy_stdlib copy_libpyston sharedmods ext_pyston ext_cpython 
ninja: Entering directory `/usr/home/tony/study/python/pyston/build/Debug'
[199/3098] Generating ../../build_deps/libunwind/pyston_patched
FAILED: cd /usr/home/tony/study/python/pyston && git submodule update build_deps/libunwind && python /usr/home/tony/study/python/pyston/tools/git_am_automated.py build_deps/libunwind /usr/home/tony/study/python/pyston/libunwind_patches/0001-pyston-add-lots-of-comments.patch /usr/home/tony/study/python/pyston/libunwind_patches/0002-pyston-stop-x86_64-setcontext-restoring-uninitialize.patch /usr/home/tony/study/python/pyston/libunwind_patches/0003-use-a-sorted-array-for-registered-objects-and-do-a-b.patch /usr/home/tony/study/python/pyston/libunwind_patches/9999-is-patched-marker.patch
Applying: pyston: add lots of comments
error: patch failed: src/dwarf/Gparser.c:429
error: src/dwarf/Gparser.c: patch does not apply
error: patch failed: src/mi/Gdyn-extract.c:33
error: src/mi/Gdyn-extract.c: patch does not apply
error: patch failed: src/mi/Gget_proc_name.c:55
error: src/mi/Gget_proc_name.c: patch does not apply
error: patch failed: src/x86_64/Gstep.c:158
error: src/x86_64/Gstep.c: patch does not apply
Patch failed at 0001 pyston: add lots of comments
The copy of the patch that failed is found in: /usr/home/tony/study/python/pyston/.git/modules/build_deps/libunwind/rebase-apply/patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
----- Running `git am --abort' -----
[199/3098] Building CXX object build_deps/libpypa/src/CMakeFiles/pypa.dir/pypa/parser/parser.cc.o
[...]

PR #1085, opened 2016-01-30
Fix finding GMP on FreeBSD

tony:
Update adds CMake module for finding GMP (https://github.com/vinzenz/libpypa/pull/50)

Fixes: #1017


kmod:
:+1:

PR #1086, opened 2016-02-01
Fix cpython parser bool operation conversion

undingen:
Because we only checked for Or's instead of checking for Ands this did not cause any issues except when printing the CFG


kmod:
Hah :)

PR #1090, opened 2016-02-05
Switch to cpythons file object implementation

undingen:
Switches to the cpython file object implementation and printing code.

This makes us pass the following cpython tests:
- test_file2k
- test_file_eintr
- test_print
- test_softspace

This also make use now call much more often into ```tp_print``` which we do not implement for many types. I think in the future we should add them because this functions can directly write to a file descriptor instead of having to always write into a buffer. I had to add it for the dict type because the default dict implementation requires it.

In addition fixes a pyopenssl integration test error:
```
ERROR: Random data can be saved and loaded to files with paths specified as
----------------------------------------------------------------------
  File "/home/marius/pyston/build/Release/from_cpython/Lib/unittest/case.py", line 329, in run
    testMethod()
  File "/home/marius/pyston/pyopenssl_test_env_pyston_release/site-packages/OpenSSL/test/test_rand.py", line 219, in test_unicode_paths
    self._read_write_test(path)
  File "/home/marius/pyston/pyopenssl_test_env_pyston_release/site-packages/OpenSSL/test/test_rand.py", line 184, in _read_write_test
    with open(path, "w"):
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2603' in position 11: ordinal not in range(128)
```


from_cpython/Modules/posixmodule.c
```diff
@@ -6916,8 +6916,7 @@ posix_fdopen(PyObject *self, PyObject *args)
     /* We now know we will succeed, so initialize the file object. */
 
     // Pyston change:
```
kmod: Tiny, but we can remove the "Pyston change" comment

kmod:
lgtm!

undingen:
merged with a fix witch increases the travis ci timeout to 30min from 20 because the debug build hit the limit.

PR #1091, opened 2016-02-07
switch to cpython traceback implementation

undingen:
The performance is the same but it removes quite a lot of unnecessary code and makes hopefully maintaining our codebase a little easier.


undingen:
ready for review :-)

kmod:
:+1:

PR #1093, opened 2016-02-09
socket fixes

undingen:
The socket implementation used to only call close on a socket when all references where gone.
This works fine for cpython which uses reference counting but for pyston (and pypy) this is a problem.
This works around the problem in a similar way as pypy does.
It also uses the same method names so that 3th party libs which already support pypy will automatically work with pyston too.
(e.g. urllib3)

I would prefer if we get ref counting but meanwhile this improves the situation a lot. (While I'm not sure if this does not break other stuff but I would prefer to not spend to much time on this now if it may get replaced soon)


kmod:
Awesome :)

PR #1094, opened 2016-02-10
Make the pycrypto integration test run the testsuite

undingen:
We did not yet run the whole testsuite but instead just ran a simple test we wrote.

### DIFF
--- a/test/CPYTHON_TEST_NOTES.md
+++ b/test/CPYTHON_TEST_NOTES.md
@@ -91,8 +91,6 @@ test_email              [unknown]
 test_enumerate          assert instead of exception in BoxedEnumerate
 test_exceptions         we are missing recursion-depth checking
 test_extcall            f(**kw) crashes if kw isn't a dict
-test_file2k             we abort when you try to open() a directory
-test_file_eintr         not sure
 test_fileio             [unknown]
 test_fork1              [unknown]
 test_frozen             [unknown]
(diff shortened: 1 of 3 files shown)


undingen:
Ok so the problem is that inside ```_fastmath.c``` we have the following code:
```c
    /* Randomly choose X, y[0] and y[1] */
	Py_BLOCK_THREADS;
	res = 1;
	res &= getRandomRange (X, lower_bound, upper_bound, randfunc);
	res &= getRandomNBitInteger (y[0], 101, randfunc);
	res &= getRandomNBitInteger (y[1], 101, randfunc);
	Py_UNBLOCK_THREADS;
	if (!res)
	{
		error = 1;
		goto cleanup;
	}
```

It looks like the first call to ```getRandomRange``` fails and sets a python error and returns 0 (0 means error). The code will than still call into ```getRandomNBitInteger``` before it will abort and jump to cleanup which will return NULL to the caller.
The problem is that we assert if there is already an error set (which makes sense because this piece will always return the last set python error or maybe even fail to report the correct ptyhon error if the error get's cleared)
How should I fix it?
Should I add another patch to pycrypto which jumps to the cleanup after the first error?. Just wondering because this functions is called ```getStrongPrime``` which sounds very important and I don't want to break it.

kmod:
Hmm I'm a bit afraid of changing this too -- we might be changing how much `randfunc` gets called.  But then again, I feel like the expectation would be that if `randfunc` throws an exception then it wouldn't get called again?  I don't think there's anything you can reasonably expect when continuing with an exception still set.  FWIW we have some cover here from the [official docs](https://docs.python.org/2/c-api/exceptions.html):
> If the error is not handled or carefully propagated, additional calls into the Python/C API may not behave as intended and may fail in mysterious ways.

I think it's fine if we change this to
```
if (res)
  res &= getRandomNBitInteger (y[0], 101, randfunc);
```

it might also be ok now to remove the asserts that check to make sure we don't already have an exception set.

undingen:
ok cool I will add the additional if checks because I think currently having this assert in is quite nice at least it showed some errors I think...

PR #1098, opened 2016-02-11
add the _locale and cPickle modules

undingen:
- add the ```_locale``` module
- add the ```cPickle``` module
the module is much faster than the python implementation (in a small benchmark by about 4x)
But I had to change the code a little bit because of our GC and more importantly because some of our types have different names etc...
I also noticed that we failed to to set ```capifunc.__module__``` in a lot of cases which made pickle error
- together this makes us run 7 more tests :-)


kmod:
nice!

Issue #1096, opened 2016-02-10
Provide a docker image to make it easy to try Pyston


kmod:
Or maybe some other tool (Vagrant was suggested at one point).  This gets asked for fairly regularly and I think is appreciated in other projects when available.

We're happy to help out from the Pyston side but if anyone with some experience setting up docker containers wants to give this a go, it would really helpful :)

corona10:
I simply made for `v0.4` 
You can pull image by `docker pull corona10/pyston:0.4`
And also here is the [link](https://hub.docker.com/r/corona10/pyston/tags/)

corona10:
but in my opinion we can do better job with automated build for provide `latest pyston docker image`

jakirkham:
+1 for a docker image. Should be pretty easy considering this should already build on Ubuntu (version? precise?). Could just use a base image that starts with that version of Ubuntu and then apt-get anything else that is needed.

aisk:
I used Vagrant to build Pyston for a while, this is the Vagrant file I used: https://gist.github.com/aisk/efd38d84575ed7087280

I think we can provide docker and vagrant envs both to let people choose which they want, and macitence both of them seems not a hard work.

funkyfuture:
for what [i can tell](https://hub.docker.com/r/funkyfuture/nest-of-serpents/builds/be8iamnrpjjkhnkkhopf9cc/) an automated build seems out of its resources limits (2h and afair 1GB RAM, 1 CPU and 30GB drive).

i suppose a debian package would be a good idea anyway. if that dependency on `texlive-extra-utils` with its vast amount of dependencies wouldn't be required a build may fit within the limits.

jakirkham:
Congrats on your new company! Glad to see this project moving forward 😄 🎉

kmod:
Thanks @jakirkham !

PR #1099, opened 2016-02-11
Add dockerfile

corona10:
#1096 Simply make develop environment by using docker
- `docker build -no-cache --tag pyston:latest ./` for make image
- `docker run -i -t pyston:lates` for run image

p.s May be we can provie daily image from `https://hub.docker.com/` by using `AUTOMATED BUILD` but need some test (As far as I know there are some cache issue. this script should always clone latest branch but dockerfile may skip if there is same instruction already executed. I don't know whether `AUTOMATED BUILD` support `-no-cache` option.)


kmod:
So right before the 0.5 release I looked into docker stuff and ended up adding a Dockerfile -- I took a bit of a different approach with it, where it downloads the latest release instead of following our build-from-scratch instructions.  This seems to be the common way that other projects handle it (at least, that's how CPython and PyPy's work).

It might be nice to have a dockerfile that generates a dev environment.  I think docker isn't really meant for this sort of thing (it doesn't want you to keep any state or treat the containers like vms), but I've definitely used it like that and it seems to work fine.  So in that sense this could still be useful, but I'm worried about the download size -- the llvm directory and build directory add up to several gb.  Maybe for this use-case ("quickly setting up a Pyston dev environment") we should go with Vagrant and try to take #1104 / #1147 to completion? (It looks like #1104 is just missing docs.)

Also, we should move this Dockerfile out of the top level, since `docker build` will import everything from the Dockerfile's directory into the container.

funkyfuture:
fyi, [here](https://github.com/funkyfuture/docker-nest-of-serpents/blob/pyston/Dockerfile) is a one-layer image based on Ubuntu Xenial as reference.

i used the build-approach as the virtualenv doesn't fit into my concept. a virtualenv is also rather a redundant isolation layer in a container.

the build works well, but the vast amount of package dependencies - including lots of latex related stuff - is hell.

kmod:
I'm going to close this because I think it is quite unlikely that a build-based dockerfile will be what we want -- as you said it will prevent automatic builds, and there isn't much benefit to it.  Separating the packaging from the building seems reasonable as well.

I don't think we can get around using virtualenv inside the container, at least if we want to provide a "Pyston environment" and not just a "Pyston binary".  Yes it's multiple layers of isolation, but docker containers don't remove the need to separate app and system pythons.  Maybe there's some better way of solving it, but providing a virtualenv makes it so that we can address the common use case of people who want to run numpy on Pyston.

Issue #1100, opened 2016-02-13
Directly use cpythons errors.c file


undingen:
We have copied nearly all of the functions inside this file directly in various files inside pyston.
It would be nice to clean up this mess by removing this implementations and instead directly using the ```from_cpython/Python/errors.c``` file (does not exist yet must be copied from cpython 2.7.7).
But some caution must be taken because some functions (may) have small modifications which must be preserved if they are necessary.

PR #1101, opened 2016-02-14
Directly use cpythons errors.c file

corona10:
Accoring to #1100 I change some files.


undingen:
oh wow that fast fast, thanks for working on this :-)
I think we don't need the declarations inside the errors.cpp file they should already be declared inside the ```pyerrors.h``` header. 
And I noticed that there is a small difference inside the ```PyErr_NormalizeException``` function please copy the change over to the new errors.c file (but in a separate commit so that we have the original version as the first commit and the changes to the file in another one)
Thanks!

corona10:
@undingen Thanks. I fix them as you reviewed.

undingen:
thanks for the patch!

PR #1103, opened 2016-02-17
check duplicate function params

aisk:


src/codegen/irgen/hooks.cpp
```diff
@@ -692,11 +692,14 @@ Box* execfile(Box* _fn, Box* globals, Box* locals) {
 
 Box* execMain(Box* boxedCode, Box* globals, Box* locals, PyCompilerFlags* flags) {
     if (PyTuple_Check(boxedCode)) {
-        RELEASE_ASSERT(!globals, "");
-        RELEASE_ASSERT(!locals, "");
+        if (globals || locals) {
```
undingen: unfortunately I think this error messages are wrong :-(
I have seen that our whole exec/eval handling has quite a lot of smaller missing parts/differences with cpython and feels like it's not really finished. I would really prefer if we soon switch to a implementation which uses more of cpythons code. Would this be something which interests you? Otherwise I would take on the task

undingen:
Thanks for working on this :-), unfortunately as I wrote in the comment I feel like current exec/eval implementations needs an update because there is several unimplemented stuff...
I feel like it makes more sense to directly switch to an implementation which is uses more of cpythons code or which is at least more similar to it instead of incrementally adding new fixes. Would this be something which interests you? Otherwise I would take on the task

aisk:
@undingen can you give me some more info about that? I think what to do is copy the  CPython's `Python/bltinmodule.c` file to Pyston repo and comment the unused code and make it compiled?

undingen:
I tried to come up with a more detailed description but found out that it's not easy to describe what should be left and what should get used because this code is quite complicated... :-( I decided to implement it see #1111. It's still in WIP state but after I finished it we should be able to merge in your duplicate function param check :-).
Sorry for recommending it as first as a task and than doing it on my own but I underestimated how much familiarity with the code base this change requires.

aisk:
@undingen never mind, I'll remove the compile related changes in this PR 😁

undingen:
So getting back to your PR because my eval and exec cleanup PR is merged :-)
Seems it missed duplicate varags and kwargs:
```eval('lambda a,*a,**a:0')```
and could you please use our ```RELEASE_ASSERT(0, "")``` instead of ``assert```
otherwise looks good :-)

undingen:
Thanks for your patch! :+1:

PR #1105, opened 2016-02-17
fix import

undingen:
this fixes some smaller differences we had and adds some previously unsupported stuff to the imp module
(we could not load C extensions and were too pedantic about some args before, ...)
In addition this gives us:
 - import lock
 - we load now a lot of the builtin modules lazely
 - support for printing out which files get loaded in verbose mode
 - we can now delete a builtin module from sys.modules and import it again (some cpython tests use this)

It also made a ```classobj.__delattr__``` problem appear


undingen:
should now be ready for review :-)

kmod:
I figured out what the delattr issue was -- it's an issue with the method-cache and how it interacts with our startup :)  What was sometimes happening was that it would look up cache the value of `classobj.__delattr__` before we had finished fully setting up the object, and then one of the CPython initialization functions (`add_operators()`) will add more attributes but not call PyType_Modified.  I guess CPython only calls add_operators() from places where it wouldn't need to call PyType_Modified(), but that's not the case for us.  So if you get (un)lucky and the cache is still valid for `classobj.__delattr__`, it might end up pulling the wrong value (the one that wraps the generic delattr).  But depending on the code you run beforehand the cache will usually get overwritten and it often won't matter :P
Anyway, I think #1106 is an ok solution to this.

undingen:
Oh cool yeah that sounds like the problem! 
Cheers for finding the root cause so fast :-)

PR #1106, opened 2016-02-18
Call PyType_Modified() in add_operators()

kmod:
Sometimes we started filling in the method-cache before calling add_operators.
add_operators directly modifies the class objects, so it needs to signal that
the cache is now invalid.

I think CPython avoids needing this since they never call add_operators
after doing any type lookups, but we are a bit more flexible with our initialization.


undingen:
LGTM, seems like the root cause of the issue I was seeing

PR #1109, opened 2016-02-21
implement correct binop lookup for subclasses

undingen:
If the rhs is a subclass of the lhs, call into the reverse op first.
Performance looks unaffected or only very minor slowdown.
@Daetalus confirmed that this is fixing the numpy problem he had.


undingen:
ready for review :-)

kmod:
:+1:

PR #1110, opened 2016-02-22
add reload() and a integration test for simplejson

undingen:
in addition I sneaked in a commit which switches the libunwind git repo to a mirror hosted on github by pathscale because the gnu server is down very often.


kmod:
:+1:

PR #1112, opened 2016-02-23
add GeopIP integration test and use cpythons do_mkvalue and fix some problems of test_builtin.py

undingen:
Unfortunately there are still a few hiding which means we can't yet run test_builtin.py right now
but after this and #1111 there shouldn't be that many problems left.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ addons:
     - g++-4.8
     - gdb
     - libbz2-dev
+    - libgeoip-dev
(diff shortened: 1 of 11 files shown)


undingen:
Oh yeah good idea, forgot about the docs...

undingen:
I merged this in with an additional commit https://github.com/undingen/pyston/commit/81d61567662db363debe01c49b46cf5a9bd48168 which makes enumerate() support longs

PR #1114, opened 2016-02-25
fixes for CFFI and add an integration test for it

undingen:
We can now run most of the CFFI tests.
Some of the changes were required because CFFI tests loading the same shared library several times from different locations.

The CFFI patch introduces a memory leak but I think this is okay for now because we will soon be able to remove it when we switch to ref counting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -337,7 +337,7 @@ endif()
 add_pyston_test(defaults cpython --exit-code-only --skip-failing -t100)
 add_pyston_test(defaults integration --exit-code-only --skip-failing -t600)
 if(ENABLE_EXTRA_TESTS)
-  add_pyston_test(defaults extra -t600 --exit-code-only)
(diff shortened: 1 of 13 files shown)


kmod:
lgtm (minus the build errors -- we have a test that looks at everything in types.py and we need to add this new type to the filter list)

undingen:
Unfortunately the CFFI test causes crashes on travis CI e.g. https://travis-ci.org/dropbox/pyston/jobs/111966763
I can't reproduce it on a non travic CI machine and it's very similar to the ones we encounter already for the ```cpython/test_uuid```  and ```cpython/test_ascii_formatd``` which we skip currently (crashes are all related to ctypes)

undingen:
I merged it with the ctypes unaligned pointer workaround (commented out the ```long double```)
added enabled the disabled ctypes test.
I had to also to let the cffi test differentiate between clang and gcc because we are getting more errors using clang...

PR #1115, opened 2016-02-25
Address minor comment from #1028

undingen:
and add float() and str() param names
In addition noticed that we did not run one of our small numpy tests


undingen:
I unfortunately had to remove the numpy ```test_abc``` run because it caused crashes in release mode. But seems to be GC related (test on it's own ran too but running several numpy tests crashed it)

PR #1117, opened 2016-02-27
add support for custom traceback entries

undingen:
tracebacks: implement PyFrame_New and PyCode_New / custom traceback entries from the CAPI

- it only allows creating empty code objects but this is enough to allow ctypes and cython to create custom traceback entries.
- It's not enough for removing the pyexpat and all cython code object workarounds (because some paths need 'real' code object / frames)
- I also changed ```ExcInfo``` to use ```NULL``` instead of ```None``` for a unset traceback which makes it more similar to cpython and allows ```PyTraceBack_Here``` to work correctly

### DIFF
--- a/from_cpython/Include/Python.h
+++ b/from_cpython/Include/Python.h
@@ -74,7 +74,6 @@
 #include "classobject.h"
 #include "cobject.h"
 #include "fileobject.h"
-#include "frameobject.h"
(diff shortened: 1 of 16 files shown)


undingen:
ready for review

kmod:
Nice!

PR #1120, opened 2016-03-08
Threading and multiprocessing related fixes

undingen:
- call threading._after_fork and threading._shutdown, return real thread id
- fix attrwrapper of an object which becomes dict backed and support non str key types
_I tried to merge the attrwrapper code with our dict code. But sadly this introduced quite some slowdown which make me switch back to this minimal fix._
- fix thread.error pickling


undingen:
should be ready for review

PR #1127, opened 2016-04-01
add refcouting annotation to zip builtin function

Daetalus:


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -1242,6 +1241,7 @@ Box* zip(BoxedTuple* containers) {
             el->elts[i] = *iterators[i];
             ++(iterators[i]);
         }
+        AUTO_DECREF(el);
```
kmod: This is minor, but this should go right after the `BoxedTuple::create` call.  If any of the iterators throw an exception (their `++` can do that) then we would want to free the tuple.

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -1786,14 +1786,19 @@ Box* builtinCmp(Box* a, Box* b) {
...
-        args = PySequence_Tuple(args);
+Box* builtinApply(Box* func, Box* _args, Box* keywords) {
+    Box* args;
+    if (!PyTuple_Check(_args)) {
+        if (!PySequence_Check(_args))
+            raiseExcHelper(TypeError, "apply() arg 2 expected sequence, found %s", getTypeName(_args));
+        args = PySequence_Tuple(_args);
         checkAndThrowCAPIException();
+    } else {
+        args = _args;
```
kmod: I think this needs to be `args = incref(_args)`.  Otherwise looks good!

PR #1129, opened 2016-04-04
small refcounting fixing in tuple

Daetalus:


kmod:
For `BoxedTuple::resize`, I think when we shrink the tuple we need to decref any elements that get dropped.  CPython has this code:

```C
    for (i = newsize; i < oldsize; i++) {
        Py_CLEAR(v->ob_item[i]);
    }
```

Otherwise looks good!

PR #1131, opened 2016-04-06
generate a decref info table

undingen:
we need this map to know which object refcounter we have to decrement in case an exception gets throwen inside an IC or the bjit.

this is the most simplistic way I come up with. 
I'm happy for name suggestions and a place where to put it...


src/asm_writing/rewriter.cpp
```diff
@@ -1264,6 +1288,10 @@ void RewriterVar::refConsumed() {
     last_refconsumed_numuses = uses.size();
 }
 
+bool RewriterVar::needsDecref() {
+    return reftype == RefType::OWNED && !this->refHandedOff();
```
kmod: I think it's possible for us to have multiple refs that we would need to remove from an object -- right now the rewriter emits all the increfs as early as possible, rather than emitting them just when they're needed.  So I think we either need to change the rewriter to emit the increfs on-demand, or to keep track of how many refs it's keeping at any given point (I think the first option sounds easier?).  Or we could just assert that it doesn't add multiple refs since that might not actually happen in practice :P

undingen:
I have updated the code, it allows now the bjit to work too.
But I had to disable the "this variable is dead at the end of this block don't store it inside the vregs"

kmod:
closed by #1144

PR #1132, opened 2016-04-06
add refcounting annotation to some set function

Daetalus:


src/runtime/set.cpp
```diff
@@ -275,19 +275,28 @@ static Box* setRepr(BoxedSet* self) {
...
-
-    for (auto elt : other_set->s) {
-        auto&& p = self->s.insert(elt);
-        if (!p.second /* already exists */) {
-            self->s.erase(p.first);
-            Py_DECREF(elt.value);
-        } else {
-            Py_INCREF(elt.value);
+static void _setSymmetricDifferenceUpdate(BoxedSet* self, Box* _container) {
+    if (PyAnySet_Check(_container)) {
```
kmod: Is there a reason to split this up like this?  I think it's better to keep the old control flow ("first convert the argument to a set, then do the symmetric-difference") since this will matter if iterating over the container throws an exception.  It looks like there was a refcounting issue if we made the new set -- maybe we could fix it with something like
```
if (!PyAnySet_Check(other))
  other = makeNewSet(self->cls, other);
else
  Py_INCREF(other);
AUTO_DECREF(other);
```
ie to make sure that after the if statement `other` always has one owned ref, which we will get rid of with the AUTO_DECREF.

kmod:
Thanks for finding these!

PR #1134, opened 2016-04-12
Ref nexedi fixing

Daetalus:
@kmod Ready to review.


src/runtime/types.cpp
```diff
@@ -1591,6 +1591,7 @@ static void funcSetName(Box* b, Box* v, void*) {
     BoxedFunction* func = static_cast<BoxedFunction*>(b);
 
     if (v == NULL || !PyString_Check(v)) {
+        Py_XDECREF(v);
```
kmod: You're right that `v` needs to get decref'd but I don't think inside this function is the right place -- I think the bug is in  `dataDescriptorSetSpecialCases` (the caller of this function) which forgets to decref `val` in the getset case.

PR #1136, opened 2016-04-18
add refcounting annotation that make test_long pass

Daetalus:


undingen:
:+1: thanks for the change!

PR #1137, opened 2016-04-19
Fix test_math refcounting error

Daetalus:


undingen:
:+1:

src/runtime/float.cpp
```diff
@@ -163,7 +163,10 @@ extern "C" double PyFloat_AsDouble(PyObject* o) noexcept {
         return -1;
     }
 
-    return static_cast<BoxedFloat*>(fo)->d;
+    val = static_cast<BoxedFloat*>(fo)->d;
+    Py_DECREF(fo);
```
kmod: Hi, shouldn't this get covered by the AUTO_DECREF on line 160?

undingen: oh yeah you are right :blush:

Daetalus: Sorry for the mistake... :worried:

kmod: It's not really your fault if we have the tests turned off :P (they are on now)

PR #1139, opened 2016-04-19
cleanup abandon generators

undingen:
in order to break cycles we need to traverse all generator owned object at yields.
The interpreter is fine because it stores all objects inside the vregs so they will already get visited.
For the llvm jit pass all owned object to the yield call as vararg argument.

We will currently leak (move objects to ```gc.garbage```) when a generator is involved in a cycle because our ```PyGen_NeedsFinalizing``` is much more conservative.

This also has the advantage that we are now much better in reusing generator stacks 
e.g. ```django_template3_10x.py``` ```generator_stack_created``` improves from 15528 to 4

### DIFF
--- a/from_cpython/Lib/test/test_abc.py
+++ b/from_cpython/Lib/test/test_abc.py
@@ -1,5 +1,3 @@
-# expected: reffail
-# - generator abandonment
 # Copyright 2007 Google, Inc. All Rights Reserved.
 # Licensed to PSF under a Contributor Agreement.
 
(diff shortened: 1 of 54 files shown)


undingen:
should be ready for a review.
we still leak memory when the generator is involved in a cycle but I think this should get resolved in another PR.

kmod:
How would you feel about determining "isBorrowedOnlyType" from the name of the variable rather than the type?  otherwise looks great, really excited for this!

undingen:
merged this and addressed the comment by checking for ```PASSED_GENERATOR_NAME```

PR #1140, opened 2016-04-22
Misc refcounting leaks encountered while fixing the generator abandonment tests

undingen:
fixes leaks in
- interactive mode
- -m mode
- when calling ```tp_setattr``` from ```setattr()```
- creating duplicate set and dict entries from an AST node

I could not enable the tests yet because the need the generator abandonment fix but this fix is not ready yet but I want to land this fixes now so we don't do duplicate work.

### DIFF
--- a/src/codegen/baseline_jit.cpp
+++ b/src/codegen/baseline_jit.cpp
@@ -854,7 +854,9 @@ Box* JitFragmentWriter::callattrHelper(Box* obj, BoxedString* attr, CallattrFlag
 Box* JitFragmentWriter::createDictHelper(uint64_t num, Box** keys, Box** values) {
     BoxedDict* dict = (BoxedDict*)createDict();
     for (uint64_t i = 0; i < num; ++i) {
-        dict->d[keys[i]] = values[i];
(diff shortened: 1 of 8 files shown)


src/codegen/baseline_jit.cpp
```diff
@@ -870,8 +872,13 @@ Box* JitFragmentWriter::createListHelper(uint64_t num, Box** data) {
 
 Box* JitFragmentWriter::createSetHelper(uint64_t num, Box** data) {
     BoxedSet* set = (BoxedSet*)createSet();
-    for (int i = 0; i < num; ++i)
-        set->s.insert(data[i]);
+    for (int i = 0; i < num; ++i) {
+        auto&& p = set->s.insert(data[i]);
+        if (!p.second /* already exists */) {
+            Py_DECREF(p.first->value);
+            *p.first = data[i];
```
kmod: I think this should use the original value and decref the duplicate one: `set([1, 1L])` is `1` for CPython and `1L` for us.

kmod: This matches the behavior for dict keys, but is different from dict values -- `{1:1, 1L:1L}` is `{1: 1L}`.

PR #1142, opened 2016-04-25
Misc refcounting leaks encountered while fixing the generator abandonment tests part 2

undingen:
this issues all got triggered by cpython test we currently have disabled.

Generally I'm not really sure what pattern to use (and I'm very inconsistent with it) when a value needs to get decrefed incase a exception get's throwen but else returned. 
- create a ```AUTO_DECREF(x)``` and do a ```return incref(x)```
- use a try catch block and only decref in the catch (probably fastest way to do it in case the exception is very unlikely but also verbose)
- use something like ```ExceptionCleanup``` which decrefs it in the destructor when an exception gets thrown...

### DIFF
--- a/src/runtime/set.cpp
+++ b/src/runtime/set.cpp
@@ -24,12 +24,19 @@ extern "C" Box* createSet() {
     return new BoxedSet();
 }
 
-void _setAddStolen(BoxedSet* self, STOLEN(BoxAndHash) val) {
+static void _setAddStolen(BoxedSet* self, STOLEN(BoxAndHash) val) {
     auto&& p = self->s.insert(val);
     if (!p.second /* already exists */) {
         // keep the original key
         Py_DECREF(val.value);
     }
+}
+
+void _setAddStolen(BoxedSet* self, STOLEN(Box*) val) {
+    AUTO_DECREF(val);
+    BoxAndHash val_hashed(val); // this can throw!
+    incref(val_hashed);
+    _setAddStolen(self, val_hashed);
 }
 
 namespace set {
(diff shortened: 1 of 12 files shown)


src/runtime/set.cpp
```diff
@@ -118,22 +118,29 @@ BoxedSet* makeNewSet(BoxedClass* cls, Box* container) {
...
+                for (auto&& elt : ((BoxedSet*)container)->s) {
+                    rtn->s.insert(incref(elt));
+                }
+            } else if (PyDict_CheckExact(container)) {
+                for (auto&& elt : ((BoxedDict*)container)->d) {
+                    rtn->s.insert(incref(elt.first));
+                }
+            } else {
+                for (auto elt : container->pyElements()) {
+                    AUTO_DECREF(elt);
```
kmod: I think the protocol should usually be that a ref-stealing function steals the ref even if an exception is thrown.  Which I think means here that maybe we should make this change inside _setAddStolen itself?

kmod:
Yeah I'm not sure either :/  I tend to do the `AUTO_DECREF(r);  ...;  return incref(r);` approach since it's the cleanest, but it definitely feels wasteful to use.

src/runtime/set.cpp
undingen: the problem is not that ```_setAddStolen``` is throwing but the implicit call of the ```BoxAndHash``` constructor (inside the ```PyHasher()``` call) is so the exception happens before we are inside ```_setAddStolen```. Maybe I should create a overloaded ```_setAddStolen(BoxedSet* self, STOLEN(Box*) val)``` which handles this case?
In general it's quite dangerous that ```BoxAndHash(box)``` can throw while we are implicit calling it in a lot of places :-(.

undingen:
I change the patch to use the ```AUTO_DECREF(r); ...; return incref(r);``` approach and added a ```_setAddStolen(BoxedSet* self, STOLEN(Box*) val)``` which will steal the reference when ```BoxAndHash()``` throwes.

PR #1144, opened 2016-04-28
Fix + reenable the bjit

kmod:
Continuation of #1131


src/codegen/baseline_jit.cpp
```diff
@@ -342,9 +354,12 @@ RewriterVar* JitFragmentWriter::emitCreateTuple(const llvm::ArrayRef<RewriterVar
         r = call(false, (void*)BoxedTuple::create2, values[0], values[1])->setType(RefType::OWNED);
     else if (num == 3)
         r = call(false, (void*)BoxedTuple::create3, values[0], values[1], values[2])->setType(RefType::OWNED);
-    else
+    else {
         r = emitCallWithAllocatedArgs((void*)createTupleHelper, { imm(num), allocArgs(values) }, values)
                 ->setType(RefType::OWNED);
+        for (auto a : values)
```
undingen: I think this should not be needed because this was the reason I introduced ```emitCallWithAllocatedArgs``` over just using calling ```call()```. But maybe we should remove ```emitCallWithAllocatedArgs``` and instead use your ```refUsed()``` mechanism.

PR #1145, opened 2016-04-29
fix set.add() for existing keys and fix set ast node when encountering keys with same hashes

undingen:
we have to insert the elements in reverse in order to replicate cpythons behaviour.
cpython is pushing the temps on it's value stack and than inserting them ona after another by poping them from the stack


kmod:
oh crazy :)

PR #1148, opened 2016-05-01
add docker

KrishMunot:


kmod:
This looks great, but would you mind adding some sort of doc on how to use it?

KrishMunot:
Yes, I will add it soon. 
Can you merge this while we are at it? 
Thanks! :smiley:

corona10:
@kmod It looks like same docker code when I made PR as #1148 but it looks good. :-)

kmod:
Hi @KrishMunot, would you mind cherry-picking corona10's original commit, rather than copying the file and creating a new commit?  You should be able to do it with
```
git fetch origin eae40ee9
git cherry-pick eae40ee9
```

Could you also add a small doc somewhere?  I'm pretty hesitant to merge something like this without at least a tiny amount of documentation, since otherwise we/I won't remember what's going on with it and it is likely to just get removed.  Even just corona10's two lines of instruction from #1099 would be great to have in the readme.

kmod:
I'm going to close this and we can keep discussing on #1099

PR #1149, opened 2016-05-01
URL alive

KrishMunot:


undingen:
We have this copyright notice in lots of files so we would need to change a lot of additional files too. :disappointed: 
But I don't think we should do it because it don't really get what this is helping with (I think not even github makes it into a link when it's inside source code).
Hope this does not sound discouraging, we have a lot of stuff which needs improvements and we are really looking forward to people joining us in improving pyston :smile:  (if you are looking for a project contacting us in our gitter channel is probably best).

KrishMunot:
Thanks @undingen for the heads up! I will look into better contributions next time. 
Can you send me the link to the gitter chat? I would love to enagage and ask my questions there.
Thanks! :smile:

kmod:
Come join us at https://gitter.im/dropbox/pyston :)

PR #1151, opened 2016-05-01
Fix the test_string segfault error

Daetalus:


undingen:
it looks like this patch is just working around the root-cause of the crash.
I think you should instead in ```BoxedMethodDescriptor::tppCall``` change ```Box* oargs_array[1]``` to ```Box* oargs_array[1] = { NULL }```. To not make ```rearrangeArguments``` decref uninitialized memory.

undingen:
:+1:  thanks for the patch!

PR #1152, opened 2016-05-03
Fix another raise-before-stealing-argument bug

kmod:
I'm not sure if this is actually triggerable or not


kmod:
At least on clang, the BoxAndHash gets evaluated before the getattrInternal(), so I think this was not triggerable.  But I think we got bit by something like this before when gcc evaluated the arguments in a different order.

PR #1153, opened 2016-05-04
Add some liveness information to the cfg

kmod:
to help clear out temporaries (ie subexpressions) in a more timely fashion.

### DIFF
--- a/src/analysis/function_analysis.cpp
+++ b/src/analysis/function_analysis.cpp
@@ -136,8 +136,13 @@ public:
     bool visit_name(AST_Name* node) {
         if (node->ctx_type == AST_TYPE::Load)
             _doLoad(node->id, node);
-        else if (node->ctx_type == AST_TYPE::Store || node->ctx_type == AST_TYPE::Del
-                 || node->ctx_type == AST_TYPE::Param)
+        else if (node->ctx_type == AST_TYPE::Del) {
+            // Hack: we don't have a bytecode for temporary-kills:
+            if (node->id.s()[0] == '#')
+                return true;
+            _doLoad(node->id, node);
+            _doStore(node->id);
+        } else if (node->ctx_type == AST_TYPE::Store || node->ctx_type == AST_TYPE::Param)
             _doStore(node->id);
         else {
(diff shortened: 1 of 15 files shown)


undingen:
I like this :-)
I think this enables us in the future to reduce the number of vregs required a lot with only a small change inside ```assignVRegs()``` by keeping track which names only live inside a BB and allowing the next BB to reuse them without having to calculate whole liveness info.
Because it looks like a lot of the compiler generated names only have a very short live inside a BB:
```
Block 54; Predecessors: 51 53 52 Successors: 55 56
    #75 = not (#71)
    if :NONZERO(#75<k>) goto 55 else goto 56
Block 55 'iftrue'; Predecessors: 54 Successors:
    #76 = CodecRegistryError
    #77 = mod
    #78 = #77<k>.__name__
    #79 = mod
    #80 = #79<k>.__file__
    #81 = (#78<k>, #80<k>)
    #82 = "module "%s" (%s) failed to register"%#81<k>
    raise #76<k>, #82<k>
Block 56 'iffalse'; Predecessors: 54 Successors: 58 59
    #84 = hasattr
    #85 = entry
    #86 = #85<k>[0]
    #87 = #84<k>(#86<k>, "__call__")
    #88 = not (#87<k>)
    #83 = #88
    if :NONZERO(#88<k>) goto 59 else goto 58
```

kmod:
Agreed, it will be great if we can do that :)  There's some more work required for that -- it seems like it would be straightforward, since we know when we introduce temporaries and we know when they die.  But the way the cfg code is set up, it doesn't really know where it ends up putting kill-flags, so it doesn't know when it's safe to reuse the name.  It's definitely a solvable problem, but I think it would require some rethinking of the cfg code (or a hacky addition) and I wanted to hold off on that now.  Also because right now the failure mode is that you get an error message like `#3 referenced before being defined`, but if we reused temporary names you might end up getting worse kinds of incorrect behavior.

undingen:
Just took a brief look at lxml again: this patch fixed the problems I was seeing :+1:  we are now down to 3 failures and 2 errrors from 1381 tests :-)

PR #1157, opened 2016-05-06
refcounting changes for our integration tests

undingen:
this fixes and reenables some of our integration tests


undingen:
there are still a few problems with the integrations tests namely cffi fails and sqlalchemy triggers an assert which I will look into tomorrow.
I also had to disable member_descriptor_ics.py because it fails in a release build.

undingen:
this is ready for review
there are still some problems with cffi and sqlalchemy :-( which I plan to fix next week but I think it's better get this one in in order to not block other peoples work (e.g. numpy) and create a new PR for the remaining stuff.
```django_test.py``` is leaking which need also more investigation. (it only passes because I disabled the total refcount is 0 test)

concerning the ```RegionAllocatorAdaptor``` adaptor I removed:
at least with my stl implementation I could not trigger it and it looks like it does not implement whole allocator trait, so I guess it never really worked and I removed it.

File: Makefile [hunk 1]
```diff
@@ -655,7 +655,7 @@ $(CMAKE_SETUP_DBG):
 	@$(MAKE) cmake_check
 	@$(MAKE) clang_check
 	@mkdir -p $(CMAKE_DIR_DBG)
-	cd $(CMAKE_DIR_DBG); CC='clang' CXX='clang++' cmake $(COMMON_CMAKE_OPTIONS) -DCMAKE_BUILD_TYPE=Debug
+	cd $(CMAKE_DIR_DBG); CC='clang' CXX='clang++' cmake $(COMMON_CMAKE_OPTIONS) -DCMAKE_BUILD_TYPE=Debug -DENABLE_REF_DEBUG
```
kmod: I think we also need to update tools/configure_travis.sh if we want to get the reference checks in CI.

src/runtime/capi.cpp
```diff
@@ -1872,7 +1872,8 @@ Box* BoxedCApiFunction::tppCall(Box* _self, CallRewriteArgs* rewrite_args, ArgPa
         rewrite_args->out_success = true;
     }
 
-    checkAndThrowCAPIException();
+    if (!rtn)
+        checkAndThrowCAPIException();
```
kmod: tiny, but this should be throwCAPIException() (in case the user returned NULL but didn't set an exception).

src/runtime/exceptions.cpp
```diff
@@ -203,6 +203,8 @@ extern "C" void raise3(STOLEN(Box*) arg0, STOLEN(Box*) arg1, STOLEN(Box*) arg2)
         startReraise();
 
     assert(!PyErr_Occurred());
+    if (PyErr_Occurred())
```
kmod: These lines seem very odd when combined with the assert above them :)  Maybe we should just get rid of the assert if we can support this case reasonably?

src/runtime/import.cpp
```diff
@@ -174,6 +178,13 @@ BoxedModule* importCExtension(BoxedString* full_name, const std::string& last_na
     if (Py_VerboseFlag)
         PySys_WriteStderr("import %s # dynamically loaded from %s\n", full_name->c_str(), path.c_str());
 
+#ifdef Py_REF_DEBUG
+    // if we load a foreign C extension we can't check that _Py_RefTotal == 0
+    if (!llvm::StringRef(path).endswith("from_cpython/Lib/" + last_name + ".pyston.so"))
+        imported_foreign_cextension = true;
+    imported_foreign_cextension = true; // HACK: disable it always for now
```
kmod: Interesting, why do we need to disable this?

kmod:
That's some pretty awesome improvements in library compatibility :)

src/runtime/import.cpp
undingen: unfortunately ```django_test.py``` is leaking and only works because of this.
I'm investigating it.

undingen:
I resolved the comments and could reenabled the leak check again now that https://github.com/dropbox/pyston/pull/1173 is merged which fixed ```django_test.py```.

cffi and sqlalchemy are still disabled but are next on my list of things to fix

PR #1159, opened 2016-05-06
Fix the lost-ref-from-oargs issue

kmod:
And update a bunch of test comments


undingen:
cool, thanks for fixing this :-)

PR #1162, opened 2016-05-08
bitset begin() function fix

corona10:
It was difficult to understand the intent of the legacy code.
So I tried to guess the original intent.


kmod:
Oh man, how did that ever work :)  How did you run into this?

corona10:
@kmod It was really confusing to understand  how it works when I read this :-) Anyway good is good

PR #1165, opened 2016-05-09
Another run-arbitrary-code-during-unwinding issue

kmod:
This time it was from trying to look at the result of isUnwinding()
to determine if the function terminated successfully or via an exception.
(And in the exception case, to decref some values that would have been returned.)

This got tripped up in a case where we through an exception, this triggers
an AUTO_DECREF which takes an object to zero decrefs, which has a custom destructor,
and calls code that ends up calling isUnwinding().  This is all happening from inside
a C++ destructor, so this counts as isUnwinding().

We had some code to try to keep isUnwinding() meaning what we wanted to use it for,
but it was only for certain cases.  I think we might be able to extend it to more cases,
but it seems like that would be a losing battle since it's hard to say when "unwinding"
ends and then we are suddenly not unwinding.

So instead, I just disabled isUnwinding() (except in debug mode where it can be somewhat
useful for writing "assert(foo || isUnwinding())"), and then used a different way to
try to determine if the function exited normally.

### DIFF
--- a/from_cpython/Lib/test/mapping_tests.py
+++ b/from_cpython/Lib/test/mapping_tests.py
@@ -69,8 +69,9 @@ class BasicTestMappingProtocol(unittest.TestCase):
         #cmp
         self.assertEqual(cmp(p,p), 0)
         self.assertEqual(cmp(d,d), 0)
-        self.assertEqual(cmp(p,d), -1)
-        self.assertEqual(cmp(d,p), 1)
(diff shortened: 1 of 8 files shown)


undingen:
oh wow now I only realize the whole problem.
 so yeah at least now we don't have 3 different approaches anymore for ``` AUTO_DECREF(r); ...; return incref(r);``` :-D

PR #1167, opened 2016-05-09
copy CPython sys.getsizeof implementation with some Pyston changes

Daetalus:
This just copy the CPython implementation, so some NumPy tests could pass.
But most of types in Pyston didn't implement '__sizeof__' attribute yet.

### DIFF
--- a/src/runtime/builtin_modules/sys.cpp
+++ b/src/runtime/builtin_modules/sys.cpp
@@ -408,6 +408,72 @@ extern "C" BORROWED(PyObject*) PySys_GetModulesDict() noexcept {
     return getSysModulesDict();
 }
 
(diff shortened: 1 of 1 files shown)


kmod:
Looks good :)  If you get a chance it would be nice to have a test for this; it sounds like the numpy test exercises it but it would be nice to have a simpler case in test/tests.py.

Daetalus:
Yeah, in theory, the new added code always should with tests. But due to Pyston's types didn't implement `__sizeof__` attribute yet. So maybe I can add test file when we add this attributes to Pyston types?

undingen:
We have some types which implement it currently already e.g. unicode type or the array from the array module. So you could use this.

PR #1170, opened 2016-05-11
some minor NumPy fixing

Daetalus:
Without `str.__rmod__`, `"%s" % numpy.string_` will force NumPy call `numpy.reminder`. This is horrible.

### DIFF
--- a/src/runtime/str.cpp
+++ b/src/runtime/str.cpp
@@ -1164,6 +1164,13 @@ extern "C" Box* strMod(BoxedString* lhs, Box* rhs) {
     return rtn;
 }
 
+Box* strRMod(BoxedString* lhs, Box* rhs) {
+    Box* rtn = PyString_Format(rhs, lhs);
+    if (!rtn)
+        throwCAPIException();
+    return rtn;
+}
+
 extern "C" Box* strMul(BoxedString* lhs, Box* rhs) {
     assert(PyString_Check(lhs));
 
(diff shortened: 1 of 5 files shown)


undingen:
what do you think about waiting with merging this PR until my https://github.com/dropbox/pyston/pull/1157 is finished and merged and than changing this PR to not only update numpy but to also change the ```numpy_test.py``` run the numpy test suite?

Daetalus:
Np!

src/runtime/str.cpp
```diff
@@ -1164,6 +1164,13 @@ extern "C" Box* strMod(BoxedString* lhs, Box* rhs) {
     return rtn;
 }
 
+Box* strRMod(BoxedString* lhs, Box* rhs) {
```
undingen: please add a test for this

src/runtime/types.cpp
```diff
@@ -1616,6 +1616,12 @@ static Box* builtinFunctionOrMethodName(Box* b, void*) {
     return incref(func->name);
 }
 
+static Box* wrapperobjectName(Box* w, void*) {
+    assert(w->cls == wrapperobject_cls);
+    BoxedWrapperObject* wrapper_obj = static_cast<BoxedWrapperObject*>(w);
+    return incref(boxString(wrapper_obj->descr->wrapper->name));
```
undingen: this incref does not look correct. and please also add a test...

test/lib/numpy
```diff
@@ -1 +1 @@
-Subproject commit e4d4b45d26813899c66e112365896aabbb7b19fa
+Subproject commit f92576298546ffe984a2df466a80bb6a1d2e60c9
```
undingen: I think it's nicer to update to the last released version and not some unreleased random revision.. maybe we should take 4092a9e160cc247a4a45724579a0c829733688ca which is https://github.com/numpy/numpy/commits/v1.11.0

undingen:
And as you have seen there is a problem with building numpy inside the source directory. (some files are left around and we can end-up with some hybrid version...)
Could you therefore please also change the way the test builds numpy, either by copying the ```test/lib/numpy``` directory into the temp virtualenv directory before compiling it. Or remove the git submodule and change the test to directly download the source tarball of the last released numpy release from https://pypi.python.org/pypi/numpy into the virtualenv dir...
For an example you can look at some of the newer integration tests e.g. ```test/extra/cffi_test.py```

And can we start to run ```import numpy as np; np.test()``` inside the test script or is it too early for this? If the test takes too long we may want to remove it into the ```test/extra/``` directory because there it will only get run in release mode... 

Thanks for working on this :-)

undingen:
thanks for the fix!

PR #1173, opened 2016-05-16
interpreter: mark created_closure as live variable inside yields

undingen:
without this we would never traverse the variable and the GC would therefore be unable to destroy the generator


src/runtime/types.cpp
```diff
@@ -4018,7 +4018,7 @@ static int type_clear(PyTypeObject* type) {
     */
 
     PyType_Modified(type);
-    if (type->tp_dict)
+    if (type->tp_dict && type->tp_dict->cls != attrwrapper_cls)
```
kmod: maybe remove this from this pr?

undingen:
merged in with your fix.

PR #1176, opened 2016-05-17
Make rearrangeArgs take a callback instead passing back values

kmod:
This lets it more-efficiently keep track of the different situations in which it might return, because it doesn't have to pass back data signalling which path it took and what cleanups need to be done afterwards.

This is a perf win and also lets us not incref the args in as many cases -- before it was kind of expensive to pass back an extra flag saying whether or not the args should be decreffed.

```
                                origin/refcounting:  personal/rearrange:
       django_template3_10x.py            13.8s (2)            13.6s (2)  -1.4%
 sqlalchemy_imperative2_10x.py            19.2s (2)            18.9s (2)  -1.5%
            pyxl_bench2_10x.py            11.3s (2)            10.4s (2)  -8.6%
                       geomean                14.4s                13.9s  -3.9%
```

### DIFF
--- a/src/runtime/rewrite_args.h
+++ b/src/runtime/rewrite_args.h
@@ -292,37 +292,8 @@ struct CompareRewriteArgs {
         : rewriter(rewriter), lhs(lhs), rhs(rhs), destination(destination), out_success(false), out_rtn(NULL) {}
 };
 
-typedef Box* (*rearrange_target_t)(void*, CallRewriteArgs*, Box*, Box*, Box*, Box**);
-
-// This is a magical helper class that converts a functor (in particular, lambdas) into a [function pointer, void* data]
-// pair.  This lets us pass the functor efficiently: we can avoid doing an allocation (by storing the functor on the
-// stack),
-// and also avoid making the receiving function have to be a template.
-//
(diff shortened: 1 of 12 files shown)


src/runtime/rewrite_args.h
```diff
@@ -292,26 +292,58 @@ struct CompareRewriteArgs {
...
+typedef Box* (*rearrange_target_t)(void*, CallRewriteArgs*, Box*, Box*, Box*, Box**);
+
+// This is a magical helper class that converts a functor (in particular, lambdas) into a [function pointer, void* data]
+// pair.  This lets us pass the functor efficiently: we can avoid doing an allocation (by storing the functor on the
+// stack),
+// and also avoid making the receiving function have to be a template.
+//
+// In effect, it does what you would have to do to create a function pointer + void* pair, like one would pass to
+// C-style functions.
+class FunctorPointer {
```
undingen: I think you can replace the whole ```FunctorPointer``` implementation with:
```using FunctorPointer = llvm::function_ref<Box*(CallRewriteArgs* rewrite_args, Box* arg1, Box* arg2, Box* arg3, Box** args)>;```

undingen:
This is a really nice change and performance improvement! :-)
LGTM

PR #1179, opened 2016-05-18
Remove numpy submodule

Daetalus:
Remove numpy submodule, clone numpy from github with v1.11.0 tag when testing.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,9 +31,6 @@
 [submodule "build_deps/jemalloc"]
 	path = build_deps/jemalloc
 	url = git://github.com/jemalloc/jemalloc.git
-[submodule "test/lib/numpy"]
-	path = test/lib/numpy
-	url = https://github.com/numpy/numpy
 [submodule "test/lib/decorator"]
 	path = test/lib/decorator
 	url = https://github.com/micheles/decorator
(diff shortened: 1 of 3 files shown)


undingen:
looking good but we run into the 
```
======================================================================
ERROR: Failure: OSError (No such file /home/marius/pyston/numpy_test_env_pyston_dbg/site-packages/numpy-1.11.0-py2.7-linux-x86_64.egg/numpy)
----------------------------------------------------------------------
  File "/home/marius/pyston/numpy_test_env_pyston_dbg/site-packages/nose/failure.py", line 42, in runTest
    raise self.exc_class(self.exc_val)
OSError: No such file /home/marius/pyston/numpy_test_env_pyston_dbg/site-packages/numpy-1.11.0-py2.7-linux-x86_64.egg/numpy
```
error with the upgraded numpy when running
```import numpy as np; np.test(verbose=2)```

I investigated a little bit and it looks like the .egg file is unziped for cpython but for us its ziped...
I think setuptools should automatically detected that it's not allowed to zip it because it contains shared libraries etc... I guess somehow there is a check which expects that our files are named differently but I did not investigate further.
As a workaround please specify in the ```setup.py```: ```zip_safe=False``` in the  ```metadata``` dict.

**Comment by undingen** [comment]
thanks for updating numpy :-) 

I checked what the current testsuite results are for the refcounting branch:
```
Ran 5781 tests in 132.691s
FAILED (KNOWNFAIL=6, SKIP=15, errors=34, failures=3)
```
:+1:

Daetalus:
Seems with #1170 , the result will become:

> ----------------------------------------------------------------------
Ran 5781 tests in 85.915s
FAILED (KNOWNFAIL=6, SKIP=15, errors=9, failures=8)

PR #1189, opened 2016-05-20
Reenable the 'extra' tests

kmod:


undingen:
oh wow did I really forget to enable the extra tests...:blush: 
thanks for enabling it.

PR #1194, opened 2016-05-21
[WIP]Some miscs numpy fixing

Daetalus:
NumPy array defined `__index__` attribute, and it is a `PySequence`. I think in Python, we should avoid to let string object to handle `str` multiply with an object who defined them both.


kmod:
Sorry this got closed -- I deleted the refcounting branch which closed any PRs against it.

PR #1201, opened 2016-05-23
[WIP] Some minor numpy fixing.

Daetalus:
- There's no rewriter support in hex and oct in Pyston for now, so just change it to calling nb_hex.

### DIFF
--- a/src/runtime/builtin_modules/builtins.cpp
+++ b/src/runtime/builtin_modules/builtins.cpp
@@ -99,43 +99,42 @@ extern "C" Box* abs_(Box* x) {
     return rtn;
 }
 
-extern "C" Box* binFunc(Box* x) {
-    static BoxedString* bin_str = getStaticString("__bin__");
-    CallattrFlags callattr_flags{.cls_only = true, .null_on_nonexistent = true, .argspec = ArgPassSpec(0) };
-    Box* r = callattr(x, bin_str, callattr_flags, NULL, NULL, NULL, NULL, NULL);
-    if (!r)
-        raiseExcHelper(TypeError, "bin() argument can't be converted to bin");
-
-    if (!PyString_Check(r))
-        raiseExcHelper(TypeError, "__bin__() returned non-string (type %.200s)", r->cls->tp_name);
-
-    return r;
+static PyObject* builtin_bin(PyObject* self, PyObject* v) noexcept {
+    return PyNumber_ToBase(v, 2);
 }
 
-extern "C" Box* hexFunc(Box* x) {
-    static BoxedString* hex_str = getStaticString("__hex__");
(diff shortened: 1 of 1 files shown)


src/runtime/builtin_modules/builtins.cpp
```diff
@@ -112,30 +112,38 @@ extern "C" Box* binFunc(Box* x) {
...
-extern "C" Box* hexFunc(Box* x) {
-    static BoxedString* hex_str = getStaticString("__hex__");
-    CallattrFlags callattr_flags{.cls_only = true, .null_on_nonexistent = true, .argspec = ArgPassSpec(0) };
-    Box* r = callattr(x, hex_str, callattr_flags, NULL, NULL, NULL, NULL, NULL);
-    if (!r)
-        raiseExcHelper(TypeError, "hex() argument can't be converted to hex");
-
-    if (!PyString_Check(r))
-        raiseExcHelper(TypeError, "__hex__() returned non-string (type %.200s)", r->cls->tp_name);
+extern "C" Box* hexFunc(Box* self) {
```
undingen: this functions use the CAPI way to raise exceptions but this functions are registered as throwing c++ exceptions...
So you can either register them as capi functions (take a look at "next" in the same file, you have to pass CAPI to ```FunctionMetadata::create``` and please declare the function as ```noexcept``` too to make sure that no exception accidentally slips through)
or you have to convert them to use ``` raiseExcHelper```

undingen:
lookg good but I added one comment about the capi/c++ exception mismatch and could you please also change ```bin()``` to call into ```PyNumber_ToBase``` like cpython does.
Thank you very much for working on this!

src/runtime/builtin_modules/builtins.cpp
undingen: oh there is a 3. option which is probably the nicest: copy the functions from cpython and register them with by adding them to the ```static PyMethodDef builtin_methods``` array

undingen:
thanks for addressing my comments in the first commit :-)
But I'm not sure about the second commit you added. It looks like a very specific fix/workaround at least I don't really understand why it's necessary here. I will wait with merging it.
@kmod could you take a look at the second https://github.com/dropbox/pyston/pull/1201/commits/46e77e8edc0f504dcb178ecbd3c76b6abc85851d commit please.

kmod:
It looks like the second commit is for the "sq_repeat should have lower priority than `__rmul__`" issue.  I think the fix here is probably too specific to the case that's failing in the numpy testsuite, since I think the solution has more to do with whether the other operand has a `__rmul__` than whether it is a sequence.  Maybe let's separate that part from this PR so we can get the first commit in?

undingen:
thanks for the fix!

PR #1210, opened 2016-05-24
str: use tp_as_sequence instead of tp_as_number

undingen:
string is special in that it is a c++ type which has tp_as_number and tp_as_sequence.
This causes problems because when we fixup the slot dispatcher we will set the tp_as_number fields but not the
tp_as_sequence because setting both can cause problems.
Some extensions (e.g. numpy) require that we use the sq_* functions instead of nb_*.
Therefore clear the tp_as_number fields (except nb_remainder which cpython has set too because it is not part of
tp_as_sequence).

### DIFF
--- a/from_cpython/Objects/stringobject.c
+++ b/from_cpython/Objects/stringobject.c
@@ -2,6 +2,8 @@
 
 #include "Python.h"
 
+#include <stddef.h>
+
 #include "stringlib/stringdefs.h"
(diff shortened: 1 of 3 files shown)


kmod:
This isn't the worst hack we have around setting up the class slots, so lgtm :)  I think we're getting closer to understanding the best way we should solve these issues in general, but until then I feel like it's fine to leave things a bit messy.

PR #1212, opened 2016-05-25
Some packaging / distributing updates

kmod:


README.md
```diff
@@ -18,14 +18,21 @@ We have some experimental docker images available; try running
 docker run --it pyston/pyston
```
undingen: this should be `-it` not `--it`

kmod:
Merging without CI results since this is a non-code change, and I'd like it to be in by the time the blog post goes up.

PR #1229, opened 2016-05-31
Some fixings that let scipy tests pass.

### LINKED ISSUE #1232: const qualifier in format member.
This is a reminder for work around which let SciPy work in Pyston. And also to give a place to discuss.

```
/* Py3k buffer interface */
typedef struct bufferinfo {
    void *buf;
    PyObject *obj;        /* owned reference */
    Py_ssize_t len;
    Py_ssize_t itemsize;  /* This is Py_ssize_t so it can be
                             pointed to by strides in simple case.*/
    int readonly;
    int ndim;
    // Pyston change: changed this from char* to const char*
    const char *format;
    Py_ssize_t *shape;
    Py_ssize_t *strides;
    Py_ssize_t *suboffsets;
    Py_ssize_t smalltable[2];  /* static store for shape and strides of
                                  mono-dimensional buffers. */
    void *internal;
} Py_buffer;
```
https://github.com/dropbox/pyston/blob/master/from_cpython/Include/object.h#L194

The `const` qualifier will cause this error:

```
scipy/cluster/_vq.c: In function ‘__pyx_pf_5numpy_7ndarray___getbuffer__’:
scipy/cluster/_vq.c:5934:5: error: assignment of read-only location ‘*__pyx_v_info->format’
     (__pyx_v_info->format[0]) = '^';
```

@kmod gave an idea:

> Maybe we could add something like

```
#ifdef _PYSTON_API
#define PYSTON_CONST const
#else
#define PYSTON_CONST
#endif
```

> and then switch all of our extra const additions to PYSTON_CONST?

The `format` should be `const`, but for real-world compatibility cases. Maybe we need to drop the `const` at last.

Daetalus:
closes #1232 by second commit.

### DIFF
--- a/from_cpython/Include/object.h
+++ b/from_cpython/Include/object.h
@@ -190,8 +190,7 @@ typedef struct bufferinfo {
                              pointed to by strides in simple case.*/
     int readonly;
     int ndim;
-    // Pyston change: changed this from char* to const char*
-    const char *format;
+    PYSTON_CONST char *format;
     Py_ssize_t *shape;
     Py_ssize_t *strides;
     Py_ssize_t *suboffsets;
--- a/from_cpython/Include/pyport.h
+++ b/from_cpython/Include/pyport.h
@@ -16,6 +16,12 @@
 #define PYSTON_NOEXCEPT
 #endif
 
+#ifdef _PYSTON_API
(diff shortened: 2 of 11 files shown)


undingen:
:+1: thanks for your patch :-)

PR #1231, opened 2016-06-01
add prefix to Pyston defined True and False

### LINKED ISSUE #1242: Pyston didn't defined PY_LLONG_MAX yet.
CPython 2.7.7 defined `PY_LLONG_MAX` and other macro in `Include/pyport.h`. But Pyston missed it. Cython rely on it:

https://github.com/Daetalus/cython/commit/9d0fff29d26f6d9aa2f251a3402db4e0b9c6a0b3#diff-b68ea089583bb1818ad9a4fb39c5d534R232

I see no harm to introduce it. Maybe it just Pyston based on an old version of CPython?

Daetalus:
Some extensions will use `True` and `False` from other package. For example, X11 defines `True` and `False` in `/usr/include/X11/Xlib.h`. Which will cause name conflict when try to build extensions which use x11.

I think this is not good solution. V8 encountered it too:
https://bugs.chromium.org/p/v8/issues/detail?id=448
But we can't control how extension handle it. So just add a prefix to it.

closes #1242

### DIFF
--- a/from_cpython/Include/boolobject.h
+++ b/from_cpython/Include/boolobject.h
@@ -9,10 +9,7 @@ extern "C" {
 #endif
 
 
-// Pyston change: Try to avoid having to support mixing ints and bools
-// typedef PyIntObject PyBoolObject;
-struct _PyBoolObject;
-typedef struct _PyBoolObject PyBoolObject;
+typedef PyIntObject PyBoolObject;
 
 // Pyston change: this is no longer a static object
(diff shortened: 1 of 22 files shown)


kmod:
looks good to me, but maybe let's get rid of `pyston_True` and just use `Py_True` everywhere?  Either that or we should figure out some way of naming it `pyston::True` instead -- but I am not sure how to make Py_True map to `pyston::True` for C extensions.

undingen:
LGTM, thanks for the change!

PR #1236, opened 2016-06-07
float.cpp: now that we have float.c use some of the functions directly

undingen:
this also workarounds the problem we were having on ubuntu 16.04 where we have to use `std::isinf` instead of `isinf()`


kmod:
lgtm, I'll merge once the CI issues are resolved

kmod:
Hmm I thought closing+reopening the PR would trigger a re-test, but maybe it doesn't anymore.  Could you rebase this to the new master?

PR #1237, opened 2016-06-07
Fix std::move error on temporary object

c-rhodes:


undingen:
thanks for the patch! :+1: 

I'm merging this even though CI says it fails because it fails because the llvm apt repo is still down and I manually verified that it works.

PR #1244, opened 2016-06-10
perf: add a script which allows 'perf report' to disassemble JITed functions

undingen:
a hack which should only get enabled when specifically looking into the bjit perf...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -819,7 +819,7 @@ endef
 
 .PHONY: perf_report
 perf_report:
-	perf report -n
+	perf report -n --objdump=tools/perf_jit.py
 
 .PHONY: run run_% dbg_% debug_% perf_%
 run: run_dbg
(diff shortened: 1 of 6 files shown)


kmod:
This looks awesome :)  Just some minor comments
- Do we need the extra #ifdef, or can we just gate it behind `-p`?
- Could we integrate this with the same functionality that we have for the llvm jit?  I think if we just put it into the same folder ("perf_map/") and merge your perf_bjit_helper.py into tools/annotate.py, we could have it work for either (annotate.py has some nice pyston-specific features as well).

Some day it would be nice to merge in the assembly comments as well :)

undingen:
oh I did not know about `tools/annotate.py` / forgot about it... I will come up with a patch which integrates both tools and integrates them into `perf report` :-)
guess we could and up having a really nice tool :)

kmod:
ok cool :)  I don't want to block merging this just because there are more potential features, but I think/hope that doing a simple merge with annotate.py wouldn't take that long.

undingen:
so I merged it together and removed all this code duplication with the annotate tool. (did not know that `func_addr_registry.registerFunction` already did all this work :unamused: )
I still left a small wrapper script `tools/perf_jit.py` around because I think it's nicer than having to support all objdump arguments directly inside `annotate.py`.
I wish I wrote this small script a long time ago it is quite handy :-D

kmod:
lgtm!

Issue #1097, opened 2016-02-11
Improve REPL support for loops


undingen:
In interactive mode we don't currently print the value of every loop iteration etc:
pyston:
```
>> for i in range(3):
...  i
... 
>> 
```
python:
```
>>> for i in range(3):
...  i
... 
0
1
2
>>> 
```

PR #1245, opened 2016-06-11
Improve REPL support

c-rhodes:
This fixes a bug with expressions inside loops in interactive mode
that resulted in expressions not being printed:

    >> for i in range(3):
    ...    i
    ...
    >>

Python prints the value every iteration:

    >> for i in range(3):
    ...    i
    ...
    0
    1
    2
    >>>

The same also applies to while loops.


c-rhodes:
This fixes https://github.com/dropbox/pyston/issues/1097

undingen:
cool! :+1: , thanks for your patch :-)

PR #1249, opened 2016-06-13
use cpython's abstract.c instead of src/capi/abstract.cpp

aisk:


undingen:
thanks for working on this!
Could you please add a commit of the original unmodified cpythons files as a separate commit because otherwise it's hard to see what changed or revert unneeded changes in the future. (Sorry for not mentioning this :disappointed: )
Can we also use cpythons `PySlice_GetIndices` and `PySlice_GetIndicesEx`?
We don't use this always :disappointed: but could you please mark your `if 0` with 
```
// pyston change: comment this out
#if 0
```
I will review it again in more detail after you have switched the modifications to the cpython files into a separate commit.
But Looks good :) :+1:

aisk:
@undingen done, and another question, should I move the `PyNumber_Coerce` functions (the only two function leaves in `abstract.cpp`) to another place? I think it shouln't be here.

undingen:
sorry for the delay in reviewing your updated patch :-(.
thanks for addressing my comments :+1: 

I have a few more comments:
There a few functions where we want to keep our pyston modified ones because the contain a faster implementation. I think this are (We did a pretty bad job in marking the changes in this file :unamused: ):
- `PyObject_Size`
- `PyObject_GetIter`
- `PyObject_IsInstance`
- `PyObject_IsSubclass`

it may be later possible to port some of this modifications to the c functions but I think this should be done in a separate PR.

We can use the implementation for `PyObject_GetBuffer`, `PyObject_Format` from `abstract.c`

You can move the `PyNumber_Coerce` to `capi/object.cpp` 

If you address this comments I will check the perf to see if we caught all optimizations (currently this patch is quite a slowdown :-()
Thanks for working on this!

aisk:
@undingen Hi sorry I forget this PR , but now it got passed the unit test, will you check it again? thanks!

undingen:
Cool :-), I will take a look at it on the weekend

undingen:
Sorry that it tooks me so long to review this :-(.
Looks good but I found some smaller things so one more round..:
Some functions are currently defined twice: once in `abstract.cpp` and `abstract.c` we should only keep the version in `abstract.c`: 
- `recursive_issubclass`
- `check_class`
- `abstract_get_bases`
but we will need to remove the static keyword and declare them in the `abstract.cpp` because they are used by `recursive_isinstance` which we can't move please also `#if 0` the version of it in the `*.c` file. 

We should also keep our implementation of the following functions because they are a little bit faster than cpythons:
- `PyObject_CallMethod`
- `PyObject_CallMethodObjArgs`
- `_PyObject_CallMethod_SizeT`

Thanks

aisk:
@undingen done, can you review it again 😀

undingen:
:+1: 
LGTM, thanks for the fix! :smile:

undingen:
I think another function we could remove from `capi.cpp` is `PyObject_Format` if you are interested.

aisk:
@undingen OK I'll have a try 😄

aisk:
Hi @undingen can you please have a look at this: https://github.com/dropbox/pyston/pull/1311

PR #1251, opened 2016-06-14
Add some helper API to Pyston.

Daetalus:
First, I try to switch from PyPy based Cython(which Pyston currently used) to CPython based Cython. Then Pyston need add some new API to process some operation in CPython based Cython.

Includes some other scipy related changes.

### DIFF
--- a/src/runtime/code.cpp
+++ b/src/runtime/code.cpp
@@ -200,9 +200,15 @@ extern "C" BORROWED(PyObject*) PyCode_GetFilename(PyCodeObject* op) noexcept {
     RELEASE_ASSERT(PyCode_Check((Box*)op), "");
     return BoxedCode::filename((Box*)op, NULL);
 }
+
 extern "C" BORROWED(PyObject*) PyCode_GetName(PyCodeObject* op) noexcept {
     RELEASE_ASSERT(PyCode_Check((Box*)op), "");
     return BoxedCode::name((Box*)op, NULL);
+}
+
+extern "C" int PyCode_HasFreeVars(PyCodeObject* _code) noexcept {
+    BoxedCode* code = (BoxedCode*)_code;
+    return code->f->source->getScopeInfo()->takesClosure() ? 1 : 0;
 }
 
 void setupCode() {
--- a/src/runtime/frame.cpp
+++ b/src/runtime/frame.cpp
@@ -347,6 +347,7 @@ extern "C" PyFrameObject* PyFrame_New(PyThreadState* tstate, PyCodeObject* code,
 extern "C" BORROWED(PyObject*) PyFrame_GetGlobals(PyFrameObject* f) noexcept {
     return BoxedFrame::globals((Box*)f, NULL);
 }
+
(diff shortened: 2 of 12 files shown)


src/runtime/code.cpp
```diff
@@ -200,11 +200,17 @@ extern "C" BORROWED(PyObject*) PyCode_GetFilename(PyCodeObject* op) noexcept {
...
 }
+
 extern "C" BORROWED(PyObject*) PyCode_GetName(PyCodeObject* op) noexcept {
     RELEASE_ASSERT(PyCode_Check((Box*)op), "");
     return BoxedCode::name((Box*)op, NULL);
 }
 
+extern "C" int PyCode_GetNumFree(PyCodeObject* _code) noexcept {
+    BoxedCode* code = (BoxedCode*)_code;
+    return code->f->source->getScopeInfo()->takesClosure();
```
kmod: It looks like this doesn't actually return the number of free variables, but rather whether there are any.  Can we change the function to `bool PyCode_HasFreeVars`?

src/runtime/frame.cpp
```diff
@@ -347,10 +347,19 @@ extern "C" PyFrameObject* PyFrame_New(PyThreadState* tstate, PyCodeObject* code,
...
 extern "C" BORROWED(PyObject*) PyFrame_GetCode(PyFrameObject* f) noexcept {
     return BoxedFrame::code((Box*)f, NULL);
 }
 
+extern "C" BORROWED(PyObject*) PyFrame_GetBack(PyFrameObject* f) noexcept {
+    return BoxedFrame::back((Box*)f, NULL);
+}
+
+extern "C" void PyFrame_SetBack(PyFrameObject* f, PyFrameObject* back) noexcept {
+    BoxedFrame::back((Box*)f, back);
```
kmod: I don't think this does what you want it to do -- it just fetches the current f_back.  We don't currently have a way to set f_back; we could add one, but it would be a bit complicated (I think we would need to update `frame_info->back` as well?) and we would need to test it really well.

I spent some time looking at the Cython code that is calling this though, and as far as I can tell, it is a hacky and buggy way of implementing a feature that shouldn't exist :P .  I am ok with us just using the if-PyPy side which does nothing, which I think would be saner and also not require us to add PyFrame_SetBack.  If we really want to support that particular Cython code, it might be better to add a higher-level API since I don't know if we can always support PyFrame_SetBack properly.

Daetalus:
@kmod This PR was updated, would you mind to take a look if you have time, please?

kmod:
Looks good to me!  What's the plan for the scipy test (should we merge it like this)?

Daetalus:
I think merge it just fine. When SciPy test could pass, just enable it.

Daetalus:
@kmod Ready to review.

kmod:
Awesome!  If you get it to the point that the tests run (even if a number fail), it'd be great to enable the test so that we can start having it run as part of our CI.

Daetalus:
Will enable the scipy test after fix all the segfaults(currently there have two segfaults) in SciPy test suite.

PR #1255, opened 2016-06-16
rewriter+bjit: remove duplicate getattrs, misc codegen improvements

undingen:
Misc improvements, please take a look at the individual commits to find out what changed
```
                                   upstream/master:    origin/bjit_opt7:
           django_template3.py             2.6s (4)             2.4s (4)  -6.7%
                 pyxl_bench.py             2.3s (4)             2.2s (4)  -1.9%
     sqlalchemy_imperative2.py             2.8s (4)             2.6s (4)  -7.9%
                pyxl_bench2.py             1.2s (4)             1.2s (4)  -2.0%
       django_template3_10x.py            13.9s (4)            13.1s (4)  -5.7%
             pyxl_bench_10x.py            17.6s (4)            17.5s (4)  -0.9%
 sqlalchemy_imperative2_10x.py            20.4s (4)            18.9s (4)  -7.5%
            pyxl_bench2_10x.py            10.2s (4)            10.0s (4)  -1.9%
                       geomean                 5.6s                 5.4s  -4.4%

```

### DIFF
--- a/src/asm_writing/rewriter.cpp
+++ b/src/asm_writing/rewriter.cpp
@@ -416,6 +416,20 @@ void Rewriter::_addAttrGuard(RewriterVar* var, int offset, RewriterVar* val_cons
 RewriterVar* RewriterVar::getAttr(int offset, Location dest, assembler::MovType type) {
     STAT_TIMER(t0, "us_timer_rewriter", 10);
 
+    // if no changing action happened we can reuse get attributes
+    if (!rewriter->added_changing_action) {
+        RewriterVar*& result = getattrs[std::make_pair(offset, (int)type)];
+        if (result) {
+            if (dest != Location::any())
+                result->getInReg(dest, true /* allow_constant_in_reg */);
+        } else {
+            result = rewriter->createNewVar();
+            rewriter->addAction([=]() { rewriter->_getAttr(result, this, offset, dest, type); }, { this },
+                                ActionType::NORMAL);
+        }
+        return result;
+    }
+
     RewriterVar* result = rewriter->createNewVar();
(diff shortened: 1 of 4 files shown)


undingen:
ready for review :-)
(had to remove some of the "do guards smarter" patch which I thought would be a improvement but turned out to sometimes causes aborts because of additional spilling)

src/asm_writing/rewriter.cpp
```diff
@@ -1148,6 +1146,20 @@ void Rewriter::_setupCall(bool has_side_effects, llvm::ArrayRef<RewriterVar*> ar
 #endif
 }
 
+void Rewriter::_callOptimalEncoding(assembler::Register tmp_reg, void* func_addr) {
```
kmod: I think this might have an issue for non-bjit rewrites -- for those we assemble into a temporary buffer, and then memcpy the data to the slot when it's done.  So in that case I think curInstPointer() will point into the temporary buffer and not the final location.  Maybe we could have a `will_be_relocated` flag or something similar?

As a random other thought, is our bjit reenttrant?  ie is it possible that while the bjit is jitting, it calls back into itself and tries to bjit that same section.

kmod: Oh sorry you handled that :) I keep forgetting that we pick the destination before we start writing everything out.

undingen: should be fine because like you wrote we are selecting the ICSlot at the start of the emitting phase.
And we were already doing this relative call optimization inside `_call` this only extended it to other places too.

As for the bjit being reentrant: the `JitCodeBlock` has a `is_currently_writing` which makes sure that we don't start another bjit emitter for the same function. So this should be fine too :+1:

Issue #1135, opened 2016-04-12
Dictionary comparisons


kmod:
We currently compare dictionaries by pointer, but CPython has a consistent ordering to them.  We should implement the same process that they have.

ie get test/tests/dict_comparisons.py working

kmod:
Closed by #1272

PR #1256, opened 2016-06-16
Dictionary comparisons #1135

sizeoftank:
Get test/tests/dict_comparisons.py working.


kmod:
This looks great :) any reason to close it?

The test failures are a bit annoying
- You'll have to run `make format` to get it to match our formatting rules
- It looks like we might have to bump our test timeouts for when Travis-CI is busy.  I'll get around to it or if you don't mind you can just change the `-t75` to `-t90` in CMakeLists.txt
- We have to remove the `expected: fail` from dict_comparisons.py (best kind of test failure!)
- The sqlalchemy_smalltest issue is weird.  I can help take a look if you want

Also, you'll have to sign the CLA before we can merge it.

src/runtime/dict.cpp
```diff
@@ -900,6 +900,136 @@ static int dict_print(PyObject* mp, FILE* fp, int flags) noexcept {
...
+        thisaval = p.second;
+        assert(thisaval);
+        Py_INCREF(thisaval);   /* keep alive */
+
+        BoxedDict::DictMap::iterator it;
+        try {
+            it = b->d.find(thiskey);
+            thisbval = it->second;
+        } catch (ExcInfo e) {
+            setCAPIException(e);
```
kmod: Should this have a `goto Fail` here?

sizeoftank: Yes, It seems a problem because of thisbval uninitialized.
The compiler no error when I run make python_release, but can not build successed in travis_ci
failed log can't open in travis ci yesterday, so I  did not realize that.

And I will use clang-format to match yours rule, to fix that, add a goto Fail is a good advice, put a new PR after this, thanks.

kmod: I'm guessing that the difference is due to pyston_release building with clang, but travis-ci building with gcc.  (This was a recent change due to clang becoming uninstallable on travis-ci.)  You can build with gcc using `make pyston_gcc`

PR #1258, opened 2016-06-17
Dictionary comparisons #1135

sizeoftank:
Get test/tests/dict_comparisons.py working.

### DIFF
--- a/src/runtime/dict.cpp
+++ b/src/runtime/dict.cpp
@@ -664,10 +664,6 @@ Box* dictEq(BoxedDict* self, Box* _rhs) {
         raiseExcHelper(TypeError, "descriptor '__eq__' requires a 'dict' object but received a '%s'",
                        getTypeName(self));
 
-    if (_rhs == Py_None) {
-        Py_RETURN_FALSE;
-    }
-
     if (_rhs->cls == attrwrapper_cls) {
         _rhs = attrwrapperToDict(_rhs);
     } else {
@@ -1031,49 +1027,26 @@ Finished:
     return res;
 }
 
-
 static PyObject* dict_richcompare(PyObject* v, PyObject* w, int op) noexcept {
     Box* res;
 
-    if (w == Py_None) {
-        if (op == Py_NE || op == Py_GT) {
-            Py_RETURN_TRUE;
-        } else {
-            Py_RETURN_FALSE;
-        }
-    }
-
     if (!PyDict_Check(v) || !PyDict_Check(w)) {
-        res = Py_NotImplemented;
-        return incref(res);
-    }
-
-    if (op == Py_EQ) {
-        return dictEq((BoxedDict*)v, (BoxedDict*)w);
+        res = incref(Py_NotImplemented);
+    } else if (op == Py_EQ) {
+        res = dictEq((BoxedDict*)v, (BoxedDict*)w);
     } else if (op == Py_NE) {
-        return dictNe((BoxedDict*)v, (BoxedDict*)w);
+        res = dictNe((BoxedDict*)v, (BoxedDict*)w);
     } else {
-        int cmp = dict_compare((BoxedDict*)v, (BoxedDict*)w);
-        switch (op) {
-            case Py_LT:
-                res = (cmp < 0) ? Py_True : Py_False;
-                break;
-            case Py_LE:
-                res = (cmp <= 0) ? Py_True : Py_False;
-                break;
-            case Py_GT:
-                res = (cmp > 0) ? Py_True : Py_False;
-                break;
-            case Py_GE:
-                res = (cmp >= 0) ? Py_True : Py_False;
-                break;
-            default:
-                RELEASE_ASSERT(0, "%d", op);
+        /* Py3K warning if comparison isn't == or !=  */
+        if (PyErr_WarnPy3k("dict inequality comparisons not supported "
+                           "in 3.x",
+                           1) < 0) {
+            return NULL;
         }
-        return incref(res);
+        res = incref(Py_NotImplemented);
     }
+    return res;
 }
-
 
 void BoxedDict::dealloc(Box* b) noexcept {
     if (_PyObject_GC_IS_TRACKED(b))
@@ -1165,6 +1138,7 @@ void setupDict() {
         = dictiteritem_cls->instances_are_nonzero = true;
 
     dict_cls->tp_hash = PyObject_HashNotImplemented;
+    dict_cls->tp_compare = (cmpfunc)dict_compare;
     dict_cls->tp_richcompare = dict_richcompare;
 
     dict_cls->giveAttr("__len__", new BoxedFunction(FunctionMetadata::create((void*)dictLen, BOXED_INT, 1)));
--- a/src/runtime/dict.cpp
+++ b/src/runtime/dict.cpp
@@ -1027,6 +1027,28 @@ Finished:
     return res;
 }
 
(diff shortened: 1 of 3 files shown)


kmod:
If you push to your github branch (the one created the PR from, in this case your `master`), the PRs will automatically include your changes :)

sizeoftank:
Thanks.
It looks like caught a new issue in sqlalchemy_smalltest.py?

7/7 Test #7: pyston_defaults_integration ......***Failed  823.87 sec

  sqlalchemy_smalltest.py    FAILED (Exited with code 1 (expected code 0))

  File "/home/travis/build/dropbox/pyston/test/integration/sqlalchemy_smalltest.py", line 254, in <module>

    getattr(n, t)()

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/test/orm/test_collection.py", line 1680, in test_mixin

    self._test_scalar_mapped(collection_class)

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/test/orm/test_collection.py", line 1514, in _test_scalar_mapped

    p.children['foo'] = Child('foo', 'value')

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/lib/sqlalchemy/orm/collections.py", line 1225, in __setitem__

    value = __set(self, value, _sa_initiator)

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/lib/sqlalchemy/orm/collections.py", line 1044, in __set

    item = executor.fire_append_event(item, _sa_initiator)

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/lib/sqlalchemy/orm/collections.py", line 713, in fire_append_event

    return self.attr.fire_append_event(

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/lib/sqlalchemy/orm/attributes.py", line 938, in fire_append_event

    state._modified_event(dict_, self, NEVER_SET, True)

  File "/home/travis/build/dropbox/pyston/test/lib/sqlalchemy/lib/sqlalchemy/orm/state.py", line 534, in _modified_event

    if previous not in (None, NO_VALUE, NEVER_SET):

TypeError: Ordered.__cmp__(x,y) requires y to be a 'Ordered', not a 'NoneType'

And how to run this test in my machine?

sizeoftank:
thanks.
It looks like cause a new issue in `sqlalchemy_smalltest`, and I don't know how to run this test in my local machine, could you tell me?

Daetalus:
You can run it by `make check_sqlalchemy_smalltest`. Maybe the problem is because our dict didn't implement `__cmp__` attribute explicit?

sizeoftank:
Thanks.
the `  cmp  ` method execute failed when dict compare to  None.
TypeError: dict.__cmp__(x,y) requires y to be a 'dict', not a 'NoneType'

src/runtime/dict.cpp
```diff
@@ -1027,6 +1027,28 @@ static int dict_compare(BoxedDict* a, BoxedDict* b) noexcept {
...
+    } else {
+        /* Py3K warning if comparison isn't == or !=  */
+        if (PyErr_WarnPy3k("dict inequality comparisons not supported "
+                           "in 3.x",
+                           1) < 0) {
+            return NULL;
+        }
+        res = Py_NotImplemented;
+    }
+    Py_INCREF(res);
```
kmod: This isn't quite right -- dictEq and dictNe return an owned (already-incref'd) object, so this will end up double-increffing them.  I think we should just say `res = incref(Py_NotImplemented)` on the other paths.

src/runtime/dict.cpp
```diff
@@ -1119,12 +1141,14 @@ void setupDict() {
...
     dict_cls->tp_hash = PyObject_HashNotImplemented;
     dict_cls->tp_compare = (cmpfunc)dict_compare;
+    dict_cls->tp_richcompare = dict_richcompare;
 
     dict_cls->giveAttr("__len__", new BoxedFunction(FunctionMetadata::create((void*)dictLen, BOXED_INT, 1)));
     dict_cls->giveAttr("__new__", new BoxedFunction(FunctionMetadata::create((void*)dictNew, UNKNOWN, 1, true, true)));
     dict_cls->giveAttr("__init__", new BoxedFunction(FunctionMetadata::create((void*)dictInit, NONE, 1, true, true)));
     dict_cls->giveAttr("__repr__", new BoxedFunction(FunctionMetadata::create((void*)dictRepr, STR, 1)));
 
+    dict_cls->giveAttr("__cmp__", new BoxedFunction(FunctionMetadata::create((void*)dict_richcompare, UNKNOWN, 2)));
```
kmod: I think this line should be removed -- `__cmp__`  is supposed to mirror tp_compare not tp_richcompare.  Also we don't need to add these attributes manually; calling `dict_cls->freeze()` later will add them automatically for us.

kmod:
I looked a bit into that sqlalchemy failure and it's due to `tp_richcompare` not being implemented (in those commits).  The logic is that comparisons first try tp_richcompare, and if that fails/doesn't exist, then they try tp_compare.  And the code that would handle `{} == None` lives in dict's tp_richcompare.  So I think we can just remove the `__cmp__` addition and it should be good.  I'm not 100% sure why adding it confuses the type-initialization code but it looks like removing it helps :)

src/runtime/dict.cpp
sizeoftank: Can I add a ` AUTO_DECREF(eq) ` here like dictNe?

sizeoftank:
the test_collections in sqlalchemy failed, the class Ordered,  I guess it doesn't initialize Ordered with dict_richcompare, so call ` Ordered.__cmp__ ` failed.

kmod:
Couple things
- I don't think it really matters that the sqlalchemy test is `Ordered` (a subclass of dict) and not just `dict`
- `{}.__cmp__(None)` is supposed to throw a TypeError
- `{} == None` is supposed to call richcompare and never even hit `tp_compare`.
- Let's not start straying from how CPython handles comparisons.  While I can't think of anything specifically that will fail if we start calling dict_compare from dict_richcompare, I think it's a risk we don't need to take.  I think the only issue with your earlier commit was the lack of dict_richcompare / the manual addition of a `__cmp__` function, and if we fix both those things we don't need to change from CPython's behavior.

sizeoftank:
If I work it like my earlier commit, `{}.__cmp__(None)` throws a TypeError and ` {} == None ` take a correct result in my machine.
But I don't why ` previous not in (None, NO_VALUE, NEVER_SET) ` call the ` __cmp__() ` and throws a TypeError in sqlalchemy_smalltest

kmod:
Hey, so I found what the issue was -- when we tiered up, your changes ended up triggered an old code path that didn't handle `__cmp__` correctly.  I patched that up and sent it to you in a PR.

I think with that new commit this PR should be ready to go.  Would you mind cleaning this up a bit by squashing the commits?  (ie it would be nice to not have some code and then revert that code in a single PR.)

sizeoftank:
Aha, Nice ! 
I have merged your commit to my fork with squashing.

I lack of git skills, it is a small trouble that I have reseted my master by ` git push -f  ` a few days ago. 
in that casr, this PR can not reopen ?

kmod:
Oh interesting, I am surprised github won't allow more updates to this PR.  Here's what I would do to:
```
git rebase origin/master
git checkout -b dict_comparisons origin/master
git cherry-pick 9fbffec
git checkout master .
git commit -am "Fix dict comparisons"
```
and then `git push` and submit a new PR.

I'm not 100% sure what your local state looks like at this point, so let me know if this doesn't work.

PR #1259, opened 2016-06-20
ICs: variable size IC slots

undingen:
ICs: variable size IC slots
before this change we had a fixed number of equal size slots inside an IC.
Because of the large difference in rewrite sizes a fixed slot size is not ideal.
We often ended up with too large slots which prevented us from emitting more slots.
(or very large ICs in general which are bad for memory usage and the instruction cache)

With this commit we will start with a single slot with the size of the whole IC and then decreasing its size to the actual bytes emitted
and creating a new slot which starts directly at the end of the previous slot.
We will repeat to to this until the space left is smaller than the number of bytes the last slot required.
This makes it a little bit less likely that we will be successful in overwritting an existing slot
but our benchmarks show that it is still a large win on all our benchmarks.

Some notable changes are:
- we pick the slot to rewrite much earlier now and prevent it from getting rewritten while we rewrite using num_inside = 1
  the reason is that we would otherwise not know how big the slot is
- when we resize a slot we have to patch the failing guard jumps to the address of the next slot

```
           django_template3.py             2.4s (4)             2.3s (4)  -4.3%
                 pyxl_bench.py             2.3s (4)             2.2s (4)  -3.0%
     sqlalchemy_imperative2.py             2.6s (4)             2.5s (4)  -0.8%
                pyxl_bench2.py             1.1s (4)             1.0s (4)  -9.7%
       django_template3_10x.py            13.1s (4)            12.5s (4)  -4.2%
             pyxl_bench_10x.py            17.7s (4)            16.9s (4)  -4.7%
 sqlalchemy_imperative2_10x.py            19.0s (4)            18.9s (4)  -0.9%
            pyxl_bench2_10x.py             9.7s (4)             8.4s (4)  -13.4%
                       geomean                 5.3s                 5.1s  -5.2%
```

### DIFF
--- a/src/asm_writing/assembler.h
+++ b/src/asm_writing/assembler.h
@@ -206,6 +206,7 @@ public:
     uint8_t* startAddr() const { return start_addr; }
     int bytesLeft() const { return end_addr - addr; }
     int bytesWritten() const { return addr - start_addr; }
+    int size() const { return end_addr - start_addr; }
     uint8_t* curInstPointer() { return addr; }
     void setCurInstPointer(uint8_t* ptr) { addr = ptr; }
     bool isExactlyFull() const { return addr == end_addr; }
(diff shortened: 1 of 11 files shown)


kmod:
:+1: sounds awesome!

PR #1266, opened 2016-06-26
enable the clang travis CI build again

undingen:


undingen:
this disables the gcc debug build. I'm not sure if we want to have it back if so I will add it back later.

PR #1267, opened 2016-06-26
Some scipy related fixing.

Daetalus:


undingen:
looks good but can you please remove the implementations of the functions we used previously.
Does this resolve all the cython function type issues I encountered?

Daetalus:
Emm.... What do you mean about "remove the implementations of the functions we used previously"?

undingen:
oh yeah sorry :blush: you are right! we have to keep them

PR #1268, opened 2016-06-26
Remove the "CPython executable parser"

Daetalus:
Pyston call CPython executable to generate CPython AST. Remove this
"parser", use CPython Parser module by default. Pass "-x" to enable
libpypa parser.


undingen:
Can we please leave at least on `-x` test in because otherwise pypa is untested.
Please update `README.md` it mentions: `-x Disable the pypa parser.`
I think you can remove the `src/codegen/parse_ast.py` file.

thanks for working on this :+1:

Daetalus:
Emm... I don't understand, if I understand it correctly, the libpypa never tested... you can verify it by something like `./pyston_dbg -X test_cmath`? `pypa` has bug to process the `complex`.

kmod:
The flag-checking looks confusing since we start with all three parsers "enabled" and then turn them off with the `-x` and `-X` flags.  I think the result is:
- no flags or `-x`: cpython parser
- `-X`: pypa
- `-x` **and** `-X`: old executable parser

You're right that `-X` (pypa) has an issue running the cpython tests.  But I don't think we should remove it from the test configurations that currently use it, at least in this PR.  (Or rather, I think we should update them to take `-x` which is the new way to use the pypa parser after this PR.)

Also, could you delete the src/codegen/parse_ast.py file?  You will also need to update CMakeLists.txt to remove the section that references it.

src/codegen/parser.cpp
```diff
@@ -998,119 +998,53 @@ AST* readASTMisc(BufferedReader* reader) {
...
-        const char* fn = "<string>";
-        mod_ty mod = PyParser_ASTFromString(code, fn, Py_file_input, &cf, arena);
-        if (!mod)
-            throwCAPIException();
-        assert(mod->kind != Interactive_kind);
-        auto rtn = static_cast<AST_Module*>(cpythonToPystonAST(mod, fn));
-        return rtn;
-    }
-
     if (ENABLE_PYPA_PARSER || inherited_flags) {
```
kmod: We should take out the `|| inherited flags` check here.

src/codegen/parser.cpp
```diff
@@ -1150,50 +1084,25 @@ static std::vector<char> _reparse(const char* fn, const std::string& cache_fn, A
...
-            file_data.insert(file_data.end(), buf, buf + nread);
-
-            for (int i = 0; i < nread; i++) {
-                checksum ^= buf[i];
-            }
-        }
-        int code = pclose(parser);
-        assert(code == 0);
-    }
+    if (ENABLE_PYPA_PARSER || inherited_flags)
```
kmod: I think we need an `else` after this, and we should also remove the `inherited_flags` check.

Daetalus:
Hi, please take a look. I don't know why get this error:
```
5: ../../src/codegen/parser.cpp:1211: pyston::AST_Module *pyston::caching_parse_file(const char *, FutureFlags): Assertion `length > 0 && length < 10 * 1048576' failed: invalid file length: 0 (file size is 9)
5: Someone called abort!
```

kmod:
That's complaining about an invalid pyc file, one which somehow doesn't contain any data in it (even an empty py file will generate a little bit of metadata).  Since you've been changing the parsing stuff, there's always a risk of leaving a buggy pyc file around, so I would try doing `find -name '*.pyc' -delete` to clear out any pyc files, and try again.

undingen:
:+1: thanks for the cleanup!

PR #1276, opened 2016-07-01
Add type.__basicsize__, type.__itemsize__

sizeoftank:
give ` type_cls ` the attribute member ` __basicsize__  ` and  ` __itemsize__ `


undingen:
thanks for the patch! :+1:

PR #1277, opened 2016-07-01
Switch almost everything to use vregs when possible

kmod:
The motivation was to reduce the time spent in our analysis passes and irgen.  Unfortunately once I started converting most things, it was hard to leave a few things unconverted, so I converted some things that didn't provide too much benefit (OSR passing, deopt passing).  The main thing that isn't converted is scoping-analysis, since that compares names across different scopes which won't have the same vreg assignments.

This cuts irgen time pretty significantly.  At this point most of the irgen time is spent inside LLVM or its data structures.  The overall effect is smaller than I hoped, but it should stack well with the other things we're doing:
```
           django_template3.py             2.3s (2)             2.3s (2)  -2.2%
                 pyxl_bench.py             2.2s (2)             2.1s (2)  -2.5%
     sqlalchemy_imperative2.py             2.5s (2)             2.4s (2)  -2.6%
                pyxl_bench2.py             1.2s (2)             1.2s (2)  -1.6%
             pyxl_bench_10x.py            16.8s (2)            16.5s (2)  -1.7%
       django_template3_10x.py            12.4s (6)            12.2s (6)  -1.3%
 sqlalchemy_imperative2_10x.py            18.0s (2)            17.9s (2)  -0.7%
            pyxl_bench2_10x.py            10.3s (2)            10.0s (2)  -2.1%
                       geomean                 5.3s                 5.2s  -1.8%
```

I think this also opens up the possibility for some memory improvements, since we can now store things as a contiguous vector rather than as a hashtable.  The tradeoff is that we end up having to store entries for unset-vregs as well, but I think it usually ends up working out (especially when we just want a bool per vreg).  I didn't measure this though.


src/analysis/fpc.h
```diff
@@ -87,43 +91,34 @@ void computeFixedPoint(typename BBAnalyzer<T>::Map&& initial_map, CFGBlock* init
...
+
+            Map& next = it->second;
+            // merge ending->next
             for (const auto& p : ending) {
-                if (next.count(p.first) == 0) {
-                    changed = true;
-                    if (initial) {
-                        next[p.first] = p.second;
-                    } else {
-                        next[p.first] = analyzer.mergeBlank(p.second);
```
undingen: can we get rid of `DefinednessBBAnalyzer::mergeBlank`?

src/analysis/function_analysis.h
```diff
@@ -50,60 +51,61 @@ class LivenessAnalysis {
...
-    bool isLiveAtEnd(InternedString name, CFGBlock* block);
+    bool isLiveAtEnd(int vreg, CFGBlock* block);
 };
 
 class PhiAnalysis;
 
 class DefinednessAnalysis {
 public:
     enum DefinitionLevel {
+        Unknown,
```
undingen: I think now that this VRegMap is an array it could save a lot of memory if we use `enum DefinitionLevel : char`

src/analysis/function_analysis.cpp
```diff
@@ -52,52 +52,49 @@ class LivenessBBVisitor : public NoopASTVisitor {
         }
     };
 
-    llvm::DenseMap<InternedString, Status> statuses;
+    VRegMap<Status> statuses;
```
undingen: usage can be `enum Usage : char`

src/analysis/function_analysis.h
```diff
@@ -50,60 +51,61 @@ class LivenessAnalysis {
...
-                CFGBlock* initial_block, bool initials_need_phis, LivenessAnalysis* liveness, ScopeInfo* scope_info);
-
-    bool isRequired(InternedString name, CFGBlock* block);
-    bool isRequiredAfter(InternedString name, CFGBlock* block);
-    const RequiredSet& getAllRequiredAfter(CFGBlock* block);
-    const RequiredSet& getAllRequiredFor(CFGBlock* block);
+    PhiAnalysis(VRegMap<DefinednessAnalysis::DefinitionLevel> initial_map, CFGBlock* initial_block,
+                bool initials_need_phis, LivenessAnalysis* liveness, ScopeInfo* scope_info);
+
+    // TODO: convert these to taking vregs
```
undingen: looks like this comments can be removed?!?

undingen:
nice cleanup! 
and probably one of the largest PR ever (except of the counting merge PR)!

PR #1284, opened 2016-07-06
Track down an elusive reference leak

kmod:
The pattern seems to be that we invalidate an IC (deregestering the decrefinfo),
but there is a stack frame in that IC and furthermore that stack frame will
get unwound via an exception.

The fix I did here is to not deregister the decrefinfo if num_inside>0.

@undingen could you give it a review?  I think this is enough to fix the issue but still allow the same amount of potential rewriting, but it's hard to be sure.

### DIFF
--- a/src/asm_writing/icinfo.cpp
+++ b/src/asm_writing/icinfo.cpp
@@ -75,8 +75,10 @@ void ICInvalidator::invalidateAll() {
 
 void ICSlotInfo::clear() {
     ic->clear(this);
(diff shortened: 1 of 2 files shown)


undingen:
Oh yeah this sounds like the leak!
Your fix looks good it's also how I thought about fixing it.

kmod:
Hopefully this fixes the test_telnetlib sporadic failures :)

PR #1286, opened 2016-07-07
Some SciPy fixing.

### LINKED ISSUE #1233: Two workaround that let cythonize scipy could run.
cython crashes when I try to run it on scipy. It needs these changes. Just for reminder.

 - Comment this `RELEASE_ASSERT`.
```
    for (auto&& var : { consts, names, varnames, freevars, cellvars })
        is_dummy = is_dummy && var == EmptyTuple;
    RELEASE_ASSERT(is_dummy, "not implemented");
    // ok this is an empty/dummy code object
```
Cythonize scipy will crashes with this:
```
python setup.py build
../../src/runtime/code.cpp:140: PyCodeObject *PyCode_New(int, int, int, int, PyObject *, PyObject *, PyObject *, PyObject *, PyObject *, PyObject *, PyObject *, PyObject *, int, PyObject *): Assertion `is_dummy' failed: not implemented
```
I have no more details, yet.

- reenable `typedef PyIntObject PyBoolObject;`
```
// Pyston change: Try to avoid having to support mixing ints and bools
// typedef PyIntObject PyBoolObject;
struct _PyBoolObject;
typedef struct _PyBoolObject PyBoolObject;
```

Will get this:
```
gcc: scipy/io/matlab/streams.c
scipy/io/matlab/streams.c: In function ‘initstreams’:
scipy/io/matlab/streams.c:6515:96: error: invalid application of ‘sizeof’ to incomplete type ‘PyBoolObject’
   __pyx_ptype_7cpython_4bool_bool = __Pyx_ImportType(__Pyx_BUILTIN_MODULE_NAME, "bool", sizeof(PyBoolObject), 0); if (unlikely(!__pyx_ptype_7cpython_4bool_bool)) __PYX_ERR(3, 8, __pyx_L1_error)
                                                                                                ^
scipy/io/matlab/streams.c: In function ‘initstreams’:
scipy/io/matlab/streams.c:6515:96: error: invalid application of ‘sizeof’ to incomplete type ‘PyBoolObject’
   __pyx_ptype_7cpython_4bool_bool = __Pyx_ImportType(__Pyx_BUILTIN_MODULE_NAME, "bool", sizeof(PyBoolObject), 0); if (unlikely(!__pyx_ptype_7cpython_4bool_bool)) __PYX_ERR(3, 8, __pyx_L1_error)
```
Should we use `typedef PyIntObject PyBoolObject;` again? 
And what is "mixing ints and bools" means  and the effect?

### LINKED ISSUE #1285: Support dummy members in BoxedCode?
Cython will trigger `PyCode_New` assert in here:
https://github.com/dropbox/pyston/blob/master/src/runtime/code.cpp#L140

```
extern "C" PyCodeObject* PyCode_New(int argcount, int nlocals, int stacksize, int flags, PyObject* code,
                                    PyObject* consts, PyObject* names, PyObject* varnames, PyObject* freevars,
                                    PyObject* cellvars, PyObject* filename, PyObject* name, int firstlineno,
                                    PyObject* lnotab) noexcept {
    // Check if this is a dummy code object like PyCode_NewEmpty generates.
    // Because we currently support dummy ones only.
    bool is_dummy = argcount == 0 && nlocals == 0 && stacksize == 0 && flags == 0;
    is_dummy = is_dummy && code == EmptyString && lnotab == EmptyString;
    for (auto&& var : { consts, names, varnames, freevars, cellvars })
        is_dummy = is_dummy && var == EmptyTuple;
    RELEASE_ASSERT(is_dummy, "not implemented");
    // ok this is an empty/dummy code object
```

SciPy need to call this function with empty bytecode but only have file attributes, probably for tracebacks.
https://github.com/cython/cython/commit/36780b2a

My idea is add some `PyCodeObject` members to `BoxedCode`. Just leave these as a "place holders". And remove the release assert. But trigger assert if the `code` parameter truelly contains something. Thoughts?

Daetalus:
Ready to review.

Some notes:
- I updated Cython from 0.22 to 0.24. Removed the old 0.22 patch and add a 0.24 patch. The main change is switch to CPython based Cython, not PyPy based Cython anymore(with some exceptions). And add some extended APIs to Cython. I will try to submit the new patch to Cython upstream in this week if it is ok for you first.
- I just update Cython 0.22 to 0.24 in lxml and numpy test. Didn't update Cython version in `README` and other files. Because it needs to update the patched Cython package's link to. So I left it to @kmod.
- Move `numpy_test.py` to `scipy_test.py` in `PYSTON/test/integration` dir. But disable it. It will took 51 mins on my laptop. We can test it manually...
- Also add `scipy_test` snippet to `numpy_full_test` not comment out in case some one want to test it locally. (maybe it is not necessary because we can do it by `make run_release_scipy_test`?)

closes #1285 
closes #1233


undingen:
The second commit it not correct:
- it has a number of exception mismatches. `method_descriptor` expects CAPI but several of this functions throw C++ exception
- but in general I don't think we want to fix this issue in this way instead I think we should special case the code in cython. If we really want to move to the capi method way I think we should switch at the same time to cpythons implementation of this functions. 

And from the cython code you linked it looks like this is a very good candidate for our extended capi project :-).
The code only want's to retrieve the `methoddescr_type`. We can directly expose it through a api in pyston or use something like `unicode.count` to retrieve it (instead of `list.append`).

test/extra/lxml_test.py
```diff
@@ -33,7 +33,7 @@ def install_and_test_lxml():
 
     subprocess.check_call([PYTHON_EXE, "setup.py", "build_ext", "-i", "--with-cython"], cwd=LXML_DIR)
  
-    expected = [{'ran': 1381, 'failures': 1}]
+    expected = [{'ran': 1381}]
```
undingen: :+1:

test/integration/scipy_test.py
```diff
@@ -0,0 +1,136 @@
...
+import os
+import sys
+import subprocess
+
+"""
+Using this test file.
+
+SciPy C source code files generated individually by Cython, it may contains
+implicit declaration of functions when compiling. It will trigger errors like:
+"error: implicit declaration of function ‘zcopy_’ " when building SciPy.
```
undingen: the ```‘``` character triggers:
```
File "test/integration/scipy_test.py", line 15
SyntaxError: Non-ASCII character '\xe2' in file test/integration/scipy_test.py on line 16, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

src/runtime/code.cpp
```diff
@@ -133,9 +133,9 @@ extern "C" PyCodeObject* PyCode_New(int argcount, int nlocals, int stacksize, in
                                     PyObject* lnotab) noexcept {
     // Check if this is a dummy code object like PyCode_NewEmpty generates.
     // Because we currently support dummy ones only.
-    bool is_dummy = argcount == 0 && nlocals == 0 && stacksize == 0 && flags == 0;
+    bool is_dummy = stacksize == 0;
     is_dummy = is_dummy && code == EmptyString && lnotab == EmptyString;
-    for (auto&& var : { consts, names, varnames, freevars, cellvars })
+    for (auto&& var : { consts, names, freevars, cellvars })
```
undingen: While I don't really like that this just pretends that we support this stuff I'm okay with this workaround for now because I think we will notice if code tries to retrieve the fields because it will crash because `BoxedCode::f` will be NULL.

Could you please update the comment because it does not match the check anymore.
I would suggest to leave the whole `is_dummy` computation unchanged and instead make the assert less strict by allowing the fields you enabled but adding a comment which says that it's not implemented for now.

test/integration/scipy_test.py
```diff
@@ -0,0 +1,136 @@
...
+
+This option was disabled in CPython. But enabled in Pyston. So if you want build
+and test SciPy in Pyston, please revert the second change in
+https://github.com/dropbox/pyston/commit/949dfeb1f675ef05ddfd04333e7e6c0f9887860c
```
undingen: while I don't like to disable `-Werror=implicit-function-declaration` for everyone just because there is some old fortran code around I would really like if one does not have to manually apply a patch to get it to compile.
Could we maybe overwrite this flag in scipy and upstream a patch? Or is there a nicer way to modify the flags?
@kmod what do you think?

test/integration/scipy_test.py
```diff
@@ -0,0 +1,136 @@
...
+    subprocess.check_call([PYTHON_EXE, "setup.py", "install"], cwd=SCIPY_DIR, env=env)
+except:
+    subprocess.check_call(["rm", "-rf", SCIPY_DIR + "/build"])
+    subprocess.check_call(["rm", "-rf", SCIPY_DIR + "/dist"])
+    raise
+
+scip_test = "import scipy; scipy.test(verbose=2)"
+
+print_progress_header("Running SciPy test suite...")
+subprocess.check_call([PYTHON_EXE, "-c", scip_test], cwd=CYTHON_DIR)
```
undingen: would be nice if the test would compare the test result to a expected number of successes+failures like we do for numpy even though the test does not get run automatically (I did not run the test yet because of missing fortran compiler - but I suspect it prints some stats otherwise ignore my comment please)

undingen:
thanks for working on the scipy support! :+1: 
I added some comments but did not yet take a look at the cython patch but will do so tomorrow

test/integration/scipy_test.py
kmod: My assumption has been that we would need to remove that flag eventually to bring ourselves in line with CPython.  But if there's some simple way to patch scipy to not need it I think that would be best for now :)

src/runtime/code.cpp
```diff
@@ -135,10 +135,12 @@ extern "C" PyCodeObject* PyCode_New(int argcount, int nlocals, int stacksize, in
     // Because we currently support dummy ones only.
     bool is_dummy = argcount == 0 && nlocals == 0 && stacksize == 0 && flags == 0;
     is_dummy = is_dummy && code == EmptyString && lnotab == EmptyString;
-    for (auto&& var : { consts, names, varnames, freevars, cellvars })
```
undingen: please leave here the `varnames` check in

src/runtime/code.cpp
```diff
@@ -135,10 +135,12 @@ extern "C" PyCodeObject* PyCode_New(int argcount, int nlocals, int stacksize, in
...
+    for (auto&& var : { consts, names, freevars, cellvars })
         is_dummy = is_dummy && var == EmptyTuple;
-    RELEASE_ASSERT(is_dummy, "not implemented");
-    // ok this is an empty/dummy code object
+    // The follwing variables are not implemented but we allow them because there is currently
+    // no way for code to retrieve them.
+    auto temp_allowed = argcount || argcount || flags || var != EmptyTuple;
```
undingen: `var != EmptyTuple` -> `varnames != EmptyTuple`

test/integration/scipy_test.py
```diff
@@ -0,0 +1,138 @@
...
+"""
+Using this test file.
+
+SciPy C source code files generated individually by Cython, it may contains
+implicit declaration of functions when compiling. Pyston set the flag
+"-Werror=implicit-function-declaration". It will trigger errors like:
+"error: implicit declaration of function 'zcopy_' " when building SciPy.
+This option was disabled in CPython. But enabled in Pyston. So if you want build
+and test SciPy in Pyston, please revert the second change in
+https://github.com/dropbox/pyston/commit/949dfeb1f675ef05ddfd04333e7e6c0f9887860c
```
Daetalus: For `implicit declaration of function` error, here is an example:
```
scipy/sparse/linalg/dsolve/SuperLU/SRC/clacon.c:157:5: error: implicit declaration of function ‘ccopy_’ [-Werror=implicit-function-declaration]
     ccopy_(n, x, &c__1, v, &c__1);
```

I took a look at SciPy source code in `scipy/sparse/linalg/dsolve/SuperLU`. It declared the function signature by `extern`. But the function implementation is in fortran code. If my understanding is correct. There has no other way except revert the change in that commit. Let the behavior more like CPython. Thoughts?

kmod: Hmm that doesn't sound like that would be the issue then :/

I looked into it a bit, and it looks like there are quite a lot of implicit-function-declaration warnings in SuperLU, to the point that I don't think it makes sense for us to try to fix up the scipy code.

This is hacky, but maybe we could do something like this?
```
if not any ('scipy' in s for s in sys.argv):
    cc_args = cc_args + ["-Werror=implicit-function-declaration"]
```

undingen:
Thanks for getting SciPy running! :+1:

src/runtime/code.cpp
```diff
@@ -137,8 +137,10 @@ extern "C" PyCodeObject* PyCode_New(int argcount, int nlocals, int stacksize, in
     is_dummy = is_dummy && code == EmptyString && lnotab == EmptyString;
     for (auto&& var : { consts, names, varnames, freevars, cellvars })
         is_dummy = is_dummy && var == EmptyTuple;
-    RELEASE_ASSERT(is_dummy, "not implemented");
-    // ok this is an empty/dummy code object
+    // The follwing variables are not implemented but we allow them because there is currently
+    // no way for code to retrieve them.
+    auto temp_allowed = argcount || argcount || flags || varnames != EmptyTuple;
+    RELEASE_ASSERT(is_dummy || temp_allowed, "not implemented");
```
kmod: I don't think this check is quite right -- we don't want to allow a non-dummy code object just because the caller requested one of the temp_allowed fields.  Can we just remove these from the is_dummy calculation?

PR #1290, opened 2016-07-11
use r12, r15 and rbx in bjit and inside bjit ICs, remove some nonzero checks

undingen:
Small bjit and rewriter improvements
- use r12, r15 and rbx in bjit and inside bjit ICs
- remove some nonzero calls when we already know that the object is a python bool
- don't check if the arguments of a function are nonzero in the entry block

Also changes the way we keep track of which registers are available inside the rewriter:
Keeping the available registers in a bitset makes it more memory efficient and
also easier and faster to calculate a subset of the registers.
I will soon implement the 'otherThan' functionality using it which would fix the current problem of only allowing to exclude one register.

### DIFF
--- a/src/asm_writing/types.h
+++ b/src/asm_writing/types.h
@@ -35,6 +35,7 @@ namespace assembler {
 
 class Assembler;
 
+struct RegisterSet;
 struct Register {
     int regnum;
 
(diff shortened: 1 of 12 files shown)


src/asm_writing/types.h
```diff
@@ -53,22 +53,87 @@ struct Register {
...
+
+    constexpr explicit RegisterSet(Regs regs) : regs(regs) {}
+    constexpr explicit RegisterSet(Register reg) : regs(1ul << reg.regnum) {}
+
+    static constexpr int max_reg_num = Register::numRegs();
+    static constexpr Regs callee_save_mask = (1ul << RBX.regnum) | (1ul << RSP.regnum) | (1ul << RBP.regnum)
+                                             | (1ul << R12.regnum) | (1ul << R13.regnum) | (1ul << R14.regnum)
+                                             | (1ul << R15.regnum);
+
+    static constexpr RegisterSet std() {
```
kmod: It might be nice to call this something more descriptive -- "RegisterSet.std" doesn't quite say what it is, so it might be nice to call this something like 'stdAllocatable'?

undingen:
I fixed a few additional assembler bugs, cleaned up the code and allowed the bjit+ICs to use rsp

PR #1291, opened 2016-07-11
VRegSet: use a llvm bitvector as representation

undingen:
This should be a little bit more efficient for large number of vregs.

### DIFF
--- a/src/core/cfg.h
+++ b/src/core/cfg.h
@@ -29,6 +29,8 @@
 
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


kmod:
I think `vector<bool>` is supposed to use a bitset internally, but [LLVM's documentation](http://llvm.org/docs/ProgrammersManual.html#bit-storage-containers-bitvector-sparsebitvector) has some information on why they don't think it's good, so :+1: to this PR.

Issue #1197, opened 2016-05-22
add missing tp_as_mapping, tp_as_sequence and tp_as_number functions


undingen:
Some of our builtin types don't overwrite all of this functions. 
This means the perf will be slightly worse because we will call into the slot wrapper which will than do an attribute lookup instead of just calling the function pointer.
One example is our `list` type which does not initialize the `tp_as_mapping` function pointers.
They are therefore initialized to `slot_sq_length`, etc.. while we should overwrite them with custom implementations like `list_length`.

undingen:
I added https://github.com/dropbox/pyston/pull/1198 as an example for such a change.

sizeoftank:
For example ` tp_as_mapping`  in  ` list `,  I think it could completed by copy the implementation from cpython, right?

kmod:
Our list object is very similar to CPython's, so that sounds like a good place to start even if there are small things that might have to be adjusted.

PR #1294, opened 2016-07-13
Add sq_slots & mp_slots to instance (issue #1197)

sizeoftank:
implement  ` tp_as_sequence ` &  ` tp_as_mapping ` for  ` instance_cls `

### DIFF
--- a/src/runtime/classobj.cpp
+++ b/src/runtime/classobj.cpp
@@ -723,14 +723,107 @@ Box* instanceLen(Box* _inst) {
     return runtimeCall(len_func, ArgPassSpec(0), NULL, NULL, NULL, NULL, NULL);
 }
 
-Box* instanceGetitem(Box* _inst, Box* key) {
+// Analoguous to CPython's, used for sq_slots, mp_slots
+static Py_ssize_t instance_length(PyObject* self) noexcept {
+    PyObject* func;
+    PyObject* res;
+    Py_ssize_t outcome;
+
+    static BoxedString* len_str = getStaticString("__len__");
+
+    func = instance_getattro(self, len_str);
+    if (func == NULL)
+        return -1;
+
+    res = runtimeCallCapi(func, ArgPassSpec(0), NULL, NULL, NULL, NULL, NULL);
+    Py_DECREF(func);
+
+    if (res == NULL)
+        return -1;
+
+    if (PyInt_Check(res)) {
+        outcome = PyInt_AsSsize_t(res);
+        if (outcome == -1 && PyErr_Occurred()) {
+            Py_DECREF(res);
+            return -1;
+        }
+#if SIZEOF_SIZE_T < SIZEOF_INT
+        /* Overflow check -- range of PyInt is more than C int */
+        if (outcome != (int)outcome) {
+            PyErr_SetString(PyExc_OverflowError, "__len__() should return 0 <= outcome < 2**31");
+            outcome = -1;
+        } else
+#endif
+            if (outcome < 0) {
+            PyErr_SetString(PyExc_ValueError, "__len__() should return >= 0");
+            outcome = -1;
+        }
+
+    } else {
+        PyErr_SetString(PyExc_TypeError, "__len__() should return an int");
+        outcome = -1;
+    }
+    Py_DECREF(res);
+    return outcome;
+}
+
+
+// Analoguous to CPython's, used for mp_slots.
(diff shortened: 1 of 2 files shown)


src/runtime/classobj.cpp
```diff
@@ -723,14 +723,115 @@ Box* instanceLen(Box* _inst) {
...
+            return NULL;
+        }
+    } else {
+        getitem_func = _instanceGetattribute(inst, getitem_str, true);
+    }
+
     AUTO_DECREF(getitem_func);
-    return runtimeCall(getitem_func, ArgPassSpec(1), key, NULL, NULL, NULL, NULL);
+    try {
+        return runtimeCall(getitem_func, ArgPassSpec(1), key, NULL, NULL, NULL, NULL);
```
undingen: I think you can replace the whole try catch block with
```
return runtimeCallInternal<S, NOT_REWRITABLE>(getitem_func, NULL, ArgPassSpec(1), key, NULL, NULL, NULL, NULL);
```

src/runtime/classobj.cpp
```diff
@@ -743,6 +844,39 @@ Box* instanceSetitem(Box* _inst, Box* key, Box* value) {
     return runtimeCall(setitem_func, ArgPassSpec(2), key, value, NULL, NULL, NULL);
 }
 
+static int instance_ass_item(Box* inst, Py_ssize_t k, PyObject* item) noexcept {
+    static BoxedString* delitem_str, *setitem_str;
+    Box* func = NULL, * res = NULL;
+    Box* key = boxInt(k);
+
+    if (item == NULL) {
+        delitem_str = getStaticString("delitem_str");
```
undingen: `"__delitem__"`

src/runtime/classobj.cpp
```diff
@@ -840,6 +974,78 @@ Box* instanceDelslice(Box* _inst, Box* i, Box* j) {
...
+
+    if (value == NULL) {
+        delslice_str = getStaticString("__delslice__");
+        func = instance_getattro(inst, delslice_str);
+        if (func == NULL) {
+            if (!PyErr_ExceptionMatches(PyExc_AttributeError))
+                return -1;
+            PyErr_Clear();
+
+            delitem_str = getStaticString("delitem_str");
```
undingen: `"__delitem__"`

undingen:
thanks for the patch! :+1: 
I added a few comments, I'm a little bit worried about the additional code duplication this introduces but it does not make things much worse than the old class support already is :-)
I guess we really just have to replace the whole implementation some point in the future with cpythons but for now this is a improvement.

sizeoftank:
thanks for your response! 👍 
oh, these issues aroused by my careless, I make a change to fix them. :-)
I also think we should refactor our codes to make it better compatibility with cpython, but I not sure I can complete these works correctly now.

src/runtime/classobj.cpp
```diff
@@ -840,6 +966,78 @@ Box* instanceDelslice(Box* _inst, Box* i, Box* j) {
     }
 }
 
+// Analoguous to CPython's, used for sq_slots.
+static int instance_ass_slice(Box* inst, Py_ssize_t i, Py_ssize_t j, PyObject* value) {
+    static BoxedString* delslice_str, *setslice_str, *delitem_str, *setitem_str;
+    Box* func = NULL, * res = NULL;
+    Box* slice = NULL;
+    Box* begin = boxInt(i);
```
undingen: `begin` and `end` will get leaked at the `return -1` statments.
better use `AUTO_DECREF(begin);`

src/runtime/classobj.cpp
```diff
@@ -840,6 +966,78 @@ Box* instanceDelslice(Box* _inst, Box* i, Box* j) {
...
+            }
+            res = runtimeCallCapi(func, ArgPassSpec(3), begin, end, (Box*)value, NULL, NULL);
+        }
+    }
+
+    Py_DECREF(begin);
+    Py_DECREF(end);
+    Py_DECREF(func);
+
+    if (slice == NULL) {
```
undingen: this check is strange because if it is NULL it would have already got used....

sizeoftank: It is a mistake that I copied the cpython implementation to here (it use ` arg = Py_BuildValue() ` every path), I think this check can be removed and change to AUTO_DECREF(slice) when I create it.

sizeoftank:
I stash these into one, other changes need todo?

undingen:
LGTM, thanks for the patch!

PR #1298, opened 2016-07-18
[WIP] Tk support

Daetalus:
There have several reasons that I add tk support in Pyston:
- Matplotlib can use variety backends, such as tk, PyGTK, PyQt, wxPython, kivy. tk is CPython "standard" GUI frame work. It give them the highest priority.
- Also Scikit-learn and SciPy could work if without matplotlib. However, it will be good if we support it if it not hard to support(I mean, not let me lose the focus).
- Visualize the achievement we got.


from_cpython/Modules/getpath.c
```diff
@@ -591,15 +592,20 @@ calculate_path(void)
     bufsz += strlen(zip_path) + 1;
     bufsz += strlen(exec_prefix) + 1;
 
-    // Pyston change: add from_cpython/Lib and lib_pyston
+    // Pyston change: add from_cpython/Lib, lib_pyston and Lib/lib-tk
```
kmod: I'm pretty skeptical of this change -- we shouldn't have to add any libraries here.  Can we just see how CPython ends up not needing this, and try to copy it?

.travis.yml
```diff
@@ -53,9 +53,11 @@ addons:
     - python-dev
     - texlive-extra-utils
     - libcurl4-openssl-dev
-    - libxml2-dev 
+    - libxml2-dev
     - libxslt1-dev
     - libssl-dev
+    - tcl8.5-dev
+    - tk8.5-dev
```
kmod: We'll also need to add any new required libraries to docs/INSTALLING.md

Issue #1191, opened 2016-05-20
Fix evaluation order for dict operations


kmod:
See [literals_order.py](https://github.com/dropbox/pyston/blob/master/test/tests/literals_order.py)

For dictionaries literals like `{a:b, c:d}`, we first evaluate all of the subexpressions, and then assemble the dict.  We should switch to CPython's order, which is incrementally adding the key-value pairs to the dict as we evaluate them.

sizeoftank:
The test literals_order.py has been removed ?
File blob/refcounting/test/tests/literals_order.py not found

Daetalus:
No, just the `refcounting` branch was removed. Please see the file in `master` branch. https://github.com/dropbox/pyston/blob/master/test/tests/literals_order.py

undingen:
I updated the link in the original comment from kmod to point to master

sizeoftank:
It looks like in create the temporary variable in ` remapDict() ` , I try to set ` wrap_with_assign = false ` and it runs the correct order. 
But a strange thing happened when I running  ` make check_literals_order `, It outputs this differ:
```
Traceback (most recent call last):
  File "./tools/tester.py", line 456, in worker_thread
  File "./tools/tester.py", line 172, in run_test
  File "./tools/tester.py", line 359, in determine_test_result
Exception: Failed on /home/chenwei/pyston/test/tests/literals_order.py:
--- /tmp/expected_FDg4Bo	2016-07-16 06:15:35.835442174 +0800
+++ /tmp/received__NL8z3	2016-07-16 06:15:35.835442174 +0800
@@ -5,15 +5,15 @@
 hashing 1L
 hashing 2
 hashing 1
 eq
 testing dicts
-Creating 4
 Creating 1
+Creating 4
 hashing 1
-Creating 5
 Creating 2
+Creating 5
 hashing 2
-Creating 6
 Creating 1L
+Creating 6
 hashing 1L
 eq
```

sizeoftank:
Oh I forgot to update IR compiler in previous.

PR #1302, opened 2016-07-19
Fix evaluation order for dict operations

sizeoftank:
Fix literals_order issue #1191
not create temporary variables when remapExpr

### DIFF
--- a/test/tests/literals_order.py
+++ b/test/tests/literals_order.py
@@ -1,5 +1,3 @@
-# expected: fail
-# - the CFG expands to the wrong code here
 
 class H(object):
     def __init__(self, n):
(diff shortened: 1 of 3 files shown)


kmod:
I don't think this is the best solution here :(  The point of this pass is to un-nest all expressions, and along the way explicitly encode the evaluation order.  So while I think "keeping it nested" might work it's not really what I was hoping for.  There are probably also going to be weird corner cases if you start throwing in other features such as exceptions and control-flow-in-expressions (what if a dict key is something like `1/0 or None`).

Can we have remapDict generate something similar to this:
```
#1 = {}
#2 = <key1>
#3 = <value1>
#1[#2] = #3
#4 = <key2>
#5 = <value2>
#1[#4] = #5
```

sizeoftank:
Thanks for your help, it sounds like good !
I will try do these in few days :-)

sizeoftank:
Hi @kmod 
Please take a look at this if you have the time

test/tests/literals_order.py
```diff
@@ -1,4 +1,3 @@
-# expected: fail
 # - the CFG expands to the wrong code here
```
kmod: Minor, but could you remove this comment too?

kmod:
Looks great!  Would you mind removing that one comment I mentioned and then squashing the two commits together?

sizeoftank:
okay, I updated that :-)

kmod:
Awesome, thanks!

PR #1303, opened 2016-07-20
Replace None with Py_None and some other fixing.

Daetalus:
Spit it from tk_support PR. I will improve that PR on weekend.

### DIFF
--- a/src/runtime/builtin_modules/sys.cpp
+++ b/src/runtime/builtin_modules/sys.cpp
@@ -48,7 +48,7 @@ Box* sysExcInfo() {
     ExcInfo* exc = getFrameExcInfo();
     assert(exc->type);
     assert(exc->value);
-    Box* tb = exc->traceback ? exc->traceback : None;
+    Box* tb = exc->traceback ? exc->traceback : Py_None;
     return BoxedTuple::create({ exc->type, exc->value, tb });
 }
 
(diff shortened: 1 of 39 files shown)


src/runtime/builtin_modules/thread.cpp
```diff
@@ -227,6 +233,9 @@ void setupThread() {
                                              FunctionMetadata::create((void*)getIdent, BOXED_INT, 0), "get_ident"));
     thread_module->giveAttr("stack_size", new BoxedBuiltinFunctionOrMethod(
                                               FunctionMetadata::create((void*)stackSize, BOXED_INT, 0), "stack_size"));
+    thread_module->giveAttr(
```
undingen: this needs a test

src/runtime/builtin_modules/sys.cpp
```diff
@@ -816,6 +816,8 @@ void setupSys() {
     sys_flags_cls->giveAttr(
         "__new__", new BoxedFunction(FunctionMetadata::create((void*)BoxedSysFlags::__new__, UNKNOWN, 1, true, true)));
     sys_flags_cls->tp_dealloc = (destructor)BoxedSysFlags::dealloc;
+
+    sys_flags_cls->giveAttr("ignore_environment", boxInt(Py_IgnoreEnvironmentFlag));
```
undingen: I don't like this half implemented `sys.flags` thing but I don't think you change makes it worse so I'm fine with merging this but I created https://github.com/dropbox/pyston/issues/1304 to make sure we fix it at some point

Daetalus:
Working on scikit-learn support. So move the `thread.interrupt_main` to `tk_support` PR. Will take a look on weekend.

undingen:
thanks for the change! :+1:

PR #1308, opened 2016-07-27
[WIP]Scikit-learn commit

Daetalus:


undingen:
It looks like your patch is improving the situation but I really would like to switch to cpythons list sorting. 
I think ours get's some corner cases wrong and does not support having `cmp` and `key` set at the same time (and we run into this several times already on real code so it is important). So if you have time and feel like fixing it "forever" I would suggest switching directly to the cpython code (I guess best with the hybrid `listobject.c` and `list.cpp` approach because is very long). 
But only if you have time otherwise a small fix for the current implementation is fine too :-).

kmod:
+1 to migrating to CPython's sort if you have time (we're going to have to do it at some point).

Also, we're running into this issue in other places as well, so it'd be great if we could separate it from the scipy PR and potentially get it in sooner :)

src/runtime/set.cpp
```diff
@@ -871,6 +871,14 @@ extern "C" PyObject* PyFrozenSet_New(PyObject* iterable) noexcept {
     }
 }
 
+extern "C" PyObject* PySet_Pop(PyObject* set) noexcept {
+    if (!PySet_Check(set)) {
+        PyErr_BadInternalCall();
+        return NULL;
+    }
+    return setPop(static_cast<BoxedSet*>(set));
```
undingen: this should be ```return callCXXFromStyle(CAPI, setPop, static_cast<BoxedSet*>(set));```

PR #1309, opened 2016-07-28
enable `PyObject_Format` in `from_cpython/Objects/abstract.c`

aisk:
found that the function in abstract.cpp have no changes with the origin CPython version :)


aisk:
Have some problems with commit history, ignore this PR plz.

PR #1310, opened 2016-07-29
Switch to CPython list sort(Not its list implementation)

Daetalus:


src/runtime/inline/list.cpp
```diff
@@ -131,15 +131,15 @@ const int BoxedList::INITIAL_CAPACITY = 8;
 // TODO the inliner doesn't want to inline these; is there any point to having them in the inline section?
 void BoxedList::shrink() {
     // TODO more attention to the shrink condition to avoid frequent shrink and alloc
-    if (capacity > size * 3) {
-        int new_capacity = std::max(static_cast<int64_t>(INITIAL_CAPACITY), capacity / 2);
+    if (allocated > size * 3) {
```
undingen: I would prefer if the rename `capacity` to `allocated` change is in it's own commit without behaviour changes.

src/runtime/inline/list.h
```diff
@@ -48,12 +48,12 @@ extern "C" inline void listAppendInternalStolen(Box* s, Box* v) {
     assert(PyList_Check(s));
     BoxedList* self = static_cast<BoxedList*>(s);
 
-    assert(self->size <= self->capacity);
```
undingen: I think we wan't to keep this assert but it should instead be `assert(self->size <= self->capacity || self->capacity == -1);`

src/runtime/inline/list.h
```diff
@@ -48,12 +48,12 @@ extern "C" inline void listAppendInternalStolen(Box* s, Box* v) {
...
     self->ensure(1);
 
-    assert(self->size < self->capacity);
+    assert(self->size < self->allocated);
     self->elts->elts[self->size] = v;
     self->size++;
+    self->allocated = self->size;
```
undingen: this looks wrong :-( - why did you have to add this?
I think this is will force us to do much more reallocs (after every append)
I checked the perf numbers and it looks like this PR is a quite a perf regression which would be explainable by this change.

from_cpython/Objects/listobject.c
```diff
@@ -109,6 +109,9 @@ PyList_Fini(void)
     }
 }
 
+static PyObject *indexerr = NULL;
+// Pyston change: comment this out
+#if 0
```
undingen: It looks like there are now a lot of functions enabled in this file which we do not use.
I would really prefer if you just put a giant `#if 0` around the file and only enable the list sorting part

undingen:
thank you very much for fixing this! :+1:

PR #1312, opened 2016-07-31
using cpython's `sys.flags` inplementation

aisk:


kmod:
Looks great!

If you want to make it even nicer you could add a test that now passes (sounds like `print sys.flags` now works).

PR #1314, opened 2016-08-02
hidden classes: split into subclasses to reduce memory consumption

undingen:
+ use pyston::DenseMap to save a little more memory

this saves about 5%-10% of peak memory on django


kmod:
:+1:

PR #1316, opened 2016-08-03
ScopeNameUsage merge dicts into a single big one

undingen:
and change it to use a unordered_map because it uses less memory in this case and is faster
(I assume because it does not have align the key value tuples)
saves about 10% of peak memory on django


kmod:
:+1:

I wonder if it might be better though to just go for "caching at the CFG level instead of AST" and get rid of lazy CFG computation (which is I think why we have to keep the scoping analysis around).

PR #1317, opened 2016-08-05
delete the llvm module after code generation

undingen:
this saves a lot of memory


kmod:
Man, you've already saved 25% of django_template memory!

I hooked up the perf tester to report memory usage -- it's still backfilling the data but so far [the graph looks quite nice](http://speed.pyston.org/timeline/#/?exe=1,3,7&base=none&ben=django_template3_10x_maxrss&env=2&revs=50&equid=off).

PR #1319, opened 2016-08-05
[WIP] make AST nodes smaller

undingen:


undingen:
closing; outdated we switched to a proper bytecode on master

PR #1321, opened 2016-08-08
use more often std::unique_ptr

undingen:


kmod:
nice :)  this has been some long-standing technical debt.

PR #1331, opened 2016-08-17
string interning: make it slightly more space efficient

undingen:
We used to store the string content twice.
This implementation makes use of `DenseSet::find_as` functionality where one can search using a different type.
I was not sure if the special `DenseMapInfo` I had to create causes any problems so I choose to create a new source file for it so that it does not get picked up somewhere else.

---

mmh I'm not sure if we want to merge this change: 
- it looks like it's a much smaller memory usage improvement than I thought (only a few hundred kb) :-(
- perf is also about the same (maybe very slightly better)

@kmod what do you think?


undingen:
ok I decided to merge this, our server codebase has much more interned strings so this is a bigger win there.

PR #1333, opened 2016-08-19
switch from the SSE prefetch to __builtin_prefetch

undingen:
`_mm_prefetch(next, _MM_HINT_T0)` expands to `__builtin_prefetch(next, 0, 3)` which are the defaults of `__builtin_prefetch(next)` too.
Reason for this change is that we did not include `xmmintrin.h` and instead of pulling this header in everywhere I thought it's better too just use the compiler intrinsic directly


kmod:
oh nice, good catch.

PR #1335, opened 2016-08-22
Misc memory size reductions

undingen:
saves between 5-10% with a very slight speed up.

### DIFF
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -19,6 +19,7 @@
 // over having them spread randomly in different files, this should probably be split again
 // but in a way that makes more sense.
 
+#include <forward_list>
 #include <memory>
 #include <stddef.h>
 #include <vector>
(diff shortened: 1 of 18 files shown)


src/core/types.h
```diff
@@ -217,31 +219,61 @@ static_assert(sizeof(ArgPassSpec) <= sizeof(void*), "ArgPassSpec doesn't fit in
...
 
-    // This members are only set if the InternedStringPool& constructor is used (aka. source is available)!
-    // They are used as an optimization while interpreting because the AST_Names nodes cache important stuff
-    // (InternedString, lookup_type) which would otherwise have to get recomputed all the time.
-    std::vector<AST_Name*> arg_names;
-    AST_Name* vararg_name, *kwarg_name;
+        const char* str;
+        AST_Name* name;
+    };
+    std::vector<NameOrStr> all_args;
```
kmod: It's not the biggest win, but one thing we could do is create a non-growable container to use instead of a fixed-size std::vector.  I guess the savings aren't that huge (just a pointer per vector) but sometimes it seems silly to have a std::vector when we don't use any of the capabilities.

PR #1337, opened 2016-08-22
Remove libpypa parser

Daetalus:
I just want to reduce the number of affecting factor when I try to support Python 3...


src/codegen/parser.cpp
```diff
@@ -1029,13 +1030,6 @@ AST_Module* parse_file(const char* fn, FutureFlags inherited_flags) {
...
 }
 
-const char* getMagic() {
-    if (ENABLE_PYPA_PARSER)
-        return "a\ncR";
-    else
-        return "a\nCR";
-}
-
 #define MAGIC_STRING_LENGTH 4
```
undingen: could you please change this to
`const int MAGIC_STRING_LENGTH = 4` and move it right beneath `const char* MAGIC = "a\nCR";`

undingen:
I agree that it is probably now time to remove it:
- it was since a long time disabled per default
- we support the cpython parser fully which has the best compatibility
- it does not make sense too have two parsers anymore

goodbye pypa :crying_cat_face: thanks @vinzenz again!
pypa helped us a lot for a long time

vinzenz:
@undingen np I am glad that I have been of service ;-)

kmod:
I guess it's time, but sad to see this go :(

PR #1340, opened 2016-08-25
These cpython tests are working now

kmod:


from_cpython/Modules/bz2module.c
```diff
@@ -1420,6 +1420,9 @@ BZ2File_init(BZ2FileObject *self, PyObject *args, PyObject *kwargs)
 
     mode = (mode_char == 'r') ? "rb" : "wb";
 
+    // Pyston change: this prevents leaking refs in the test_bz2.py test
+    Py_CLEAR(self->file);
```
Daetalus: Would you mind to tell a bit about how did you locate this error, please?

kmod: Sure!  So running the test gave this output:
```
Leaked references:
0x2aaaaec1f640 [1] str     watch -l ((PyObject*)0x2aaaaec1f640)->ob_refcnt
0x2aaaaab2bdc0 [1] file     watch -l ((PyObject*)0x2aaaaab2bdc0)->ob_refcnt
0x2aaaaec1f8e0 [1] str     watch -l ((PyObject*)0x2aaaaec1f8e0)->ob_refcnt
0x2aaaaab2be80 [1] file     watch -l ((PyObject*)0x2aaaaab2be80)->ob_refcnt
0x2aaaadb9e040 [2] str     watch -l ((PyObject*)0x2aaaadb9e040)->ob_refcnt
0x2aaaaaae9040 [4] NoneType     watch -l ((PyObject*)0x2aaaaaae9040)->ob_refcnt
```

I started off by looking into the leaked file objects, since they can point to some of the other ones (our leak checker is supposed to catch that but it can't always).  Also leaked strings and None are harder to track down since they tend to get reused.

So then I used gdb with the "watch" command that it suggested, and it showed that the file was being created here.  So the file was being created in `__init__` but not decreffed, so then there's a couple things to check: does the parent object get freed, and if so does it decref the child object.  In this case the BZ2File did get deallocated, and the destructor decrefs `self->file`, but at the time of the destructor the `file` field pointed to a different object.  Then it's a matter of putting a watchpoint on `self->file` to see when it gets changed, and in this case it got changed again in `__init__`.

This is actually a fairly common issue, that `__init__` functions in CPython will leak references if called more than once, so I might have guessed it was that at some point.

Daetalus: ah, I see now, thanks for your explanation!

PR #1342, opened 2016-08-26
[WIP] delete FunctionMetadata, SourceInfo and AST nodes

undingen:


undingen:
closing: we fixed this in a nicer way on master

PR #1350, opened 2016-09-01
Let Pyston compatible with GCC 5.x

Daetalus:


undingen:
thanks for the patch!

PR #1352, opened 2016-09-01
Get rid of FunctionMetadata

kmod:
Instead, merge it into BoxedCode.  This way we'll get refcounting
of our codegen datastructures.

I didn't do anything more than inline all of FunctionMetadata's fields
into BoxedCode, and then delete the FunctionMetadata object.


kmod:
Submitted as part of #1353

PR #1354, opened 2016-09-05
Switch to CPython way to build standard C extensions.

Daetalus:
Copy CPython setup.py and add lots of Pyston changes.

### DIFF
--- a/from_cpython/setup.py
+++ b/from_cpython/setup.py
@@ -8,9 +8,6 @@ from glob import glob
 from platform import machine as platform_machine
 import sysconfig
 
-from distutils import log
-from distutils import text_file
-from distutils.errors import *
 from distutils.core import Extension, setup
 from distutils.command.build_ext import build_ext
 from distutils.command.install import install
@@ -29,11 +26,6 @@ def get_platform():
     return sys.platform
 host_platform = get_platform()
 
-# Were we compiled --with-pydebug or with #define Py_DEBUG?
-COMPILED_WITH_PYDEBUG = ('--with-pydebug' in sysconfig.get_config_var("CONFIG_ARGS"))
-
-# This global variable is used to hold the list of modules to be disabled.
-disabled_module_list = []
 
 def add_dir_to_list(dirlist, dir):
     """Add the directory 'dir' to the list 'dirlist' (at the front) if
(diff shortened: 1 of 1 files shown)


from_cpython/CMakeLists.txt
```diff
@@ -158,21 +156,7 @@ add_library(FROM_CPYTHON OBJECT ${STDMODULE_SRCS} ${STDOBJECT_SRCS} ${STDPYTHON_
 add_dependencies(FROM_CPYTHON copy_stdlib)
 
 set(STDMODULES
-    ${CMAKE_BINARY_DIR}/lib/python2.7/lib-dynload/bz2.pyston.so
```
kmod: Just curious, why do you want to remove these lines?  It's definitely annoying that we have to duplicate all this between the setup.py and CMakeLists.txt (and same with the source files), but it means that our cmake system will end up knowing what the setup.py will do.  I think concretely the features it gets us are 1) not having to run setup.py if none of the dependencies change, and 2) properly packaging the .so's into a release tarball.

How would you feel about keeping these definitions + potentially adding any new ones from the new setup.py?  We could probably handle the .c files by using a glob_recurse, but we might have to list out all the .so files we are expecting.

from_cpython/CMakeLists.txt
```diff
@@ -181,30 +165,6 @@ add_custom_command(OUTPUT ${STDMODULES}
                        pyston
                        copy_stdlib
```
kmod: Minor, but while we're here -- we should add "setup.py" to this list.

from_cpython/CMakeLists.txt
Daetalus: ok, I am fine with keep these definitions. And what about comment out bunch of codes inside setup.py. My idea is enable those commented code when necessary. For example, this PR comment out the `tk` related code. And we can enable it in my `tk_support` PR. But not enable it for now. What do you think?

kmod: First of all, this is minor, but let's try to always copy from v2.7.7 of CPython.  There aren't that many changes between the micro versions so it's not that huge of a deal, but it helps us if we have a convention around it (for example if we try to find any remaining Pyston changes).

I think changing setup.py is perfectly fine if we don't support all of it yet.  My only recommendation would be to try to minimize the size of the diff so that it's more understandable (1500 changed lines is quite a lot).  For example, if we are commenting out the calls to certain functions, we probably don't need to comment out the functions themselves.  And if we don't support some feature such as multi-platform, instead of commenting out all of that code we could maybe just hard-code the platform.

Daetalus:
Hi, @kmod, after some simplification, the second patch's diff deduced `from +1,753 −1,682` to `+434 −369`. The remaining diffs mostly are the disabled extensions. I think we can enable them one by one in different PRs. This PR is ready for review now.

kmod:
Awesome, really nice improvement :)

PR #1358, opened 2016-09-06
Merge AST/BST work into master

kmod:
It still has some memory regressions, but now everything is in place.  I could see the argument etiher way on whether it should be merged back yet, but I'm thinking let's err on the side of simplicity and do the merge.


undingen:
I merged this branch now, while it still increases the memory usage it's now not as huge anymore because we delete the AST nodes.
I will incrementally reduce the BST memory usage soon.

PR #1364, opened 2016-09-13
[WIP]Some other building issue

Daetalus:


from_cpython/Lib/distutils/unixccompiler.py
```diff
@@ -246,6 +246,7 @@ def find_library_file(self, dirs, lib, debug=0):
         shared_f = self.library_filename(lib, lib_type='shared')
         dylib_f = self.library_filename(lib, lib_type='dylib')
         static_f = self.library_filename(lib, lib_type='static')
+        non_pyston_shared_f = '.'.join(shared_f.split('.')[:-2] + ['so'])
```
undingen: I would prefer if this would explicitly check if `shared_f` ends in `.pyston.so` and only than remove it.
+ comments which mention that this is a pyston change before every change please

Makefile
```diff
@@ -155,11 +152,6 @@ COMMON_LDFLAGS += `pkg-config tinfo 2>/dev/null && pkg-config tinfo --libs || ec
 # TODO should probably do the linking before MCJIT
 COMMON_LDFLAGS += -Wl,-E
 
-# We get multiple shared libraries (libstdc++, libgcc_s) from the gcc installation:
```
undingen: I think instead of removing parts of the unused makefile code in this commit we should leave it in and instead remove the whole non cmake related build code from the makefile sometime in the future.
If it really does not work without removing `ifneq ($(GCC_DIR),/usr)` than I guess there is more stuff used than I thought I will have review it more carefully again.

from_cpython/setup.py
```diff
@@ -743,6 +743,10 @@ def detect_modules(self):
             missing.extend(['imageop'])
 
         # readline
+        # Pyston change: in some container system, the dependencies are provided by
```
undingen: mmh I'm not so sure about this change, I don't think it hurts but I'm not sure and would prefer if you server would not need this.
I found cpythons https://bugs.python.org/issue21571 but it's still open :-(.
@kmod what do you think about this change?

Daetalus: CPython uses `CPPFLAGS` and `LDFLAGS`. https://github.com/Daetalus/cpython/blob/master/configure#L2796

So I can also modify the CMakefile, for example(The `$ENV{CPPFLAGS}`):

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e660dc3..2250f89 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,7 +273,7 @@ if(ENABLE_OPROFILE)
 endif()
 
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Werror -Wreturn-type -Wno-sign-compare -Wno-unused -Wno-unused-parameter -fno-omit-frame-pointer -g")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CMAKE_C_FLAGS} -std=c++11 -fno-rtti -fexceptions -fvisibility-inlines-hidden -ffunction-sections -fdata-sections -Woverloaded-virtual -Wno-invalid-offsetof -Wcast-qual -Wno-sign-conversion -Wnon-virtual-dtor -Winit-self -Wmissing-include-dirs -Wstrict-overflow=5 -Wpointer-arith -Wtype-limits -Wwrite-strings -Wempty-body -Waggregate-return -Wmissing-field-initializers -Wredundant-decls -Winline -Wint-to-pointer-cast -Wlong-long -Wvla -Wno-attributes -g")
+set(CMAKE_CXX_FLAGS "$ENV{CPPFLAGS} ${CMAKE_CXX_FLAGS} ${CMAKE_C_FLAGS} -std=c++11 -fno-rtti -fexceptions -fvisibility-inlines-hidden -ffunction-sections -fdata-sections -Woverloaded-virtual -Wno-invalid-offsetof -Wcast-qual -Wno-sign-conversion -Wnon-virtual-dtor -Winit-self -Wmissing-include-dirs -Wstrict-overflow=5 -Wpointer-arith -Wtype-limits -Wwrite-strings -Wempty-body -Waggregate-return -Wmissing-field-initializers -Wredundant-decls -Winline -Wint-to-pointer-cast -Wlong-long -Wvla -Wno-attributes -g")
```

Thoughts?

kmod: I'm not clear on the details, but this change does seem strange since it's not Pyston-specific -- does CPython have this same issue when built in this kind of environment?

Daetalus: When builds CPython in this kind of environment(a.k.a SlapOS). The build script only needs to provide `LDFALGS` and `CPPFLAGS`. Something like this:

```
CPPFLAGS=-I${zlib:location}/include -I${readline:location}/include ...
LDFLAGS=-L${zlib:location}/lib -L${readline:location}/lib -L${libexpat:location}/lib ...
```

kmod: My vote is to keep it as close to the way that CPython does it as possible.  If it's something in their configure script, then it seems like we should put it in our CMake configuration.

I don't know if it matters here, but I think there's also a small confusion between `CPPFLAGS` and `CXXFLAGS` -- confusingly, `CPPFLAGS` doesn't have anything to do with C++, and is instead about the "C pre-processor".  I wouldn't be surprised if our cmake configuration just doesn't pick up that variable when it should.

PR #1365, opened 2016-09-13
Change section memory manager to fix performance problem in long-runn…

corona10:
Change section memory manager to fix performance problem in long-running SectionMemoryManagers discussed at https://reviews.llvm.org/D13156

### DIFF
--- a/src/codegen/memmgr.cpp
+++ b/src/codegen/memmgr.cpp
@@ -97,9 +97,8 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     uintptr_t RequiredSize = Alignment * ((Size + Alignment - 1) / Alignment + 1);
     uintptr_t Addr = 0;
 
-    // Note that all sections get allocated as read-write.  The permissions will
-    // be updated later based on memory group.
-    //
+    // Look in the list of free memory regions and use a block there if one
+    // is available.
     for (FreeMemBlock& FreeMB : MemGroup.FreeMem) {
         if (FreeMB.Free.size() >= RequiredSize) {
             Addr = (uintptr_t)FreeMB.Free.base();
@@ -147,6 +146,7 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     // Save this address as the basis for our next request
     MemGroup.Near = MB;
 
+    // Remember that we allocated this memory
     MemGroup.AllocatedMem.push_back(MB);
     Addr = (uintptr_t)MB.base();
     uintptr_t EndOfBlock = Addr + MB.size();
@@ -154,6 +154,7 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     // Align the address.
     Addr = (Addr + Alignment - 1) & ~(uintptr_t)(Alignment - 1);
 
+    // The part of the block we're giving out to the user is now pending
     MemGroup.PendingMem.push_back(sys::MemoryBlock((void*)Addr, Size));
 
     // The allocateMappedMemory may allocate much more memory than we need. In
@@ -165,15 +166,14 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
         FreeMB.PendingPrefixIndex = (unsigned)-1;
         MemGroup.FreeMem.push_back(FreeMB);
     }
+
+    // Return aligned address
     return (uint8_t*)Addr;
 }
 
 bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
     // FIXME: Should in-progress permissions be reverted if an error occurs?
     llvm_error_code ec;
-
-    // Don't allow free memory blocks to be used after setting protection flags.
-    CodeMem.FreeMem.clear();
 
     // Make code memory executable.
     // pyston: also make it writeable so we can patch it later
@@ -185,6 +185,7 @@ bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
         return true;
     }
 
+    // This code was removed in latest LLVM's repo.
     // Don't allow free memory blocks to be used after setting protection flags.
     RODataMem.FreeMem.clear();
 
@@ -225,6 +226,11 @@ static sys::MemoryBlock trimBlockToPageSize(sys::MemoryBlock M) {
     return Trimmed;
 }
 
+// Copy from LLVM 3.9 repo's /include/llvm/ADT/STLExtras.h
+template <typename R, class UnaryPredicate> auto remove_if(R&& Range, UnaryPredicate&& P) -> decltype(Range.begin()) {
+    return std::remove_if(Range.begin(), Range.end(), P);
+}
+
 llvm_error_code PystonMemoryManager::applyMemoryGroupPermissions(MemoryGroup& MemGroup, unsigned Permissions) {
 
     for (sys::MemoryBlock& MB : MemGroup.PendingMem) {
@@ -245,9 +251,8 @@ llvm_error_code PystonMemoryManager::applyMemoryGroupPermissions(MemoryGroup& Me
     }
 
     // Remove all blocks which are now empty
-    MemGroup.FreeMem.erase(std::remove_if(MemGroup.FreeMem.begin(), MemGroup.FreeMem.end(), [](FreeMemBlock& FreeMB) {
-        return FreeMB.Free.size() == 0;
-    }), MemGroup.FreeMem.end());
+    MemGroup.FreeMem.erase(remove_if(MemGroup.FreeMem, [](FreeMemBlock& FreeMB) { return FreeMB.Free.size() == 0; }),
+                           MemGroup.FreeMem.end());
 
 #if LLVMREV < 209952
     return llvm_error_code::success();
--- a/src/codegen/memmgr.cpp
+++ b/src/codegen/memmgr.cpp
@@ -182,12 +182,8 @@ bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
         if (ErrMsg) {
             *ErrMsg = ec.message();
         }
-        return true;
+        RELEASE_ASSERT(0, "finalizeMemory failed");
     }
-
-    // This code was removed in latest LLVM's repo.
-    // Don't allow free memory blocks to be used after setting protection flags.
-    RODataMem.FreeMem.clear();
 
     // Make read-only data memory read-only.
     ec = applyMemoryGroupPermissions(RODataMem, sys::Memory::MF_READ | sys::Memory::MF_EXEC);
@@ -195,7 +191,7 @@ bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
         if (ErrMsg) {
             *ErrMsg = ec.message();
         }
-        return true;
+        RELEASE_ASSERT(0, "finalizeMemory failed");
     }
 
     // Read-write data memory already has the correct permissions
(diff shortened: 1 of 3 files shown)


src/codegen/memmgr.cpp
```diff
@@ -230,12 +284,10 @@ uint64_t PystonMemoryManager::getSymbolAddress(const std::string& name) {
...
 PystonMemoryManager::~PystonMemoryManager() {
-    for (unsigned i = 0, e = CodeMem.AllocatedMem.size(); i != e; ++i)
-        sys::Memory::releaseMappedMemory(CodeMem.AllocatedMem[i]);
-    for (unsigned i = 0, e = RWDataMem.AllocatedMem.size(); i != e; ++i)
-        sys::Memory::releaseMappedMemory(RWDataMem.AllocatedMem[i]);
-    for (unsigned i = 0, e = RODataMem.AllocatedMem.size(); i != e; ++i)
-        sys::Memory::releaseMappedMemory(RODataMem.AllocatedMem[i]);
+    for (MemoryGroup* Group : { &CodeMem, &RWDataMem, &RODataMem }) {
+        for (sys::MemoryBlock& Block : Group->AllocatedMem)
+            sys::Memory::releaseMappedMemory(Block);
```
kmod: It looks like they free PendingMem here as well

kmod:
Just curious -- there seem to be more changes in this PR than in the patch you linked to, are those additional changes from backporting the latest MemoryManager code?

kmod:
Oh and just FYI, I don't think github sends out any notifications when you change the title of a PR, so if something becomes ready to review you might sometimes need to ping us to take a look.

corona10:
@kmod Yes I referenced latest code of theit repo. They don`t free pending memory.

src/codegen/memmgr.cpp
```diff
@@ -185,15 +207,47 @@ bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
...
-    for (int i = 0, e = MemGroup.AllocatedMem.size(); i != e; ++i) {
+    for (sys::MemoryBlock& MB : MemGroup.PendingMem) {
         llvm_error_code ec;
-        ec = sys::Memory::protectMappedMemory(MemGroup.AllocatedMem[i], Permissions);
+        ec = sys::Memory::protectMappedMemory(MB, Permissions);
         if (ec) {
             return ec;
         }
     }
+    MemGroup.PendingMem.clear();
```
corona10: @kmod 
According to this [diff file](https://reviews.llvm.org/file/data/vxoazq7dntpccoreh7ro/PHID-FILE-ewe7kqapgt45lwhupue5/D15202.diff)
They might think that they do not have to free pending memory when destructor is called.

kmod: Oh ok, I was just looking at that diff you mentioned in the commit message.

kmod:
A change like this is very hard to review because it's essentially impossible to tell if it does what it is supposed to (rebase our patches onto the newer SectionMemoryManager.cpp).  I think in these cases it's worth coming up with a systematic way of doing the change, both because it reduces the chances for minor issues, but also because it means that we can review the process and not the (unreviewable) output.

For example, here's what I did -- I took diff of SectionMemoryManager.cpp, using Pyston's pinned LLVM rev as the base (after clang-formatting it), and the latest version as the target of the diff (also clang-formatted to our style).  Then, I diffed that against the diff in this PR.  The output is pretty hard to read but turned up a couple things:
- I think you missed the last two commits that were made to llvm's SectionMemoryManager: "Use a range variant" and "Don't just drop the RO free list".
- Some of the comments / whitespace didn't get transferred :/
- It looks like the following block got deleted in their version:
```
    // Don't allow free memory blocks to be used after setting protection flags.
    CodeMem.FreeMem.clear();
```

It might not be worth it for this PR at this point, but in the future I'd definitely recommend coming up with a systematic approach because doing this kind of 3-way merge by hand is tedious and not very fun :(

kmod:
Oh ok well I guess we can't use that latest commit ("Use a range variant") because I think it relies on other changes in LLVM that we haven't pulled in yet.

I used a hacky "munge the LLVM diff until it applies to our file" approach to generate a rough diff that can be applied to this PR::
```diff
diff --git a/src/codegen/memmgr.cpp b/src/codegen/memmgr.cpp
index 71d8486..9f25c86 100644
--- a/src/codegen/memmgr.cpp
+++ b/src/codegen/memmgr.cpp
@@ -15,11 +15,11 @@
-    llvm_error_code applyMemoryGroupPermissions(MemoryGroup& MemGroup, unsigned Permissions);
+    std::error_code applyMemoryGroupPermissions(MemoryGroup& MemGroup, unsigned Permissions);
 
     uint64_t getSymbolAddress(const std::string& Name) override;
 
@@ -97,9 +92,8 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     uintptr_t RequiredSize = Alignment * ((Size + Alignment - 1) / Alignment + 1);
     uintptr_t Addr = 0;
 
-    // Note that all sections get allocated as read-write.  The permissions will
-    // be updated later based on memory group.
-    //
+    // Look in the list of free memory regions and use a block there if one
+    // is available.
     for (FreeMemBlock& FreeMB : MemGroup.FreeMem) {
         if (FreeMB.Free.size() >= RequiredSize) {
             Addr = (uintptr_t)FreeMB.Free.base();
@@ -124,6 +118,7 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
             return (uint8_t*)Addr;
         }
     }
+
     // No pre-allocated free block was large enough. Allocate a new memory region.
     // Note that all sections get allocated as read-write.  The permissions will
     // be updated later based on memory group.
@@ -133,12 +128,12 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     //
     // FIXME: Initialize the Near member for each memory group to avoid
     // interleaving.
-    llvm_error_code ec;
+    std::error_code ec;
     sys::MemoryBlock MB = sys::Memory::allocateMappedMemory(RequiredSize, &MemGroup.Near,
                                                             sys::Memory::MF_READ | sys::Memory::MF_WRITE, ec);
     if (ec) {
         // FIXME: Add error propogation to the interface.
-        return nullptr;
+        return NULL;
     }
 
     std::string stat_name = "mem_section_" + std::string(SectionName);
@@ -147,6 +142,7 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     // Save this address as the basis for our next request
     MemGroup.Near = MB;
 
+    // Remember that we allocated this memory
     MemGroup.AllocatedMem.push_back(MB);
     Addr = (uintptr_t)MB.base();
     uintptr_t EndOfBlock = Addr + MB.size();
@@ -154,6 +150,7 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
     // Align the address.
     Addr = (Addr + Alignment - 1) & ~(uintptr_t)(Alignment - 1);
 
+    // The part of the block we're giving out to the user is now pending
     MemGroup.PendingMem.push_back(sys::MemoryBlock((void*)Addr, Size));
 
     // The allocateMappedMemory may allocate much more memory than we need. In
@@ -165,15 +162,14 @@ uint8_t* PystonMemoryManager::allocateSection(MemoryGroup& MemGroup, uintptr_t S
         FreeMB.PendingPrefixIndex = (unsigned)-1;
         MemGroup.FreeMem.push_back(FreeMB);
     }
+
+    // Return aligned address
     return (uint8_t*)Addr;
 }
 
 bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
     // FIXME: Should in-progress permissions be reverted if an error occurs?
-    llvm_error_code ec;
-
-    // Don't allow free memory blocks to be used after setting protection flags.
-    CodeMem.FreeMem.clear();
+    std::error_code ec;
 
     // Make code memory executable.
     // pyston: also make it writeable so we can patch it later
@@ -185,9 +181,6 @@ bool PystonMemoryManager::finalizeMemory(std::string* ErrMsg) {
         return true;
     }
 
-    // Don't allow free memory blocks to be used after setting protection flags.
-    RODataMem.FreeMem.clear();
-
     // Make read-only data memory read-only.
     ec = applyMemoryGroupPermissions(RODataMem, sys::Memory::MF_READ | sys::Memory::MF_EXEC);
     if (ec) {
@@ -225,15 +218,12 @@ static sys::MemoryBlock trimBlockToPageSize(sys::MemoryBlock M) {
     return Trimmed;
 }
 
-llvm_error_code PystonMemoryManager::applyMemoryGroupPermissions(MemoryGroup& MemGroup, unsigned Permissions) {
 
-    for (sys::MemoryBlock& MB : MemGroup.PendingMem) {
-        llvm_error_code ec;
-        ec = sys::Memory::protectMappedMemory(MB, Permissions);
-        if (ec) {
-            return ec;
-        }
-    }
+std::error_code PystonMemoryManager::applyMemoryGroupPermissions(MemoryGroup& MemGroup, unsigned Permissions) {
+    for (sys::MemoryBlock& MB : MemGroup.PendingMem)
+        if (std::error_code EC = sys::Memory::protectMappedMemory(MB, Permissions))
+            return EC;
+
     MemGroup.PendingMem.clear();
 
     // Now go through free blocks and trim any of them that don't span the entire
@@ -245,15 +235,10 @@ llvm_error_code PystonMemoryManager::applyMemoryGroupPermissions(MemoryGroup& Me
     }
 
     // Remove all blocks which are now empty
-    MemGroup.FreeMem.erase(std::remove_if(MemGroup.FreeMem.begin(), MemGroup.FreeMem.end(), [](FreeMemBlock& FreeMB) {
-        return FreeMB.Free.size() == 0;
-    }), MemGroup.FreeMem.end());
-
-#if LLVMREV < 209952
-    return llvm_error_code::success();
-#else
-    return llvm_error_code();
-#endif
+    MemGroup.FreeMem.erase(llvm::remove_if(MemGroup.FreeMem, [](FreeMemBlock& FreeMB) { return FreeMB.Free.size() == 0; }),
+                           MemGroup.FreeMem.end());
+
+    return std::error_code();
 }
 
 void PystonMemoryManager::invalidateInstructionCache() {
@@ -289,7 +274,6 @@ PystonMemoryManager::~PystonMemoryManager() {
             sys::Memory::releaseMappedMemory(Block);
     }
 }
-
 std::unique_ptr<llvm::RTDyldMemoryManager> createMemoryManager() {
     return std::unique_ptr<llvm::RTDyldMemoryManager>(new PystonMemoryManager());
 }
```

corona10:
@kmod Yes that codes are depend on LLVM version :-)

corona10:
@kmod I will fix them what you commented

corona10:
@kmod 
I fix them except some codes.

* Use std::remove_if rather than llvm::remove_if 
* RODataMem.FreeMem.clear(); and CodeMem.FreeMem.clear(); should be exist. when I remove these codes. it makes error when I build pyston. I think it may be relate with memory permission.

```
ninja -C /home/corona10/pyston/build/Debug pyston copy_stdlib sharedmods ext_pyston 
ninja: Entering directory `/home/corona10/pyston/build/Debug'
[4/4] Generating basic_test.pyston.so
/home/corona10/pyston/build/Debug/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'zip_safe'
  warnings.warn(msg)
running build
running build_ext
[4/4] Generating ../lib/python2.7/lib-dynload/...../lib/python2.7/lib-dynload/termios.pyston.so
FAILED: cd /home/corona10/pyston/from_cpython && /home/corona10/pyston/build/Debug/pyston setup.py build --build-lib /home/corona10/pyston/build/Debug/lib/python2.7/lib-dynload
/home/corona10/pyston/build/Debug/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'zip_safe'
  warnings.warn(msg)
running build
running build_ext
Segmentation fault (core dumped)
ninja: build stopped: subcommand failed.
Makefile:657: recipe for target 'pyston_dbg' failed
make: *** [pyston_dbg] Error 1

```

kmod:
Because of the trickiness of this, I think we need a better understanding than just that backporting a certain change avoids a segfault.  I looked into it, and what's happening is that finalizeMemory is failing, but apparently nothing is checking the return code of it.  The problem is that the changes you backported depend on some other changes which we don't have.

I think the fix is pretty straightforward though:
- Can you backport llvm revision 255760, at least the changes to lib/Support/Unix/Memory.inc?  You'll have to do this by dropping a patch file into `llvm_patches/`.
- Can you change the places where `finalizeMemory` does a `return true` to instead do `RELEASE_ASSERT(0, "finalizeMemory failed")`

corona10:
@kmod You're right. It works well when we apply a patch into our llvm's lib/Support/Unix/Memory.inc

I fixed it as you commented.

kmod:
Awesome!  Thanks for sticking through this even though it ended up being pretty gnarly.

PR #1369, opened 2016-09-19
Fix docs about stats option

sh92:
I think the option about stats is outdated. Thx.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Pyston-specific flags:
 <dt>-v</dt>
   <dd>Increase verbosity by 1</dd>
 
-<dt>-s</dt>
+<dt>-T</dt>
   <dd>Print out the internal stats at exit.</dd>
 
 <dt>-n</dt>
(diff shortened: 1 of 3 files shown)


undingen:
good catch  :+1:, thanks for the change!

Could you please remove the libunwind commit it's causing the travis-ci error :unamused: .
Unfortunately our build system currently applies some patches to libunwind (libunwind_patches/*) and git commits them during the build which causes cause libunwind to change - which means we normally never want to actually commit this changes because the revision will only exist on you host locally which causes the travis-ci error.

I have seen that there is another '-s' in `README.md`: `-s Print out the internal stats at exit.` could you change change that one too.

And would you mind signing the the CLA please: https://opensource.dropbox.com/cla/
I don't think it's absolutely necessary for this change but would be nice to have for the future :-)
Sorry for all my requests :stuck_out_tongue:

sh92:
OK, I fix it. This is my first project. I want to learn a lot in this project. Thank you for review.

undingen:
thanks for your patch :+1:
I notified colleagues of mine to take a look into why the CLA site is down but I merged this in because it is a very small change. But if you plan to submit future PRs (please do :smile: ) you will have to sign it please.

PR #1370, opened 2016-09-20
Add type.__instancecheck__ and __subclasscheck__

kmod:
We supported classes that overrode them, but didn't provide a
default implementation in case people called it by hand.

The tricky part is making sure that these new additions don't
mess with our "does this class override __instancecheck__" optimizations.

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -1400,6 +1400,28 @@ static int type_sub_set_dict(BORROWED(Box*) obj, BORROWED(Box*) val, void* conte
     return 0;
 }
 
+static PyObject* type___instancecheck__(PyObject* type, PyObject* inst) noexcept {
+    switch (_PyObject_RealIsInstance(inst, type)) {
+        case -1:
+            return NULL;
+        case 0:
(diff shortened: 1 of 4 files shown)


src/runtime/types.cpp
```diff
@@ -4399,6 +4421,12 @@ void setupRuntime() {
     type_cls->giveAttrDescriptor("__abstractmethods__", (getter)type_abstractmethods, (setter)type_set_abstractmethods);
 
     type_cls->giveAttr("__call__", new BoxedFunction(typeCallObj));
+    type_cls->giveAttr(
+        "__subclasscheck__",
+        new BoxedFunction(BoxedCode::create((void*)type___subclasscheck__, BOXED_BOOL, 2, "type.__subclasscheck__")));
```
undingen: I think this should be, `BoxedCode::create((void*)type___subclasscheck__, BOXED_BOOL, 2, "type.__subclasscheck__", ParamNames::empty(), CAPI)`

kmod: Good catch!

PR #1374, opened 2016-09-26
[WIP]we can use Cython 0.25 without modification now

Daetalus:


test/extra/lxml_test.py
```diff
@@ -13,11 +13,7 @@ def install_and_test_lxml():
     os.makedirs(SRC_DIR)
 
     url = "https://github.com/cython/cython"
-    subprocess.check_call(["git", "clone", "--depth", "1", "--branch", "0.24", url], cwd=SRC_DIR)
-
-    PATCH_FILE = os.path.abspath(os.path.join(os.path.dirname(__file__), "../integration/Cython-0.24.patch"))
-    subprocess.check_call(["patch", "-p1", "--input=" + PATCH_FILE], cwd=CYTHON_DIR)
-    print ">>> Applied Cython patch"
+    subprocess.check_call(["git", "clone", "--depth", "1", "--branch", "0.25a0", url], cwd=SRC_DIR)
```
kmod: FYI it looks like they uploaded this to pypi as well so we should be able to do `pip install cython==0.25a0`!

kmod:
Bummer -- it looks like lxml 3.0.1 is incompatible with cython 0.25.  Newer versions of lxml drop some old hacks of theirs, which makes them compatible with the new cython, but Dropbox still uses the old lxml in some places and I'm not sure if we have a timetable on upgrading it.

kmod:
I think we should still get this PR in -- how about just undoing the changes to lxml_test.py, but upgrading the others to cython 0.25?

Daetalus:
Np, sorry for the delay. I will change it soon.

PR #1377, opened 2016-09-27
BST: convert all nodes to directly operate at vregs instead of names

undingen:
**basic design:**
This PR changes our BST nodes to directly operate on vregs instead of pointers to other nodes and names (except a few exceptions: `BST_Invoke`, `BST_MakeFunction` and `BST_MakeClass` which still needs to get converted).
Most nodes got a destination vreg and one or more source vregs. Currently all of them are 32bit long but I plan to store them more compact very soon. Some nodes support a variable size of operands (e.g. the tuple node) but the size can't change after creating the node. I removed several unneeded opcodes and split a lot of nodes into separate opcodes (it may make sense to split them even further in the future). 
Generally all instructions except `CopyVReg` kill the source operand vregs except if the source is a ref to a constant. If one needs the preserve the source vreg on needs to create a new temporary using the `CopyVReg` opcode.

There is a special vreg number: `VREG_UNDEFINED = std::numeric_limits<int>::min()`.
- when it's set as an operand vreg: it means that this is a not-set optional argument. (e.g. for a slice which only has `lower` set, `upper` would be `VREG_UNDEFINED`)
- if it's the destination it's means the result value should get immediately killed (e.g. `invoke 15 16: %undef = %11(%14)` this is a call whose result gets ignored)

all other negative vreg numbers are indices into a constant table (after adding 1 and making them positive).
(e.g. `(4, 2, 'lala')` generates:  `%undef = (%-1|4|, %-2|2|, %-3|'lala'|)` this creates a tuple whose elements are the constant idx -1, -2 and -3. In order to make it easier for a human to understand we print the actual value of the constant between | characters)
- constants can be all str and numeric types and 'None'.
- every constant will only get stored once in the table

this reduces the total memory usage by about 20% currently but I'm very sure with the future changes it will be significantly lower.

**near future:**
- change the jump and branch instruction to reference `CFGBlocks` by index.
- store all `InternedString` inside a table and use indices into the the table to access them.
- remove the `BoxedCode*` member
- devirtualize the classes
= with this changes the bytecode can get freely copied around (only need to update the CFGBlock table) which allows us to store them directly next to each other.

- I plan to use one bit of the the opcode to mark the instruction as only requiring 8bit vreg operands (which should handle the majority of cases with 128 temps and 127 constants + 1undef vreg value) 
- another bit will get used to specify if this instruction is inside an `invoke`. if this bit is set there are two 1 or 4 bytes long block indices directly behind the instruction.
 
- serialize the bytecode to disk. (maybe serialize the constant table using pickle)

**thing which need to get improved**
- currently the constant table get's attached to the `BoxedModule` maybe there is a better location, I also needed to pass the `BoxedModule` into some functions e.g. BST printing because otherwise we could not pretty-print the constants
- `BST_Name` is not an opcode it's just used to initialize the arguments when a function get's called and stores where and how the arguments need to get stored.
- more consistent opcode names and rename `TmpValue` to something better
- we currently don't print the `InternedString` name - we only print the vreg number

**additional changed made which are hidden in the large diff** :-1:
- removed unused code initializing the items of `BST_Dict` (we use/used separate assignments to add the items)
- lower `ExtSlice` inside the CFG phase to a tuple of slices
- separated opcode for load subscript when it needs to be a slice and when it's only lower and upper (=`__getslice__`) before this got handled in the interpreter/jit
- generate a constant `None` load inside the CFG when `None` gets loaded by name

### DIFF
--- a/src/codegen/irgen.cpp
+++ b/src/codegen/irgen.cpp
@@ -337,6 +337,38 @@ llvm::Value* handlePotentiallyUndefined(ConcreteCompilerVariable* is_defined_var
     return phi;
 }
 
+// This is used to filter out any names set by an invoke statement at the end of the previous block, if we're in the
+// unwind path.
+class SymTableDstVRegDeleter : NoopBSTVisitor {
+private:
+    SymbolTable* sym_table;
+    bool created_new_sym_table;
+
+    SymTableDstVRegDeleter(SymbolTable* sym_table)
+        : NoopBSTVisitor(true /* skip child CFG nodes */), sym_table(sym_table), created_new_sym_table(false) {}
+
+protected:
+    bool visit_vreg(int* vreg, bool is_dst = false) override {
+        if (!is_dst || *vreg == VREG_UNDEFINED || !(*sym_table)[*vreg])
+            return false;
+
+        if (!created_new_sym_table) {
+            sym_table = new SymbolTable(*sym_table);
+            created_new_sym_table = true;
+        }
+        (*sym_table)[*vreg] = NULL;
+        return false;
+    }
+
+public:
+    static std::pair<SymbolTable*, bool /* created_new_sym_table */> removeDestVRegsFromSymTable(SymbolTable* sym_table,
+                                                                                                 BST_Invoke* stmt) {
+        SymTableDstVRegDeleter visitor(sym_table);
+        stmt->accept(&visitor);
+        return std::make_pair(visitor.sym_table, visitor.created_new_sym_table);
+    }
+};
+
 static void emitBBs(IRGenState* irstate, TypeAnalysis* types, const OSREntryDescriptor* entry_descriptor,
                     const BlockSet& blocks) {
     SourceInfo* source = irstate->getSourceInfo();
@@ -551,7 +583,7 @@ static void emitBBs(IRGenState* irstate, TypeAnalysis* types, const OSREntryDesc
         if (block == cfg->getStartingBlock()) {
             assert(entry_descriptor == NULL);
 
-            if (ENABLE_REOPT && effort < EffortLevel::MAXIMAL && source->ast_type != BST_TYPE::Module) {
+            if (ENABLE_REOPT && effort < EffortLevel::MAXIMAL && source->ast_type != AST_TYPE::Module) {
                 llvm::BasicBlock* preentry_bb = llvm::BasicBlock::Create(
                     g.context, "pre_entry", irstate->getLLVMFunction(), llvm_entry_blocks[cfg->getStartingBlock()]);
                 llvm::BasicBlock* reopt_bb = llvm::BasicBlock::Create(g.context, "reopt", irstate->getLLVMFunction());
@@ -695,16 +727,6 @@ static void emitBBs(IRGenState* irstate, TypeAnalysis* types, const OSREntryDesc
                 // analysis frameworks can't (yet) support the idea of a block flowing differently to its different
                 // successors.
                 //
-                // There are four kinds of BST statements which can set a name:
-                // - Assign
-                // - ClassDef
-                // - FunctionDef
-                // - Import, ImportFrom
-                //
-                // However, all of these get translated away into Assigns, so we only need to worry about those. Also,
-                // as an invariant, all assigns that can fail assign to a temporary rather than a python name. This
-                // ensures that we interoperate properly with definedness analysis.
-                //
                 // We only need to do this in the case that we have exactly one predecessor, because:
                 // - a block ending in an invoke will have multiple successors
                 // - critical edges (block with multiple successors -> block with multiple predecessors)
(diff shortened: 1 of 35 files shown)


undingen:
ok this should be ready for an initial highlevel review + suggestions for renaming etc...

src/core/bst.h
```diff
@@ -188,361 +133,540 @@ class BST {
...
 
-    int cxx_exception_count = 0;
-
     BST_stmt(BST_TYPE::BST_TYPE type) : BST(type) {}
+    BST_stmt(BST_TYPE::BST_TYPE type, uint32_t lineno) : BST(type, lineno) {}
 };
 
-class BST_slice : public BST {
+// base class of all nodes which have a single destination vreg
+class BST_dst : public BST_stmt {
```
kmod: Only because you asked for thoughts on naming :P  When I first saw `BST_dst` I thought it was the destination that a bst node assigns to, not a bst node with a destination.  I have a very small preference for something like `BST_stmt_withdest`, and maybe calling `isBST_dst` something like `has_dest`.

src/core/bst.h
```diff
@@ -163,8 +108,8 @@ static const char* stringify(int n) {
 class BSTVisitor;
 class ExprVisitor;
 class StmtVisitor;
-class SliceVisitor;
-class BST_keyword;
+
+static constexpr int VREG_UNDEFINED = std::numeric_limits<int>::min();
```
kmod: Small comment here would be nice, especially if this is pulling double duty as both a source and a destination vreg.

src/core/bst.h
```diff
@@ -188,361 +133,540 @@ class BST {
...
+    int vreg_dst[1];
 
     virtual void accept(BSTVisitor* v);
     virtual void accept_stmt(StmtVisitor* v);
 
-    BST_Assert() : BST_stmt(BST_TYPE::Assert) {}
+    static BST_UnpackIntoArray* create(int num_elts) {
+        BST_UnpackIntoArray* o
+            = (BST_UnpackIntoArray*)new char[offsetof(BST_UnpackIntoArray, vreg_dst) + num_elts * sizeof(int)];
+        new (o) BST_UnpackIntoArray(num_elts);
```
kmod: Just a small personal preference, but I usually like to override `operator new` rather than using placement-new.  I like it because it makes it clear you are just changing how the allocation works, while still letting the compiler deal with the casting + constructor.  It also blocks the user from doing something like `new BST_UnpackIntoArray(*BST_UnpackIntoArray::create(10))`, which I think would currently compile but crash.

src/core/bst.h
```diff
@@ -188,361 +133,540 @@ class BST {
...
+        }
+    }
 };
 
-class BST_Assign : public BST_stmt {
+class BST_CopyVReg : public BST_dst {
 public:
-    BST_expr* target;
-    BST_expr* value;
+    int vreg_src = VREG_UNDEFINED; // this vreg will not get killed!
```
kmod: I think it'd be great to have a comment saying why we need a CopyVReg bytecode.  And maybe have a comment towards the top of the file that goes over the properties of the bytecode, such as that source vregs are always killed after use, how constants are represented, etc.

src/core/bst.h
```diff
@@ -188,361 +133,540 @@ class BST {
...
-class BST_expr : public BST {
-public:
-    virtual void* accept_expr(ExprVisitor* v) = 0;
 
-    BST_expr(BST_TYPE::BST_TYPE type) : BST(type) {}
-    BST_expr(BST_TYPE::BST_TYPE type, uint32_t lineno) : BST(type, lineno) {}
+    virtual bool isBST_dst() const { return false; }
 };
 
 class BST_stmt : public BST {
```
kmod: It looks like this is the only class that inherits from BST now --  can we just fold these fields into BST itself?

src/core/bst.h
```diff
@@ -787,126 +1030,130 @@ template <typename T> T* bst_cast(BST* node) {
...
+    virtual bool visit_getiter(BST_GetIter* node) override { return false; }
+    virtual bool visit_hasnext(BST_HasNext* node) override { return false; }
+    virtual bool visit_importfrom(BST_ImportFrom* node) override { return false; }
+    virtual bool visit_importname(BST_ImportName* node) override { return false; }
+    virtual bool visit_importstar(BST_ImportStar* node) override { return false; }
     virtual bool visit_invoke(BST_Invoke* node) { return false; }
-    virtual bool visit_keyword(BST_keyword* node) { return false; }
-    virtual bool visit_langprimitive(BST_LangPrimitive* node) { return false; }
+    virtual bool visit_jump(BST_Jump* node) { return false; }
+    virtual bool visit_landingpad(BST_Landingpad* node) override { return false; }
```
kmod: Can you remove the `override` markers here or add them to all of the functions?  newer gccs complain.  same with FlattenVisitor and BasicBlockTypePropagator.

Maybe we should turn off `-Winconsistent-missing-override` until we can get that checked in CI :P

src/core/bst.h
```diff
@@ -188,361 +133,540 @@ class BST {
...
+    BST_DeleteSubSlice() : BST_stmt(BST_TYPE::DeleteSubSlice) {}
 
-    // Only valid for lookup_type == DEREF:
-    DerefInfo deref_info = DerefInfo({ INT_MAX, INT_MAX });
-    // Only valid for lookup_type == CLOSURE:
-    int closure_offset = -1;
+    static const BST_TYPE::BST_TYPE TYPE = BST_TYPE::DeleteSubSlice;
+};
 
+class BST_Ellipsis : public BST_dst {
```
kmod: I think this can be treated as a constant rather than another node (looks like this is how CPython does it).

src/runtime/types.h
```diff
@@ -1270,6 +1271,8 @@ class BoxedModule : public Box {
 public:
     DEFAULT_CLASS(module_cls);
 
+    std::vector<Box*> constants;
```
kmod: I think this is worth a comment here -- I wasn't expecting to find that constants are stored twice on the module, once in the `str/int/long/etc_constants` and once in `constants`.  I think we should eventually move to having the constants be stored on the code objects, but for now I think the current way is fine but should have a comment / todo.

src/core/bst.h
```diff
@@ -165,7 +165,55 @@ class BST_stmt_with_dest : public BST_stmt {
     bool has_dest_vreg() const override { return true; }
 };
 
+#define BSTVARVREGS(opcode, base_class, num_elts, vreg_dst)                                                            \
```
undingen: @kmod Is this what you had in mind? 
I also tried to reduce the code duplication using macros... But in general I don't worry too much about the Nodes currently because I'm currently working on a python script which auto generates them in order to support the 1 and 4 byte vregs...

undingen:
I cleaned it up, it's ready now for review :+1:.
But I left my changes in a separate commit to it's easier to see what I changed.

src/codegen/irgen.cpp
```diff
@@ -337,6 +337,36 @@ llvm::Value* handlePotentiallyUndefined(ConcreteCompilerVariable* is_defined_var
     return phi;
 }
 
+class SymTableDstVRegDeleter : NoopBSTVisitor {
```
kmod: Could we have a comment for what this is used for?

kmod:
:+1: :+1:

PR #1378, opened 2016-09-28
Add test for typing.py and fix the issues it uncovers

kmod:
typing.py likes to create lots of class objects, and it exposed an issue that we didn't properly invalidate ICs for freed class objects.


undingen:
I'm sorry for introducing this hard to find bug :-(

kmod:
I don't think it was you ;)

PR #1380, opened 2016-10-05
update to cpython 2.7.8

aisk:
An attempt to update cpython code base to 2.7.8 (hg tag v2.7.8), just copied the changed codes to pyston's codes. Most changes is under `from_cpython/Lib/test/` folder.

I use [meld](http://meldmerge.org/)'s tree way comparison fuction to find the changes between cpython2.7.7 to cpython2.7.8 to pyston, I think it's usable for anyone who want to review this PR or do some other works like this.

Actually I want try to update cpython code to current version 2.7.12, but the change set is to large to merge, so I think we should update it one by one.

---
#### from_cpython/Lib/test/test_support.py

test_support.py:is_resource_enabled in 2.7.7

``` python
def is_resource_enabled(resource):
    """Test whether a resource is enabled.  Known resources are set by
    regrtest.py."""
    return use_resources is not None and resource in use_resources

```

test_support.py:is_resource_enabled in 2.7.8

``` python
def is_resource_enabled(resource):
    """Test whether a resource is enabled.

    Known resources are set by regrtest.py.  If not running under regrtest.py,
    all resources are assumed enabled unless use_resources has been set.
    """
    return use_resources is None or resource in use_resources
```

So this function changes to always returns True from always returns False. In CPython's tests codes, this function is to reduce the tests run time, but break down Pyston's tests. `from_cpython/Lib/test/test_ssl.py` and `from_cpython/Lib/test/test_tokenize.py` have two place have called this function and I have modified them just to pass it.
#### from_cpython/Lib/test/test_collections.py

Another Python change is I disabled the whole `test_Set_interoperability_with_real_sets` test function in `from_cpython/test/test_collections.py` which was introduced in cpython 2.7.8. This test case added a lot of set (fronzen set, user set) compare function tests, and I find the current Pyston set compare functions (`__eq__`, `__gt__` etc) was wrong when deal with user set (inherif from collections.Set). And I think we should copy the cpython's `set_richcompare` function to `set_cls` and `fronzenset`'s `richcompare` slot. I tried to do this, and it works well with this test case, but sadly a lot of other exsited test cases was broken, I think maybe they have already depenced on the wrong sets compare implementation in some ways.

So this two changes won't break current existed test cases, so can we review it first, and then fix them in another PR?


aisk:
Seems like `test_tokenizer.py` is slower than before in debug mode, so the test is always timeouted.

kmod:
Hi @aisk, I like this PR a lot, but reviewing it is going to be tricky.  Can you post some details about the process you used to put this PR together?  For example, if there is some way to automatically merge most changes, then I could do that and then only review the parts that did not merge automatically.

aisk:
Hi @kmod, really sorry for I can't provide any tool to decide which fill can be automatically merged.

But I wrote a quick and dirty script to help review the changes between `cpython_v2.7.7 & cpython_v2.7.8 & pyston` using vimdiff in three panels in same time.

The script is on this [gist](https://gist.github.com/aisk/eddfbf96d013c27880440be1db31d0dc), and you need vimdiff / hg / python2 installed on your machine, a clone of cpython hg repo and a clone of pyston repo with this PR's branch.

Then run this command:

```sh
python2 merge_helper.py path-to-your-cpython-hg-repo/ path-to-pyston/cpython 0
```

And sorry I wrote this scrippt in a hurry, please don't missing the last '/' to the path params. Then this script will find the diff files between cpython v2.7.7 tag to v2.7.8 tag, and open `vimdiff` with this two files and pyston file.

The last param is `skip`, if you have any break in review (like my script crashed `: (` )
, you can start re-review with this last break point. (Tip: you can press ctrl-z to detach vimdiff to saw the script's output, which have the skip number).

Though we still have about 200 files to review, but most of them is straightforward (we did not change them in pyston except some `# Pyston changes:`).

kmod:
Awesome, thanks for the script!  I updated the script to automatically try to apply the changes, and surprisingly it didn't run into any conflicts.  I [put up the results](https://github.com/kmod/pyston/tree/1380_check), since there are a number of differences (mostly minor).  Could you take a look?  You'll need to do

```
git rebase origin/master # I did my changes against a different version from yours
git fetch https://github.com/kmod/pyston
git checkout -b 1380_kmod FETCH_HEAD
git d 1380_kmod
```

We should definitely check in the script along with the PR!

aisk:
Hi @kmod sorry I'm a little busy last week, I will have a look at this as soon as I can (probably this night).

aisk:
Hi I compared the codes between this PR and your branch, now they have only small but nessesary changes, can you review it again?

kmod:
`git diff 735ace2034502f3d1f8a9edb48c58a5958329384 a6a776fa74e3ba04dac9bd7236a4e36420407552` is still showing a number of differences.  You're right that many are just comments, but there are others that I don't understand.  For instance why were the ssl tests disabled, or why were some files removed / not added?

aisk:
Hi @kmod, I have updated this PR, now it shows only a few diffes from your branch.

In `from_cpython/Lib/test/test_ssl.py`, I found the test code using a cert file `from_cpython/Lib/test/https_svn_python_org_root.pem` to connect to `https://svn.python.org`, but seems like the cert file is invalid now, and the latest cpython 2.7 branch (2.7.12) have removed using this URL to test. So I temporarily changed the codes to using system's root ca file to run the tests.

Another change is some tests was disabled in `from_cpython/Lib/test/test_set.py`, because our custom set compare implementation have some probplems. I think we could just using the cpython's implementation.

kmod:
Yeah the set-comparison behavior changed between 2.7.7 and 2.7.8, so if we only import their new tests we will have issues with them.  I think either changing the tests (like you did, though I think a note about the version discrepency would be more relevant than saying our code is invalid) or changing our code is fine.  Separately, I agree I think it probably makes sense to switch to CPython's set implementation at some point so that we don't have to worry about these things.

Anyway, I'm seeing this diffstat between our branches: `12 files changed, 383 insertions(+), 461 deletions(-)`-- using `git diff` like I mentioned should be a help in tracking them down.  At this point it is just a file being misplaced, and the cpython_merge_helper script not being checked in -- please copy in that script from my branch so that we can have it for next time.

aisk:
@kmod Oh sorry I did not noticed the misplace. I updated the PR, and now this is the result of running `git diff kmod/1380_check --stat`:

```
 from_cpython/CMakeLists.txt               |  1 +
 from_cpython/Lib/distutils/ccompiler.py   |  1 +
 from_cpython/Lib/test/string_tests.py     |  3 +++
 from_cpython/Lib/test/test_collections.py | 25 +++++++++++++++++--------
 from_cpython/Lib/test/test_compile.py     |  2 +-
 from_cpython/Lib/test/test_ssl.py         |  9 ++++++---
 from_cpython/Lib/threading.py             |  2 ++
 src/runtime/tuple.cpp                     | 14 +++++++++++---
 src/runtime/types.cpp                     |  2 ++
 9 files changed, 44 insertions(+), 15 deletions(-)
(END)
```

Sorry for let you review this so many times...

kmod:
Awesome, this looks great!  Sorry if this seemed like some unnecessary work, but doing it this way both made it feasible to review, and also means that doing this again to other micro revisions should be quite easy now :)

PR #1384, opened 2016-10-08
BST: remove some of the remaining pointers, earlier boxed code destruction

undingen:
This is a prerequisite for https://github.com/dropbox/pyston/pull/1381 and removes the `InternedString` fields and pointers to `BoxedCode`. The code objects are now all currently refcounted which means we will free them when the are unused.


kmod:
Not critical for this PR, but re the "remove pointer to FunctionDef and ClassDef" commit, I think we can clean that up -- we should be able to use the same constants strategy for the BoxedCode objects that we do for the other objects.  And I think at this point we shouldn't need the parent BST node to keep a reference to the child BST node?  I haven't verified that but it seems like all of the relevant information should already be included in the code object.

I'd also like to see us kill InternedString rather than keep cludging it.  I think your `unsafe()` function already makes use of the fact that InternedString == BoxedString, and I think the better way to use that fact is to just get rid of InternedString :P

Anyway, these are just cleanups, which we can do at any time.

PR #1391, opened 2016-11-03
transform BST to gapless bytecode inplace

undingen:
this is a modified version of https://github.com/dropbox/pyston/pull/1381 which creates the bytecode in-place and does not evolve duplicating some CFG structures.
- all instructions of the CFG are directly emitted a single memory region.
- all instructions are packed and follow each other directly
- the CFGBlock just point into the bytecode block where the first instruction of the block is located
- invokes are only a bit in the opcode field which when set means that the pointers to the normal and exc `CFGBlocks*` directly follow the normal instruction

### DIFF
--- a/src/runtime/types.cpp
+++ b/src/runtime/types.cpp
@@ -30,6 +30,7 @@
 #include "codegen/ast_interpreter.h"
 #include "codegen/entry.h"
 #include "codegen/unwinding.h"
+#include "core/bst.h"
 #include "core/options.h"
 #include "core/stats.h"
 #include "core/types.h"
@@ -4054,6 +4055,8 @@ void CodeConstants::dealloc() const {
     owned_refs.clear();
     for (auto&& e : funcs_and_classes) {
         Py_DECREF(e.second);
+        assert(e.first->type() == BST_TYPE::FunctionDef || e.first->type() == BST_TYPE::ClassDef);
+        delete[] e.first;
     }
     funcs_and_classes.clear();
 }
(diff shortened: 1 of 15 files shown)


undingen:
memory improvements are:
```
django_template3.py:          131.4 --> 108.5 MB
sqlalchemy_imperative2.py:     97.0 -->  85.0 MB
```

undingen:
this also improves max rss for:
```python
s = """
class C(object):
 def f(self):
  return 'lalala'
C().f()
"""
for i in xrange(50000):
 exec s
```
from 124MB to 38MB.
But it is still growing (but much slower) with the number of the iterations so needs further investigation

src/runtime/types.cpp
```diff
@@ -4054,6 +4055,8 @@ void CodeConstants::dealloc() const {
     owned_refs.clear();
     for (auto&& e : funcs_and_classes) {
         Py_DECREF(e.second);
+        assert(e.first->type() == BST_TYPE::FunctionDef || e.first->type() == BST_TYPE::ClassDef);
+        delete[] e.first;
```
kmod: Should this be just `delete`?

undingen:
I fixed the `delete` ugliness by merging `MakeFunction` with `FunctionDef` and `MakeClass` with `ClassDef` which removes the whole special allocation and `CodeConstant` handling for `FunctionDef` and `ClassDef` nodes I added previously.
I think after this this PR should be ready for merging, please let me know what you think :-)

kmod:
lgtm!

PR #1392, opened 2016-11-03
Get dict ordering closer to CPython's

kmod:
Rather than switching to CPython's dict implementation, which would
be quite a bit of code changes (since we use our internal dict interface
a lot), change our current dict implementation to more-closely match CPython's.

In particular, change the growth strategy and probing strategy to match cpython's.

### DIFF
--- a/src/core/from_llvm/DenseMap.h
+++ b/src/core/from_llvm/DenseMap.h
@@ -446,6 +446,7 @@ private:
 
     TheBucket->getFirst() = Key;
     new (&TheBucket->getSecond()) ValueT(Value);
+    TheBucket = growMaybe(Key, TheBucket);
     return TheBucket;
   }
 
@@ -455,6 +456,7 @@ private:
 
     TheBucket->getFirst() = Key;
     new (&TheBucket->getSecond()) ValueT(std::move(Value));
+    TheBucket = growMaybe(Key, TheBucket);
     return TheBucket;
   }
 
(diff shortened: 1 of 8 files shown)


kmod:
I'm not sure how I feel about this -- there are still a number of differences.
In theory this is still an improvement over the current status quo (fewer ordering
differences), but I don't know if we need to go all the way and make sure to have
zero ordering differences.

kmod:
Ok, I think at this point the ordering is pretty close for the cases that I was concerned about, which is primarily string-keyed dictionaries that don't have any deletes.

There are still quite a few cases when we will have different ordering:
- I didn't make any attempt to get the deletion code to be the same
- Our string type has the same hash code as CPython's, but other types might not
- We don't use dicts to store attributes or globals, so those orderings are still different.

So I think this is still mostly an improvement and I think it makes sense to check in even if this isn't ultimately the approach we want to take.

kmod:
Perf results are neutral:
```
                                           00d51fe:              8939e9:
           django_template3.py             2.3s (2)             2.3s (2)  +0.0%
                 pyxl_bench.py             2.1s (2)             2.0s (2)  -0.6%
     sqlalchemy_imperative2.py             2.5s (2)             2.5s (2)  +0.3%
                pyxl_bench2.py             1.1s (2)             1.1s (2)  -0.9%
             pyxl_bench_10x.py            15.7s (2)            15.5s (2)  -1.4%
       django_template3_10x.py            11.0s (2)            11.0s (2)  -0.4%
 sqlalchemy_imperative2_10x.py            17.0s (2)            17.0s (2)  +0.2%
            pyxl_bench2_10x.py             8.9s (2)             8.9s (2)  -0.1%
                       geomean                 4.9s                 4.9s  -0.4%
```

PR #1393, opened 2016-11-05
jemalloc: build with profiling support

undingen:
After installing `sudo apt-get install libjemalloc-dev` [1] one can use jemallocs heap profiling with:
```
~/pyston$ MALLOC_CONF="prof:true,prof_prefix:jeprof.out" ./pyston_release minibenchmarks/raytrace.py
~/pyston$ build_deps/jemalloc/bin/pprof pyston_release jeprof.out.*
(pprof) top3
Total: 4.5 MB
     1.0  22.2%  22.2%      1.0  22.2% pyston::BST_FunctionDef::operator new (inline)
     0.5  11.1%  33.3%      0.5  11.1% pyston::DenseMap::allocateBuckets (inline)
     0.5  11.1%  44.5%      0.5  11.1% pyston::CFG::addBlock
```
I could not measure any performance regression so I think we can just always compile it with profiling support.

[1]: Unfortunately it can't find the jemalloc header file - instead of finding out how to tell the configure script where to look at I just installed the jemalloc from the distro which is not the same version but nevertheless works...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,7 @@ ExternalProject_Add(libjemalloc
                     SOURCE_DIR ${CMAKE_SOURCE_DIR}/build_deps/jemalloc
                     DEPENDS gitsubmodules
                     UPDATE_COMMAND autoconf
-                    CONFIGURE_COMMAND ${CMAKE_SOURCE_DIR}/build_deps/jemalloc/configure --prefix=${CMAKE_BINARY_DIR}/jemalloc --enable-autogen --enable-prof-libunwind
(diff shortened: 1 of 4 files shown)


kmod:
sgtm :)  I think it'd be helpful to have your PR comment as a longer-lived doc, such as in `docs/MEMORY_PROFILING.md`

I think jemalloc had an option to install the header file, and then we could set up our cmake configuration to know where to find it, but I agree that this doesn't seem that valuable right now.

undingen:
Good idea, I added very basic documentation and fixed the avro test failure we are seeing since a few days (domain is offline? cpython encounters the same error)

PR #1398, opened 2016-11-11
Some small fixings

Daetalus:


src/runtime/classobj.cpp
```diff
@@ -1927,6 +1927,18 @@ extern "C" PyObject* PyMethod_Self(PyObject* im) noexcept {
...
+ extern "C" void PyMethod_SetSelf(PyObject* im, PyObject* instance) noexcept {
+     if (!PyMethod_Check(im)) {
+         PyErr_BadInternalCall();
+     }
+
+     Box* self = ((BoxedInstanceMethod*)im)->obj;
+     Py_XDECREF(self);
+     Py_INCREF(instance);
+     ((BoxedInstanceMethod*)im)->obj = instance;
+     Py_INCREF(im);
```
kmod: I think the pattern is usually
```
Box* old = o->attr;
Py_INCREF(new_attr);
o->attr = new_attr;
Py_DECREF(old);
```
ie you want to decref the old value after replacing it, not before.  I also don't think we need the `Py_INCREF(im)` call here.

src/runtime/builtin_modules/builtins.cpp
```diff
@@ -2381,6 +2386,7 @@ void setupBuiltins() {
         = BoxedClass::create(type_cls, object_cls, 0, 0, sizeof(Box), false, "ellipsis", false, NULL, NULL, false);
     ellipsis_cls->giveAttr("__repr__",
                            new BoxedFunction(BoxedCode::create((void*)ellipsisRepr, STR, 1, "ellipsis.__repr__")));
+    ellipsis_cls->tp_repr = ellipsis_repr;
```
kmod: The proper fix here is to call `ellipsis_cls->freeze()` (and get rid of `ellipsis_repr`), which will automatically fill in tp_repr for us.

src/runtime/types.cpp
```diff
@@ -4405,7 +4403,7 @@ void setupRuntime() {
     type_cls->giveAttr(
         "__new__", new BoxedFunction(BoxedCode::create((void*)typeNewGeneric, UNKNOWN, 4, false, false, "type.__new__"),
                                      { NULL, NULL }));
-    type_cls->giveAttr("__repr__", new BoxedFunction(BoxedCode::create((void*)typeRepr, STR, 1, "type.__repr__")));
```
kmod: What code relies on the type of `type.__repr__`?  This change is a bit scary because it could affect a lot of things.

Daetalus: 'instancemethod` will store the `im_self`(aka `obj` in `BoxedInstanceMethod`. In some situation, it will cause `str(wrapped)` and  `repr(wrapped)` returns different result.

Daetalus:
@undingen Ready to review. :)

src/capi/object.cpp
```diff
@@ -614,6 +614,40 @@ extern "C" PyObject** _PyObject_GetDictPtr(PyObject* obj) noexcept {
     }
 }
 
+// This function returns the copy of object's dict, not a writable dict.
+extern "C" PyObject* PyObject_GetDict(PyObject* obj) noexcept {
+    PyTypeObject* tp = Py_TYPE(obj);
+    if (!tp->instancesHaveHCAttrs()) {
+        if (!tp->instancesHaveDictAttrs())
+            return NULL;
+        return obj->getDict();
```
undingen: I think this will not always return a copy of the dict  (the `obj->getDict();` case)
How do you feel about naming it `PyObject_GetDictCopy`? Or does some other python implementation already use `PyObject_GetDict`?

src/capi/object.cpp
```diff
@@ -614,6 +614,40 @@ extern "C" PyObject** _PyObject_GetDictPtr(PyObject* obj) noexcept {
...
+    }
+}
+
+extern "C" void PyObject_ClearDict(PyObject* obj) noexcept {
+    obj->clearAttrsForDealloc();
+}
+
+// This API just add or update key-value pairs. It will not remove existed items
+// if the key is not contained in the new dict.
+extern "C" void PyObject_SetDict(PyObject* obj, PyObject* dict) noexcept {
```
undingen: What about naming it `PyObject_UpdateDict`?

src/capi/object.cpp
```diff
@@ -614,6 +614,40 @@ extern "C" PyObject** _PyObject_GetDictPtr(PyObject* obj) noexcept {
...
+    PyObject* d_key, *d_value;
+    Py_ssize_t i = 0;
+    while (PyDict_Next(dict, &i, &d_key, &d_value)) {
+        if (PyString_CheckExact(d_key)) {
+            Py_INCREF(d_key);
+            PyString_InternInPlace(&d_key);
+            Py_DECREF(d_key);
+        }
```
undingen: Look like to me like this block is not necessary, `PyObject_SetAttr` should already do the interning

src/capi/object.cpp
Daetalus: Only Jython has a public `getDict` API for different class.

kmod:
Oh yes Marius is right -- there are two issues with `PyObject_GetDictPtr`, the first is that it is supposed to be a pointer to the dict, which `PyObject_GetDict` solves, but the second is that the "dict" might be a copy of the attributes and updates won't propagate back.  So I think `PyObject_GetDictCopy` is a better name.

src/capi/object.cpp
undingen: thanks for changing the name but I think this should call `PyDict_Copy(obj->getDict())` because otherwise it's not always a copy as the name suggests...

undingen:
:+1:  thanks for the patch!

PR #1400, opened 2016-11-15
jemalloc: update to 4.3.1 (latest release)

undingen:
There is no particular reason for upgrading to the last version.
I just wanted to check if memory usage improvement or if the profiling tools get improved but did not notice much changes except for a speed improvement :wink: .


undingen:
ok I had to change the link order because jemalloc with enabled libunwind profiling needs libunwind but got linked last and therefore could not find some symbole

PR #1414, opened 2017-01-19
[Clean Up] Remove unused parmameter in BST_Name.

corona10:
Happy new year, it looks like a little bit late.
Anyway, I found an unused parameter. It looks like pyston wanted to record line numbers.
But, it looks like not using right now.

### DIFF
--- a/src/core/bst.h
+++ b/src/core/bst.h
@@ -775,8 +775,7 @@ public:
     // Only valid for lookup_type == CLOSURE:
     int closure_offset = -1;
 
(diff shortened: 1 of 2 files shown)


corona10:
For reviewers, I tested it on my own Travis CI, with a current master branch.
It failed with ssl_test. It looks like this failing is not related to this PR.

undingen:
yeah looks like `svn.python.org` SSL cert expired today....

undingen:
:+1:  thanks for your patch!

PR #1417, opened 2017-01-21
disable some tests on test_ssl.py since svn.python.org's cert is expired now

aisk:
cpython's tests don't reply on it since v2.7.12, so i think we should just disable it until we updated to 2.7.12.


kmod:
Yeah, not great but it doesn't seem like there's a more targeted solution here :/

PR #1418, opened 2017-01-30
Micro optimizations for string operations.

corona10:
I love the Pyston's useful strategey about an one length string and an empty string.
But some operations are did not using this lovely strategy. so I change it to avoid  unnecessary operations. For a swapcase() operation, we can just handle it by touching a 1bit.
And I checked Travis CI were passed except Avro integration tests, but I found that it happens on master branch. so it is not related with this PR.


corona10:
For reviewers, Original swapcase() might be better. I will revert this function.

PR #1421, opened 2017-03-20
remove "version 0.7 coming soon"

deronnax:
let's be realistic

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,6 @@ See [travis-ci.org/dropbox/pyston/builds](https://travis-ci.org/dropbox/pyston/b
 
 ### Roadmap
 
-##### v0.7: coming soon
-
 ##### v0.6.1: [released 1/31/2017](https://blog.pyston.org/2017/01/31/pyston-0-6-1-released-and-future-plans/)
 - smaller performance optimizations and bug fixes
 - updated CPython runtime to 2.7.8


leycec:
**So, it's come to this.** I've only been following Pyston tangentially, but the last commit date of January 31st coupled with this pull request makes for an alarming picture.

Pyston is dead, isn't it? Why does this continually happen to performance-oriented Python efforts? It's like the `setuptools` replacement curse all over again. I'm beginning to lament that Python will *ever* have a general-purpose JIT. Ruby and Lua developers have long enjoyed general-purpose JITs – and here we are, still languishing in the unoptimized darkness of CPython.

At least we still have Numba.... *for now.*

leycec:
`</sigh>`

To no one's surprise, Dropbox [dropped Pyston](https://blog.pyston.org/2017/01/31/pyston-0-6-1-released-and-future-plans) on January 31st – and nobody sent me the memo. What probably needs to happen now is that somebody else (*...not me*):

* Officially fork this repository, ideally under a different project name.
* Refactor the codebase to support the new JIT API introduced with Python 3.6. (See: "What is [Pyjion](https://github.com/Microsoft/Pyjion)?")

It'll never happen, of course. But the lazy man can dream big.

xoviat:
What's wrong with pypy? I mean some extensions don't work but that's being improved.

PR #1423, opened 2017-06-26
Subclassing the bool from long instead int object.

Daetalus:
In Python 2, the bool object is essentially an int object. In Python 3, it is long object. This PR subclass the bool object from long instead of the int. Preparing for removing the int object completely.


kmod:
Seems reasonable to me!  I'd recommend adding `# expected: py3ready` or similar so that we can still run some tests and get feedback from the CI system.

Also I just created a "py3" branch where I was thinking we could do this development.

PR #1424, opened 2017-07-01
Subclassing the bool from long instead int object.

Daetalus:
In Python 2, the bool object is essentially an int object. In Python 3, it is long object. This PR subclass the bool object from long instead of the int. Preparing for removing the int object completely.


Daetalus:
@kmod

src/runtime/types.cpp
```diff
@@ -4304,9 +4304,9 @@ void setupRuntime() {
     attrwrapper_cls->tp_mro = BoxedTuple::create({ attrwrapper_cls, object_cls });
     dict_cls->tp_mro = BoxedTuple::create({ dict_cls, object_cls });
     int_cls->tp_mro = BoxedTuple::create({ int_cls, object_cls });
-    bool_cls->tp_mro = BoxedTuple::create({ bool_cls, int_cls, object_cls });
     complex_cls->tp_mro = BoxedTuple::create({ complex_cls, object_cls });
     long_cls->tp_mro = BoxedTuple::create({ long_cls, object_cls });
+    bool_cls->tp_mro = BoxedTuple::create({ bool_cls, int_cls, object_cls });
```
kmod: You'll need to update this to be long_cls instead of int_cls.

kmod:
Looks awesome!  Let's get a small test file set up, something like this:

```python
long_ = type(1L)
assert isinstance(True, long_)
assert issubclass(bool, long_)
print(bool.__bases__)
```

PR #1425, opened 2017-07-11
Use CPython 3 to build Pyston and run Pyston tests.

Daetalus:
When I try to add tests for new long based bool implementation. I found I have to use CPython 3 to run the tests. This is why this PR came from. @kmod 

This PR includes these major changes.:

- Use CPython 3.5.2 to build and run Pyston tests.
- Remove all CPython 2 tests : I am ok if you prefer to update the tests one by one.
- Add some temporary fixings in the source code: Due to Pyston is a large code base. It is hard to update code without temporary fixing. I added "py3updationg" marker to notice this is a temporary fixing. All these temporary fixings will be addressed when releasing Python Python 3 support.
- Use `2to3` to update the test running tools.


kmod:
Sounds reasonable!  We should have a quick discussion about the Python 3 version that we want to target; 3.5.2 sounds reasonable, as would 3.6.1 (the current latest).  It's not a huge deal, but it's potentially something that will stick for a while so I think it's worth talking about.

What do you think about copying in the Python 3 tests, instead of only deleting the Python 2 tests?  It could help keep the diff a bit cleaner.  (This is the sort of thing where we'll want to pick a single CPython revision to start copying from.)

from_cpython/Objects/stringlib/formatter.h
```diff
@@ -1419,8 +1419,10 @@ FORMAT_LONG(PyObject *obj,
             STRINGLIB_CHAR *format_spec,
             Py_ssize_t format_spec_len)
 {
-    return format_int_or_long(obj, format_spec, format_spec_len,
-                              long_format);
+    // py3updating: Temparary solution.
```
kmod: If these are compile-time issues, you could put something like `RELEASE_ASSERT(0, "unimplemented");` to make it even more obvious.

test/tests/builtins.py
```diff
@@ -1,163 +0,0 @@
-import sys
```
kmod: What happens if we keep these tests (but leave them disabled)?

kmod: What happens if we keep these tests (but leave them disabled)?

Daetalus:
Updated. @kmod

kmod:
Looks great!  btw, why only change those particular tests?

